PR #2, opened 2022-09-26
Create test_norm2mort.py

GTBC0:


espg (commented):
...need to use the `assert` keyword for this to be a unit test. General practice is to create a folder (tests) for all the unit tests, and then import the functions that you are testing using relative imports-- see [here](https://github.com/scikit-learn/scikit-learn/tree/main/sklearn/cluster) and [here](https://github.com/scikit-learn/scikit-learn/tree/main/sklearn/cluster/tests) for how the file hierarchy is setup...

PR #3, opened 2022-10-07
Fixed txt file

GTBC0:
The txt file now contains the data in correct columns and got rid of unwanted text.


espg:
Can you please unzip the file, so it's just plain text? General rule is that no binary files go into git-- so for cython, the .pyx get committed, but not the binary object that is compiled from the source code.

You may have to revert or delete the file in git to make it look correct in the git log... moves and deletes are handled a bit strangely in git, so changing `ant_grounded_drainagesystem_polygons.txt.zip` to `ant_grounded_drainagesystem_polygons.txt` will require a bit of poking around.

PR #6, opened 2025-11-13
Rust <> Python API compatibility

espg:
The way that the rust name space is setup causes easy conflicts with python... this should fix things by naming the rust extension `rustie` and the python library/interface `mortie`.


espg:
...and using `mortie` will just call `rustie`

Issue #28, opened 2026-06-03
morton_coverage returns non-deterministic output on identical inputs


espg:
# `morton_coverage` returns non-deterministic output on identical inputs

## Summary

Repeated calls to `morton_coverage` with the exact same arguments return
**different cell sets** intermittently. Two distinct outputs are produced
on the same input; the smaller is always a strict subset of the larger.
About 20-40% of calls give the smaller (incomplete) result.

This breaks any code that relies on `morton_coverage` being a pure
function — including any spatial-intersection pipeline that compares
coverage sets across polygons.

## Reproducer

```python
import numpy as np
from mortie import morton_coverage

lats = np.array([-89.0, -59.09804617, -59.09804617, -89.0, -89.0])
lons = np.array([105.5108378, 105.5108378, 106.5108378, 106.5108378, 105.5108378])

sizes = []
sets = []
for _ in range(20):
    c = morton_coverage(lats, lons, order=10)
    sizes.append(len(c))
    sets.append(frozenset(int(x) for x in c))

print('sizes:', sizes)
print('unique outputs:', len(set(sets)))

small = min(sets, key=len)
big = max(sets, key=len)
print(f'|small|={len(small)}  |big|={len(big)}')
print('small ⊂ big?', small.issubset(big))
```

Output:

```
sizes: [4392, 1166, 1166, 1166, 1166, 1166, 1166, 1166, 1166, 4392,
        4392, 4392, 1166, 1166, 1166, 4392, 1166, 1166, 4392, 1166]
unique outputs: 2
|small|=1166  |big|=4392
small ⊂ big? True
```

- Same input, same call, different outputs.
- Output is always one of exactly two values.
- The smaller output is a strict subset of the larger (missing ~3,200 cells).
- `lats`/`lons` arrays are not mutated between calls.

## Environment

- `mortie 0.6.3` (PyPI wheel for Python 3.13, manylinux x86_64)
- `numpy 2.x`
- Fedora 42, Python 3.13.5, x86_64

## Observations

- The polygon is a thin lon-strip from lat -89 to -59 across ~1° longitude
  near the south pole. Polar geometry may matter.
- Both outputs contain only valid HEALPix cells at the requested order.
- Calling `morton_coverage` in isolation (script that does nothing else)
  on simpler polygons gives consistent output — so the bug is not always
  triggered. Polygons spanning near-polar latitudes seem to be a
  trigger condition.

## Impact

We hit this while building a sphere-aware `build_catalog` backend in
[...]

Issue #23, opened 2026-02-26
Support polygons with holes in morton_coverage


espg:
`morton_coverage()` currently only supports simple polygon rings (outer boundary). Polygons with holes (e.g., a ring with interior cutouts) are not detected or handled — passing a flat list of outer + inner ring vertices produces undefined results.

### What needs to happen

- Detect hole rings (e.g., via winding order convention: CCW = outer, CW = hole)
- Exclude cells whose centres fall inside a hole from the coverage
- Validate and raise a clear error if hole geometry is malformed

### Context

- Issue #20 and PR #21 both note this as deferred work
- `test_holes_not_supported` in `test_coverage.py` currently skips with a reference to this
- The gnomonic PIP infrastructure (PR #21) should make hole exclusion straightforward — run the winding number test against each hole ring and subtract

espg:
closed by #31

Issue #22, opened 2026-02-26
Support hemisphere+ polygons via S2-style edge crossing PIP


espg:
## Summary

The current `morton_coverage` PIP implementation (after PR #21) uses gnomonic projection + 2D winding number, which works correctly for all polygons within one hemisphere of the test point. This covers all practical geospatial use cases (Antarctic drainage basins, country boundaries, etc.).

However, for polygons that span more than a hemisphere (e.g., "everything except Antarctica"), the gnomonic approach fails because the projection is only valid within a hemisphere of the projection center.

## Proposed Enhancement

Implement **S2-style edge crossing** as an alternative PIP backend for hemisphere+ polygons:

- Uses crossing number from a fixed reference point
- Core operation: scalar triple product `A · (B × C)` to test if great circle arcs cross
- Handles all polygon sizes, antimeridian, poles
- Robust with exact predicates (Shewchuk-style)
- Zero trig functions
- Higher implementation complexity than gnomonic

## References

- Google S2 Geometry Library: https://s2geometry.io/
- Ketzner, Ravindra & Bramber (2022) — Robust fast edge-crossing, lune-based spatial indexing
- [Issue #20 comment on spherical PIP research](https://github.com/espg/mortie/issues/20)

## Implementation Notes

- Could be added as a fallback: detect when any polygon vertex is >90° from any test point, switch to S2-style
- Or expose as a separate `pip_method` parameter
- The gnomonic approach should remain the default (faster for typical polygons)

## Priority

Low — no current use case requires hemisphere+ polygons. This is tracked for completeness.

espg:
## Re-scoping #22 after the hierarchical-coverage rework (#30 / #31)

The hierarchical coverer (#31) changes the picture a lot. The gnomonic limitation
is now confined to a **very small surface**, and closing #22 is mostly about one
robust point-in-polygon primitive plus one cull guard — not a second pipeline.

### Key insight: the backend is used in only one place

In the new descent the per-cell fill is maintained by **edge-culling +
fill-propagation**: each node inherits its parent's fill, flipped by the parity
of polygon edges crossing the short parent-centre → child-centre arc
(`arc_crossing_parity`, orientation-only, always valid). The hemisphere-limited
point-in-polygon test (`parity_filled` → `gnomonic_pip` / `point_in_ring_edgecross`)
runs **only at the 12 base-cell seeds** (`coverage.rs::base_node`). So:

- For small/medium polygons the backend is ~12 calls — performance-irrelevant.
[...]

Issue #20, opened 2026-02-26
Polygon to morton coverage


espg:
This is mostly notes for claude to implement this feature, but posted publicly for reproducibility and insight; think of it as a mini-ATBD. For the record, I honestly find it shocking that this hasn't already been done given how widely used healpix is-- but the JPL polygon bindings (while usually quite good) are totally lacking in this area-- they only work for convex polygons which makes them useless to us. Apparently astronomers don't use complicated shapes when querying things. 

We want a function that takes a polygon as an input, and returns an array of morton indices (at a user defined order) that cover the polygon extent completely, while also being optimally compact.

We recently added rust bindings in this library for a new buffer operation, `morton_buffer`; we're going to leverage this to build our algorithm. Here's the outline of what we're implementing:

  1. Take a set of polygon vertices and cast them to morton indices at the user requested morton order.
  2. Use `morton_buffer` to get the k=1 neighbors of the polygon; we're going to separate these nieghbors to a new partion, and then we're going to segment the neighbors into two groups-- inner neighbors, and outer neighbors.
  3. We won't know which group are the inner neighbors, and which group are the outer neighbors, so we'll have to test. In order to test this, well take the segmented group, and then cast them *back* from morton indices to lat/lon points... technically it could be any coordinate space, but we'll realistically implement this for lat/lon ; we can talk about an optional projection option if we find issues with stability, but ideally we just use spherical coordinates.
  4. The that the partitioned coordinate point sets are now back as coordinates, we'll sample some number of those points, from each partition, to hold out for testing at the next step. This could a small percentage (e.g., 0.5%), or just a fixed number (e.g., 5 or 10)-- likely, the best approach will be to set a floor on the minimum number of points (for stability), the max number of points (for performance) , and take a log transform on the len() of the partitions where we either use the log or snap to floor / ceiling we define if the log is outside of that range.
  5. Next we'll take the original **input** polygon, and we'll test both of the partitioned groups (using the random test points from above) for if those points are inside, or outside of the input polygon. I assume that we'll use the Ray Casting Algorithm for this, but feel free to suggest an alternative if appropriate.
[...]

PR #31, opened 2026-06-03
feat: hierarchical coverage from polygon (correctness, native MOC coverage, 'donut' polygon support)

espg:
New algorithm for **native** Multi-order Coverage (MOC) as output from a given polygon. Addresses limitations in current implementation (i.e., correctness, maximum coverage and input geometry restrictions).

Improvement on #20 , addresses #22, #23, as well as the follow on in #28 (and indirectly , the issue in #11 )


espg:
Phase 1 summary / notes — sphere.rs

  New module (src_rust/src/sphere.rs, registered in lib.rs) with the spherical primitives the hierarchical coverer is built on:

  - Core predicates: orient(a,b,c) (scalar triple product) and arcs_cross(a,b,c,d) (4-orientation great-circle arc crossing, minor-arc precondition documented).
  - Two PIP backends: gnomonic_pip (carried over from the merged coverage code, the ≤hemisphere fast path) and point_in_ring_edgecross (orientation-only, the basis for #22
  hemisphere+ support).
  - Even-odd ring-set fill: parity_filled(p, rings, backend) — the one predicate that gives holes and multipart for free (inside an odd number of rings).
  - Backend auto-select: choose_backend (gnomonic until the geometry approaches a hemisphere, then edge-cross) plus ring_cap_axis / ring_angular_extent helpers.
  - 8 unit tests: orientation signs, arc crossing (crossing/disjoint), gnomonic↔edge-cross parity agreement, even-odd donut, multipart, backend selection, southern-hemisphere
  triangle. All green; full suite 76/76, no warnings.

  Two notes from the build:
  - The edge-cross reference point is taken ~90° from the cap axis (not the antipode) to avoid arcs_cross's near-180° degeneracy — found and fixed via a failing test.
  - A test caught itself: the great-circle edge convention means a "constant-latitude" triangle edge bulges poleward (~−72.5° here), so the gnomonic backend correctly excludes
  points the flat-lat-lon intuition would include. Good confirmation the edge convention is working as designed.
  - #![allow(dead_code)] is on the module intentionally (it's built ahead of being consumed); that allow gets removed in Phase C when the descent uses these functions.

espg:
Phase 2 summary / notes — cell_geom.rs

  New module (registered in lib.rs) providing cell geometry + the ring-set classifier the descent will call at every node:

  - cell_corners(depth, pixel) — wraps boundaries_scalar into 4 corner unit-vectors; cell_centroid for a known-interior reference point.
  - point_in_cell — orientation-only convex test (a point is inside iff it's on the centroid's side of all 4 directed edges); winding-direction-independent.
  - classify(depth, pixel, rings, backend) -> Inside | Outside | Straddle — the single rule from the plan:
    a. any ring edge crosses any cell edge → Straddle;
    b. else any ring vertex inside the cell → Straddle (catches a whole tiny polygon/hole sitting inside the cell with no edge crossing);
    c. else uniform → the centroid's even-odd parity decides Inside/Outside.
  - 8 tests: inside / outside / straddle-on-edge / contains-tiny-polygon, plus the three donut cases (annulus→Inside, hole→Outside, hole-rim→Straddle), and point_in_cell. All
  green first build; full suite 84/84, no warnings.

espg:
## Phase 3 summary — descent + MOC

  coverage.rs rewritten as a top-down hierarchical coverer:

  - descend() — parallel stack-DFS over the 12 base subtrees, refining only Straddle cells to the target order. No boundary rasterization, buffer ring, or flood fill — so it's a pure deterministic function of the inputs.
  - polygon_to_morton_coverage() — flat single-order output (back-compatible), via moc::to_order.
  - polygon_to_morton_moc() — compact mixed-order MOC, via moc::normalize (new, ready for the Phase D Python binding).
  - Kept the three great-circle helpers linestring.rs depends on; removed the dead boundary/buffer/gnomonic/winding code and their now-redundant tests.

  moc.rs (new) — normalize (ancestor-prune + sibling-merge) and to_order (densify), exploiting that mortie's decimal morton self-encodes order and ancestry is a digit-prefix.

  6 tests incl. densify-invariance.

  cell_geom.rs — the classifier, with one important correction found mid-build.

 ## Bug worth flagging

First classifier treated a HEALPix cell as a great-circle quad. But HEALPix cell edges are not great-circle arcs — so that test disagreed with HEALPix's own ang2pix near  boundaries, and, fatally for a tree descent, disagreed between a parent and its child: a point provably inside child cell 621 tested as outside its parent 155, pruning the correct subtree (tiny polygons vanished at order ≥4). Fix: vertex-in-cell containment is now decided exactly by HEALPix — each vertex's leaf cell is computed once and its ancestor at any depth is a bit-shift — backed by corner+centre even-odd parity and edge-crossing. Added test_classify_consistent_parent_child as a regression guard.

## Verification

  - Rust: 85/85, no warnings.
  - Python: 206 passed, 3 skipped (incl. the tiny-polygon test that was failing).
  - #28: deterministic at 3139 cells (vs the hotfix's 3074 — the expected one-cell-thick boundary shift under contract (a)).
  - Independent oracle: 0 interior holes — correct intersection cover.

  #28 is now fixed by construction — there's no sampling, no hash iteration, no flood-fill seed anywhere in the path.

espg:
## Phase 4 — benchmark results: hierarchical descent vs. merged hotfix

Wall-clock, median of 7 runs, same machine. Cell counts now match the hotfix
closely, and correctness is independently verified (0 interior holes on every
shape via a center-inside oracle).

| polygon | verts | order | hotfix | hierarchical | speedup |
|---|--:|--:|--:|--:|--:|
| triangle | 3 | 10 | 10.9 ms | **1.1 ms** | 9.9× |
| triangle | 3 | 12 | 141.2 ms | **6.5 ms** | 21.7× |
| square | 4 | 10 | 19.8 ms | **3.6 ms** | 5.4× |
| square | 4 | 12 | 288.9 ms | **13.5 ms** | 21.4× |
| polar strip | 4 | 12 | 30.3 ms | **7.7 ms** | 4.0× |
| circle | 100 | 10 | 7.5 ms | **4.0 ms** | 1.9× |
| circle | 500 | 10 | 24.1 ms | **12.1 ms** | 2.0× |
| **circle** | **100** | **6** | **0.19 ms** | 0.54 ms | **0.34× ⚠️** |
| **circle** | **500** | **6** | **0.22 ms** | 0.85 ms | **0.26× ⚠️** |

### Read

- **Big wins everywhere realistic** — 2× to **22×** faster at working
  resolutions (orders 10–12). The descent captures the interior as a few coarse
  cells (MOC) instead of flood-filling millions of leaves, so runtime scales
  with the *boundary*, not the area.
- **One regime is slower — the #29 corner itself:** high-vertex polygons at
  *coarse* order (6), producing a ~46-cell result. There `classify` is still
  **O(vertices) per cell** (it scans every vertex/edge), so a 500-vertex circle
  pays per visited cell. Sub-millisecond in absolute terms, but ~3–4× slower
  than the hotfix.

### What closes the slow corner

The plan's promise — *"work ∝ boundary length, not vertex count"* — needs the
**edge-culling / fill-propagation** optimization that isn't built yet:

1. **Edge culling** — thread the *relevant* polygon edges down each subtree so
   deep cells only test nearby edges (`classify` becomes O(local edges)).
2. **Fill propagation** — carry the inside/outside state from parent to child,
   so a uniform cell with no nearby edge needs *no* point-in-polygon call (only
   the 12 base cells pay a full parity test).

Together these turn per-cell cost from O(all vertices) into O(local edges) and
should flip the order-6 circles to wins as well. This was listed as a Phase B/C
TODO in #30; the benchmark confirms it's the missing piece, and I'd suggest
landing it as the next step before the holes/multipart wiring.

### Two bugs the benchmarking surfaced (fixed, staged)

1. **Far-cell over-cover.** The 5°-radius circle came out at **12 904 cells
   (~2×)** with spurious cells near the *north pole / antimeridian*. Cause: the
   gnomonic point-in-polygon test returns garbage for test points >90° from the
   polygon (all vertices land on the far hemisphere and get clamped), and the
   top-down descent classifies *far* cells that the old flood-fill never
   touched. Fixed with **bounding-cap culling** (`Cap::of_rings` + an early
   `Outside` for any cell entirely outside the polygon's cap) — which doubles as
   the far-subtree perf prune. Circle is now **6 452 cells** (thin 254-cell
   boundary, 0 holes), matching the hotfix's 6 422.
2. Re-verified cell counts and the 0-holes property across circle / polar after
   the fix. Full Python suite **206 passed**, Rust **85 passed**, no warnings.

### Bench harness changes (this PR)

- `src_rust/benches/coverage_bench.rs` — added `circle500` across orders 6/8/10
  (the #29 corner + interior-dominated regime) and a flat-vs-MOC output group.
- `benchmarks/test_bench_coverage.py` — added order-10 circle benches.

> **CodSpeed caveat:** because `main` is the hotfix baseline, CodSpeed will flag
> the order-6 circle benches as regressions until either edge-culling lands or
> the baseline resets on merge. Everything at order ≥ 10 improves.

Issue #35, opened 2026-06-10
Major Feature / Enhancement — Full resolution Morton MOC datatype


espg:
This is a deep tech issue for mortie, which involves a fundamental change to **all** of the functionality in this library.

Currently, mortie only has support for 19 orders of indexing -- starting at order 0 for the base cells, and then proceeding up to order 18, with an approximate cell resolution of ~25 meters. This has been 'fine' for a large part of what mortie actually was originally used for-- determining hive prefix strings for S3 hierarchical storage. These are expected to have a 'normal' data type (i.e., int64), and be human readable. Packing morton indices this way isn't particularly efficient, so we "run out" of space in the 64bit integer past order 18.

We now have **current** users and use cases for morton cells at the ~10 meter resolution, that we *cannot* support. So a rewrite for full resolution MOC is needed to both meet current needs, as well as to future-proof additional development.

The healpix specification itself is defined up to order 29 (30 total orders including base cells at order 0), so the goal is to match that encoding range so that we can be compatible with any cross libraries in the healpix ecosystem (via the normal calls to the rust crate to convert via nested and UNIQ representations). We also want this to be a more intrinsic MOC representation-- that captures the order explicitly rather than implicitly. Finally, we want to *keep* the existing string representation-- we'll still want to parse strings from hive labels, and be able to emit and operate on them like we do today... but with the added ability to operate on our new `decimal_morton` datatype (or `morton_index` or whatever it gets called and registered as).

One major point that I'm still fuzzy on is how to declare and register new datatypes to numpy -- my guess is that this will involve defining operator overloads, which will require quite a bit of thought as we move forward with the overhaul.

## Datatype design

First and foremost, we want this to be a 64bit type. Here's the proposed layout:

[4-bit prefix] + [54-bit body] (body holds 27 bit pairs) + [6-bit suffix]

### Prefix

The prefix encodes the base cell -- there are 12 base cells, so we need 4-bits here. Our general representation of the morton index is *signed* so that cells fully in the northern hemisphere are positive, and cells fully in the southern hemisphere are negative (mid latitude cells are mixed... although thinking of it now, I *hope* that the 2 cells east of the prime meridian are positive, and the two cells west of it are negative). In any case, we'll need to either have a 'signed' 4-bit integer, or map values 0-15 on decode/encode to their signed equivalents. In any case, the 4-bit prefix defines the base cell.

### Body

[...]

Issue #34, opened 2026-06-06
Optimization & duplication audit + moc/RangeMOC adoption research


espg:
# Mortie — Optimization & Duplication Audit + `moc`/RangeMOC Research

> Deep-dive audit of the mortie codebase (Rust core + PyO3 bindings + Python layer)
> for performance, duplication with the upstream `healpix` 0.3.2 crate, and enhancement
> opportunities; plus research on the CDS `moc` crate / `RangeMOC` and what adopting it
> would mean for mortie.
>
> Findings were produced by parallel code-reading agents and the **highest-leverage items
> were re-verified against source** (marked ✓ verified). Nothing here is benchmarked —
> "hot path" calls are from call-graph reading, primarily the polygon-coverage descent
> which dominates boundary-heavy real inputs (Antarctic basins, ATL06 granules).

---

# PART 1 — Codebase optimization & duplication audit

### 1. [perf ✓] Cache cell corners / circumradius on `Node` in the coverage descent
`src_rust/src/coverage.rs` — `cell_corners()` is called **three independent times per descent node**: `node_straddles:361`, `node_children:394` (per child), and `node_radius:411` (tolerance path). Each call runs `boundaries_scalar` → `get(depth).vertices(pixel)` → 4 `LonLat` + 8 trig evals. Corners computed in `node_straddles` are discarded and recomputed in `node_children`.
**Fix:** compute a node's corners (and circumradius — already produced as `cos_cr/sin_cr` in `base_node`/`node_children` then thrown away) **once** and thread them through (store on `Node`, or return them from `node_straddles`). This is the single biggest per-cell win in the descent hot loop. *Confidence: high.*

### 2. [perf ✓] Precompute per-edge great-circle normal `cross(a,b)` in `Edge`
`src_rust/src/sphere.rs:83-92` (`arcs_cross`), used in the descent inner loop (`coverage.rs:255,364,382`). `arcs_cross(a,b,c,d)` recomputes `cross(b,c)`/`cross(b,d)` every call, but the polygon edge `(a,b)` is fixed across many cell corners. The upstream `healpix` crate avoids exactly this by caching `Polygon::cross_products` once per edge (`sph_geom/mod.rs:111,514-526`).
[...]

Issue #1, opened 2022-09-08
Inconsistent behavior due to numba... 


espg:
see https://github.com/pangeo-data/pangeo-docker-images/issues/297 and https://github.com/numpy/numpy/issues/20735

currently use of jit'ed code will work on older systems but not newer due to bugs in the numba / numpy interaction.

Issue #32, opened 2026-06-04
morton_coverage: polygon-edge interpretation differs from S2 / shapely geodesic


espg:
# `morton_coverage` polygon-edge interpretation differs from S2 / shapely geodesic

## Background

While verifying a sphere-aware `build_catalog` backend in [zagg](https://github.com/englacial/zagg) against the production ATL06 cycle 22 catalog, I found a small but consistent disagreement between `mortie.morton_coverage` and S2-backed polygon intersection (via [spherely](https://github.com/benbovy/spherely)) at high latitudes.

This was originally noted as a "Secondary finding" in [#28](https://github.com/espg/mortie/issues/28), but that issue closed with the non-determinism fix in #29. Filing as a clean standalone issue since the underlying interpretation gap remains.

## The disagreement

For 4103 CMR ATL06 granules in cycle 22, building the catalog three ways:

| backend | (granule, shard) pairs | false negatives vs ground truth |
|---|---|---|
| shapely-EPSG:3031 (legacy, ground truth) | 76,575 | — |
| spherely (S2 geodesic) | 76,575 | 4 (0.005%) |
| **mortie at order 6-10** | **76,566-76,581** | **13 (0.017%) at every order** |

The 13 mortie misses are **stable across orders** — increasing `order` doesn't reduce them. The cause is interpretation, not resolution:

- `morton_coverage` treats polygon edges as rhumb-line-ish (cells covered by polygon interior under a lat/lon-constant edge interpretation).
- S2 (and shapely-in-projection) treat edges as geodesics — great-circle arcs that bulge poleward at high latitudes.

For polygons whose edges *just barely* miss a HEALPix cell under the rhumb-line interpretation but whose geodesic bulge *does* graze the cell, mortie drops the cell from its coverage.

## Reproducer

```python
import numpy as np
from mortie import morton_coverage
import spherely

# A thin granule polygon spanning -88° to -70° lat near the south pole
g_lats = np.array([-88.0, -70.0, -70.0, -88.0, -88.0])
g_lons = np.array([45.0, 45.0, 46.0, 46.0, 45.0])

# A small shard polygon nearby but lon-disjoint
s_lats = np.array([-83.0, -81.0, -81.0, -83.0, -83.0])
s_lons = np.array([40.0, 40.0, 44.5, 44.5, 40.0])

# Mortie says: no overlap (rhumb-line interpretation)
g_cells = set(int(c) for c in morton_coverage(g_lats, g_lons, order=10))
[...]

PR #41, opened 2026-06-14
#34 perf cluster (before #35): GIL release + descent hot-loop + micro-wins + batched vec2ang

espg:
Refs #34 — the "land **before** #35" perf cluster from the [audit split you green-lit](https://github.com/espg/mortie/issues/34#issuecomment-4700873633). #34 stays open as the tracker. Phased, one commit per phase; every phase is output-invariant **except** #10 (phase 4), which additionally fixes a latent bug.

## Phases — the "before #35" perf cluster (#34)

- [x] **#5** release the GIL across all bindings *(phase 1)*
- [x] **#1** cache cell corners / circumradius on `Node` in the coverage descent *(phase 2)*
- [x] **#2** precompute per-edge great-circle normal `cross(a,b)` on `Edge` *(phase 2)*
- [x] **#3** `SmallVec` for `Node.relevant` — **needs a dependency decision** (adds the `smallvec` crate; §4 — see Questions)
- [x] **#10** batch `vec2ang` + vectorize antimeridian fix-up in `mort2bbox`/`mort2polygon` *(phase 4 — also fixes a latent multi-cell bug)*
- [x] **#6** shift/mask + `ilog10` micro-wins *(phase 3)*

(The encoding-coupled items #4/#7/#8/#9 fold into #35, not here.)

## Phase 1 (#5) — release the GIL

Every `#[pyfunction]` in `src_rust/src/lib.rs` previously ran its (often rayon-parallel) compute **while holding the GIL**. This wraps the pure-Rust compute region of all 15 bindings in `py.allow_threads(|| ...)`, so other Python threads (e.g. Dask worker threads) run during the heavy work. Output-invariant. Inputs are copied out of the numpy buffers (`.to_vec()`) before the wrapped region, so each closure captures only owned/`Copy`/`&Vec` data (`Send + Ungil`); the `catch_unwind` guards move inside; the `warnings` import in the MOC paths stays **outside** `allow_threads`. `mortie/tests/test_threading.py` covers six bindings under concurrency and `test_gil_released_during_rust_compute` proves the GIL is actually released (fails on a revert).

## Phases 2–4 (this run)

Per [@espg's note](https://github.com/espg/mortie/pull/41#issuecomment-4705019405) ("why did you stop implementing?? Are all the above items done yet?") the rest of the cluster is now on this branch:

- **Phase 2 (`7d2f906`) — descent hot-loop (#1 + #2).** `Node` caches its 4 corners and cosine-circumradius (already computed at construction for edge culling), so `node_straddles`/`node_radius` reuse them instead of refetching corners from HEALPix. Added `arcs_cross_n` — a normal-aware form of `arcs_cross` where `orient(a,b,x)=(a×b)·x` becomes a plain dot product — and a precomputed per-edge great-circle normal `n_ab = a×b` on `Edge`. The probe-arc normal is computed once per edge fan and the cell-edge/boundary normals once per cell, so the descent's inner crossing tests are dot-products only. **Output-invariant** (the two `orient` formulations differ only in float *sign* on exact-zero ties — the touching/coplanar case `arcs_cross` already treats as `false`; a new `cargo` parity test confirms agreement off the degeneracy).
- **Phase 3 (`89909cb`) — micro-wins (#6).** Power-of-two divide/modulo → shift/mask in `geo2mort_scalar`, `nested2mort`, `mort2nested`; `decimal_digit_count` loop → `u64::ilog10`.
- **Phase 4 (`2e7148f`) — batched `vec2ang` (#10).** `mort2bbox`/`mort2polygon` do **one** `vec2ang` over all cells' flattened vertices instead of one Rust round-trip per cell. **This also fixes a latent bug:** `mort2bbox` indexed the boundary array on the wrong axis (`boundaries[:, :, i]`) for multi-cell input — it raised `ValueError` for any array length ≠ 3 and returned *wrong* boxes at length 3 (the old array test only checked `len`/keys, never values). A regression test now asserts array results equal the per-element scalar reference at N ∈ {2,3,4,5} including an antimeridian cell.

## How it was tested

- `cargo test --release` → **86 passed** (adds the `arcs_cross_n`↔`arcs_cross` parity test).
- `maturin develop --release` + full `pytest` → **227 passed, 8 skipped** (the 8 are the `MORTIE_FORCE_PYTHON`/optional-dep skips); identical pass set to baseline plus the new regression test, confirming phases 2–3 are output-invariant.

## Questions for review

1. **#3 `SmallVec` needs your sign-off (new dependency).** Item #3 (`SmallVec<[usize; 8]>` for `Node.relevant`, to kill the per-child `Vec` reallocations the audit flags) adds the **`smallvec`** crate — a new dependency, so per §4 I did **not** add it silently. It's a tiny, ubiquitous MIT/Apache-2.0 crate with no transitive deps. **OK to add `smallvec` and land #3?** Otherwise I can leave #3 out or try a fixed-capacity inline alternative — your call.
2. **Pre-existing `cargo fmt` drift / clippy warnings** are untouched (the crate isn't globally `fmt`-clean and `clippy` reports ~21 pre-existing `useless_conversion`/`needless_range_loop` warnings on `main`); none introduced here, left per the "don't fix unrelated pre-existing CI noise" rule.

### DIFF
--- a/mortie/tests/test_mort_inverse.py
+++ b/mortie/tests/test_mort_inverse.py
@@ -132,6 +132,11 @@ class TestMort2Geo:
             poly_ref = [tools.mort2polygon(int(m)) for m in mortons]
             assert poly_arr == poly_ref, f"mort2polygon mismatch at n={n}"
 
+            # step > 1 is a separate boundary branch (ncols = 4*step).
+            poly2_arr = tools.mort2polygon(mortons, step=2)
+            poly2_ref = [tools.mort2polygon(int(m), step=2) for m in mortons]
+            assert poly2_arr == poly2_ref, f"mort2polygon(step=2) mismatch at n={n}"
+
     def test_negative_morton_hemisphere(self):
         """Test that negative morton indices encode polar proximity correctly"""
         # Test high northern latitude - should be positive morton
--- a/mortie/tests/test_mort_inverse.py
+++ b/mortie/tests/test_mort_inverse.py
@@ -112,6 +112,26 @@ class TestMort2Geo:
         assert len(polygons) == len(mortons)
         assert all(poly[0] == poly[-1] for poly in polygons)
 
+    def test_bbox_polygon_array_matches_scalar(self):
+        """Array mort2bbox/mort2polygon must equal the per-element scalar calls.
+
+        Regression: mort2bbox indexed the boundary array on the wrong axis for
+        multi-cell input (it only avoided a shape error at length 3), silently
+        returning wrong boxes.  The batched path fixes this and must agree with
+        the scalar reference at every length, including an antimeridian cell.
+        """
+        lats = np.array([40.0, -70.0, 12.0, 5.0, 80.0])
+        lons = np.array([-120.0, 30.0, 179.9, -179.9, 0.0])
+        for n in (2, 3, 4, 5):
+            mortons = tools.geo2mort(lats[:n], lons[:n], order=6)
+            bbox_arr = tools.mort2bbox(mortons)
+            bbox_ref = [tools.mort2bbox(int(m)) for m in mortons]
+            assert bbox_arr == bbox_ref, f"mort2bbox mismatch at n={n}"
+
+            poly_arr = tools.mort2polygon(mortons)
+            poly_ref = [tools.mort2polygon(int(m)) for m in mortons]
+            assert poly_arr == poly_ref, f"mort2polygon mismatch at n={n}"
+
     def test_negative_morton_hemisphere(self):
         """Test that negative morton indices encode polar proximity correctly"""
         # Test high northern latitude - should be positive morton
--- a/mortie/tests/test_threading.py
+++ b/mortie/tests/test_threading.py
@@ -21,7 +21,9 @@ def _run_concurrent(fn, n_threads=8):
     barrier = threading.Barrier(n_threads)
 
     def worker(i):
-        barrier.wait()  # maximize overlap inside the GIL-released region
+        # timeout so a worker that dies before the barrier fails the test red
+        # instead of deadlocking the whole suite
+        barrier.wait(timeout=30)
         results[i] = fn()
 
     threads = [threading.Thread(target=worker, args=(i,)) for i in range(n_threads)]
@@ -45,6 +47,16 @@ def test_morton_coverage_concurrent():
     lons = np.array([46.0, 46.0, 48.0, 48.0])
     expected = mortie.morton_coverage(lats, lons, order=10)
     for got in _run_concurrent(lambda: mortie.morton_coverage(lats, lons, order=10)):
+        np.testing.assert_array_equal(got, expected)
+
+
+def test_morton_coverage_moc_concurrent():
+    # exercises the MOC path, whose binding also touches Python `warnings`
+    # outside the GIL-released region
+    lats = np.array([40.0, 42.0, 42.0, 40.0])
+    lons = np.array([46.0, 46.0, 48.0, 48.0])
+    expected = mortie.morton_coverage_moc(lats, lons, order=12)
+    for got in _run_concurrent(lambda: mortie.morton_coverage_moc(lats, lons, order=12)):
         np.testing.assert_array_equal(got, expected)
 
 
--- a/mortie/tests/test_threading.py
+++ b/mortie/tests/test_threading.py
@@ -69,3 +69,69 @@ def test_morton_buffer_concurrent():
     expected = mortie.morton_buffer(cells, k=1)
     for got in _run_concurrent(lambda: mortie.morton_buffer(cells, k=1)):
         np.testing.assert_array_equal(got, expected)
+
+
+def test_mort2geo_concurrent():
+    # decode path: exercises rust_pix2ang / rust_vec2ang under concurrency
+    cells = mortie.morton_coverage(
+        np.array([40.0, 42.0, 42.0, 40.0]),
+        np.array([46.0, 46.0, 48.0, 48.0]),
+        order=10,
+    )
+    exp_lat, exp_lon = mortie.mort2geo(cells)
+    for got_lat, got_lon in _run_concurrent(lambda: mortie.mort2geo(cells)):
+        np.testing.assert_array_equal(got_lat, exp_lat)
+        np.testing.assert_array_equal(got_lon, exp_lon)
+
+
+def test_linestring_coverage_concurrent():
+    lats = np.array([40.0, 41.0, 42.0])
+    lons = np.array([46.0, 47.0, 48.0])
+
+    def call():
+        return mortie.linestring_coverage(lats, lons, order=10)
+
+    expected = call()
+    for got in _run_concurrent(call):
+        np.testing.assert_array_equal(got, expected)
+
+
+def test_gil_released_during_rust_compute():
+    """Prove the GIL is actually released during compute, not merely that output
+    is uncorrupted (the other tests in this file pass even on unmodified ``main``).
+
+    A pure-Python counter thread spins while a heavy Rust call runs.  With the
+    GIL held for the whole C call (no ``allow_threads``), CPython cannot preempt
+    the call, so the counter cannot advance during it; with the GIL released, the
+    Python thread runs freely and the counter climbs by many thousands.
+    """
+    lats = np.linspace(-80.0, 80.0, 2_000_000)
+    lons = np.linspace(-179.0, 179.0, 2_000_000)
+
+    counter = [0]
+    stop = threading.Event()
+    started = threading.Event()
+
+    def busy():
+        started.set()
+        while not stop.is_set():
+            counter[0] += 1
+
+    b = threading.Thread(target=busy)
+    b.start()
+    started.wait()
+    try:
+        # Bracket the GIL-released region: read the counter immediately before
+        # and after the Rust call from this thread.
+        pre = counter[0]
+        mortie.geo2mort(lats, lons, order=12)
+        progressed = counter[0] - pre
+    finally:
+        stop.set()
+        b.join()
+
+    # GIL held for the whole call -> progressed ~= 0; released -> >> 1000.
+    assert progressed > 1000, (
+        f"Python thread made ~no progress during the Rust compute ({progressed}); "
+        "the GIL was likely held (allow_threads not in effect)"
+    )
--- a/mortie/tests/test_threading.py
+++ b/mortie/tests/test_threading.py
@@ -1,0 +1,59 @@
+"""Concurrency tests for the Rust bindings.
+
+The Rust `#[pyfunction]`s release the GIL around their pure-Rust (often
+rayon-parallel) compute via `py.allow_threads`.  These tests exercise that
+path from many Python threads at once and assert the results are identical to
+a single-threaded run -- i.e. releasing the GIL does not corrupt output or
+crash.  They are deliberately *correctness* tests, not timing tests, so they
+are not flaky under CI load.
+"""
+
+import threading
+
+import numpy as np
+
+import mortie
+
+
+def _run_concurrent(fn, n_threads=8):
+    """Call ``fn`` from ``n_threads`` threads; return the list of results."""
+    results = [None] * n_threads
+    barrier = threading.Barrier(n_threads)
+
+    def worker(i):
+        barrier.wait()  # maximize overlap inside the GIL-released region
+        results[i] = fn()
+
+    threads = [threading.Thread(target=worker, args=(i,)) for i in range(n_threads)]
+    for t in threads:
+        t.start()
+    for t in threads:
+        t.join()
+    return results
+
+
+def test_geo2mort_concurrent():
+    lats = np.linspace(-80.0, 80.0, 5000)
+    lons = np.linspace(-179.0, 179.0, 5000)
+    expected = mortie.geo2mort(lats, lons, order=12)
+    for got in _run_concurrent(lambda: mortie.geo2mort(lats, lons, order=12)):
+        np.testing.assert_array_equal(got, expected)
+
+
+def test_morton_coverage_concurrent():
+    lats = np.array([40.0, 42.0, 42.0, 40.0])
+    lons = np.array([46.0, 46.0, 48.0, 48.0])
+    expected = mortie.morton_coverage(lats, lons, order=10)
+    for got in _run_concurrent(lambda: mortie.morton_coverage(lats, lons, order=10)):
+        np.testing.assert_array_equal(got, expected)
+
+
+def test_morton_buffer_concurrent():
+    cells = mortie.morton_coverage(
+        np.array([40.0, 42.0, 42.0, 40.0]),
+        np.array([46.0, 46.0, 48.0, 48.0]),
+        order=10,
+    )
+    expected = mortie.morton_buffer(cells, k=1)
+    for got in _run_concurrent(lambda: mortie.morton_buffer(cells, k=1)):
+        np.testing.assert_array_equal(got, expected)
--- a/src_rust/src/coverage.rs
+++ b/src_rust/src/coverage.rs
@@ -19,6 +19,7 @@ use std::collections::BinaryHeap;
 use std::f64::consts::PI;
 
 use rayon::prelude::*;
+use smallvec::SmallVec;
 
 use crate::cell_geom::{cell_center_vec, cell_corners, Cap};
 use crate::geo2mort::{ang2pix_scalar, boundaries_step_scalar};
@@ -280,7 +281,10 @@ struct Node {
     corners: [Vec3; 4],
     cos_cr: f64,
     fill: bool,
-    relevant: Vec<usize>,
+    /// Polygon edge indices whose caps reach the cell.  Inline-stored for the
+    /// common ≤8-edge case to avoid a heap allocation per descent node, with
+    /// transparent heap spill beyond.
+    relevant: SmallVec<[usize; 8]>,
 }
 
 /// Build a base-cell node, or `None` if the base cell is entirely outside the
@@ -299,7 +303,7 @@ fn base_node(
     if dot(&cap.axis, &center).clamp(-1.0, 1.0).acos() > cap.radius + cr {
         return None;
     }
-    let relevant: Vec<usize> = (0..edges.len())
+    let relevant: SmallVec<[usize; 8]> = (0..edges.len())
         .filter(|&i| edge_relevant(&edges[i], &center, cos_cr, sin_cr))
         .collect();
     let fill = parity_filled(&center, rings, backend);
@@ -419,7 +423,7 @@ fn node_children(node: &Node, edges: &[Edge]) -> Vec<Node> {
             let corners = cell_corners(depth, pixel);
             let center = cell_center_vec(depth, pixel);
             let (cos_cr, sin_cr) = cell_cos_radius(&center, &corners);
-            let relevant: Vec<usize> = node
+            let relevant: SmallVec<[usize; 8]> = node
                 .relevant
                 .iter()
                 .copied()
--- a/src_rust/src/coverage.rs
+++ b/src_rust/src/coverage.rs
@@ -24,7 +24,7 @@ use crate::cell_geom::{cell_center_vec, cell_corners, Cap};
 use crate::geo2mort::{ang2pix_scalar, boundaries_step_scalar};
 use crate::morton::nested2mort;
 use crate::sphere::{
-    arcs_cross, choose_backend, dot, latlon_to_unit_vec, normalize, parity_filled, Vec3,
+    arcs_cross_n, choose_backend, cross, dot, latlon_to_unit_vec, normalize, parity_filled, Vec3,
 };
 
 // ── public entry points ──────────────────────────────────────────────────
@@ -195,6 +195,9 @@ fn build_ring(lats: &[f64], lons: &[f64]) -> Vec<Vec3> {
 struct Edge {
     a: Vec3,
     b: Vec3,
+    /// Great-circle normal `a × b`, precomputed once so the per-cell crossing
+    /// tests reduce to dot products (see [`arcs_cross_n`]).
+    n_ab: Vec3,
     mid: Vec3,
     cos_rho: f64,
     sin_rho: f64,
@@ -221,6 +224,7 @@ fn build_edges(rings: &[Vec<Vec3>], order: u8) -> Vec<Edge> {
             edges.push(Edge {
                 a,
                 b,
+                n_ab: cross(&a, &b),
                 mid,
                 cos_rho,
                 sin_rho,
@@ -250,9 +254,13 @@ fn edge_relevant(e: &Edge, center: &Vec3, cos_cr: f64, sin_cr: f64) -> bool {
 /// Flips the even-odd fill state between two nearby points.
 #[inline]
 fn arc_crossing_parity(p: &Vec3, q: &Vec3, relevant: &[usize], edges: &[Edge]) -> bool {
+    // The probe arc p→q is fixed across the whole edge fan, so compute its
+    // great-circle normal once; each edge already carries its own (`n_ab`).
+    let n_pq = cross(p, q);
     let mut crossings = 0u32;
     for &i in relevant {
-        if arcs_cross(p, q, &edges[i].a, &edges[i].b) {
+        let e = &edges[i];
+        if arcs_cross_n(p, q, &n_pq, &e.a, &e.b, &e.n_ab) {
             crossings += 1;
         }
     }
@@ -265,6 +273,12 @@ struct Node {
     pixel: u64,
     depth: u8,
     center: Vec3,
+    /// The cell's four corners and the cosine of its circumradius, cached at
+    /// construction (both are computed there anyway to cull edges).  Caching
+    /// lets the straddle and radius tests reuse them instead of refetching
+    /// corners from HEALPix — the descent hot loop.
+    corners: [Vec3; 4],
+    cos_cr: f64,
     fill: bool,
     relevant: Vec<usize>,
 }
@@ -289,7 +303,7 @@ fn base_node(
         .filter(|&i| edge_relevant(&edges[i], &center, cos_cr, sin_cr))
         .collect();
     let fill = parity_filled(&center, rings, backend);
-    Some(Node { pixel: base, depth: 0, center, fill, relevant })
+    Some(Node { pixel: base, depth: 0, center, corners, cos_cr, fill, relevant })
 }
 
 /// Does this cell sit close enough to a pole that its HEALPix edges deviate
@@ -357,11 +371,21 @@ fn node_straddles(node: &Node, edges: &[Edge], order: u8) -> bool {
         return true;
     }
 
-    // (2) cheap 4-corner geodesic-quad straddle test (the common path).
-    let corners = cell_corners(node.depth, node.pixel);
+    // (2) cheap 4-corner geodesic-quad straddle test (the common path).  Corners
+    // are cached on the node; precompute the four cell-edge normals once so each
+    // edge-vs-edge test is dot-products only (`arcs_cross_n`).
+    let corners = &node.corners;
+    let n_quad: [Vec3; 4] = [
+        cross(&corners[0], &corners[1]),
+        cross(&corners[1], &corners[2]),
+        cross(&corners[2], &corners[3]),
+        cross(&corners[3], &corners[0]),
+    ];
     let quad_straddles = node.relevant.iter().any(|&i| {
         let e = &edges[i];
-        (0..4).any(|ci| arcs_cross(&e.a, &e.b, &corners[ci], &corners[(ci + 1) % 4]))
+        (0..4).any(|ci| {
+            arcs_cross_n(&e.a, &e.b, &e.n_ab, &corners[ci], &corners[(ci + 1) % 4], &n_quad[ci])
+        })
     }) || corners
         .iter()
         .any(|c| arc_crossing_parity(&node.center, c, &node.relevant, edges));
@@ -377,9 +401,10 @@ fn node_straddles(node: &Node, edges: &[Edge], order: u8) -> bool {
     }
     let bnd = boundaries_step_scalar(node.depth, node.pixel, near_pole_step(node.depth));
     let n = bnd.len();
+    let n_bnd: Vec<Vec3> = (0..n).map(|ci| cross(&bnd[ci], &bnd[(ci + 1) % n])).collect();
     node.relevant.iter().any(|&i| {
         let e = &edges[i];
-        (0..n).any(|ci| arcs_cross(&e.a, &e.b, &bnd[ci], &bnd[(ci + 1) % n]))
+        (0..n).any(|ci| arcs_cross_n(&e.a, &e.b, &e.n_ab, &bnd[ci], &bnd[(ci + 1) % n], &n_bnd[ci]))
     }) || bnd
         .iter()
         .any(|b| arc_crossing_parity(&node.center, b, &node.relevant, edges))
@@ -401,16 +426,14 @@ fn node_children(node: &Node, edges: &[Edge]) -> Vec<Node> {
                 .filter(|&i| edge_relevant(&edges[i], &center, cos_cr, sin_cr))
                 .collect();
             let fill = node.fill ^ arc_crossing_parity(&node.center, &center, &node.relevant, edges);
-            Node { pixel, depth, center, fill, relevant }
+            Node { pixel, depth, center, corners, cos_cr, fill, relevant }
         })
         .collect()
 }
 
 /// A cell's angular radius (centre→corner), in radians.
 fn node_radius(node: &Node) -> f64 {
-    let corners = cell_corners(node.depth, node.pixel);
-    let (cos_cr, _) = cell_cos_radius(&node.center, &corners);
-    cos_cr.clamp(-1.0, 1.0).acos()
+    node.cos_cr.clamp(-1.0, 1.0).acos()
 }
 
 /// Top-down descent producing the covering cells as `(nested_pixel, depth)`.
--- a/src_rust/src/lib.rs
+++ b/src_rust/src/lib.rs
@@ -114,16 +114,18 @@ fn fast_norm2mort<'py>(
         return Err(PyValueError::new_err("Max order is 18 (to output to 64-bit int)."));
     }
 
-    // Parallel computation using rayon
-    let results: Vec<i64> = (0..max_len)
-        .into_par_iter()
-        .map(|i| {
-            let order_val = order_arr[if order_arr.len() == 1 { 0 } else { i }];
-            let normed_val = normed_arr[if normed_arr.len() == 1 { 0 } else { i }];
-            let parents_val = parents_arr[if parents_arr.len() == 1 { 0 } else { i }];
-            morton::fast_norm2mort_scalar(order_val, normed_val, parents_val)
-        })
-        .collect();
+    // Parallel computation using rayon (GIL released for the pure-Rust region)
+    let results: Vec<i64> = py.allow_threads(|| {
+        (0..max_len)
+            .into_par_iter()
+            .map(|i| {
+                let order_val = order_arr[if order_arr.len() == 1 { 0 } else { i }];
+                let normed_val = normed_arr[if normed_arr.len() == 1 { 0 } else { i }];
+                let parents_val = parents_arr[if parents_arr.len() == 1 { 0 } else { i }];
+                morton::fast_norm2mort_scalar(order_val, normed_val, parents_val)
+            })
+            .collect()
+    });
 
     // Return as NumPy array
     Ok(results.into_pyarray_bound(py).into_any().unbind())
@@ -140,7 +142,7 @@ fn split_children_rust(
     max_depth: Option<usize>,
 ) -> PyResult<PyObject> {
     let data = morton_array.to_vec()?;
-    let flat = prefix_trie::split_children_flat(&data, max_depth);
+    let flat = py.allow_threads(|| prefix_trie::split_children_flat(&data, max_depth));
 
     // Convert Vec<FlatNode> to a Python list of tuples
     let py_list = pyo3::types::PyList::empty_bound(py);
@@ -214,14 +216,16 @@ fn rust_geo2mort<'py>(
         ));
     }
 
-    let results: Vec<i64> = (0..max_len)
-        .into_par_iter()
-        .map(|i| {
-            let lat = lat_arr[if lat_arr.len() == 1 { 0 } else { i }];
-            let lon = lon_arr[if lon_arr.len() == 1 { 0 } else { i }];
-            geo2mort::geo2mort_scalar(lat, lon, order)
-        })
-        .collect();
+    let results: Vec<i64> = py.allow_threads(|| {
+        (0..max_len)
+            .into_par_iter()
+            .map(|i| {
+                let lat = lat_arr[if lat_arr.len() == 1 { 0 } else { i }];
+                let lon = lon_arr[if lon_arr.len() == 1 { 0 } else { i }];
+                geo2mort::geo2mort_scalar(lat, lon, order)
+            })
+            .collect()
+    });
 
     Ok(results.into_pyarray_bound(py).into_any().unbind())
 }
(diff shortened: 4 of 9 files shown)


mortie/tests/test_threading.py
```diff
@@ -0,0 +1,59 @@
+"""Concurrency tests for the Rust bindings.
+
+The Rust `#[pyfunction]`s release the GIL around their pure-Rust (often
+rayon-parallel) compute via `py.allow_threads`.  These tests exercise that
+path from many Python threads at once and assert the results are identical to
+a single-threaded run -- i.e. releasing the GIL does not corrupt output or
+crash.  They are deliberately *correctness* tests, not timing tests, so they
+are not flaky under CI load.
+"""
```
espg, line 9: 🤖 *from Claude (review)*

**Test is largely vacuous w.r.t. the change it's meant to guard.** These are *correctness*-only assertions (each thread's result == single-threaded result). But result-correctness here doesn't depend on `allow_threads`: with or without the GIL release, every call still copies inputs into owned `Vec`s before the parallel region and returns a fresh array — so these exact assertions **also pass on unmodified `main`** (no `allow_threads` at all). The test therefore can't distinguish "GIL released correctly" from "GIL never released," and would not catch a regression that, say, reverted one binding's wrapping.

That's not nothing — it's a real crash/data-corruption smoke test under concurrent load, and the docstring is honest that it's "correctness not timing." But if the intent is to *guard the new path*, consider adding one assertion that actually demonstrates GIL release, e.g. measure that N concurrent calls on a large input complete in meaningfully less than N× the single-threaded time (with generous slack to avoid CI-load flakiness), or assert wall-clock overlap. Otherwise the suite green-lights even a full revert of the PR. Flagging for @espg to decide whether a timing-based assertion is worth the flake risk you deliberately avoided.

---

mortie/tests/test_threading.py
```diff
@@ -0,0 +1,59 @@
...
+import numpy as np
+
+import mortie
+
+
+def _run_concurrent(fn, n_threads=8):
+    """Call ``fn`` from ``n_threads`` threads; return the list of results."""
+    results = [None] * n_threads
+    barrier = threading.Barrier(n_threads)
+
```
espg, line 22: 🤖 *from Claude (review)*

**`barrier.wait()` has no timeout — a failure before the barrier hangs the whole suite.** All N workers must reach `barrier.wait()` for any to proceed. If the call under test (or anything between `start()` and the barrier) raises in one worker, that thread dies without tripping the barrier, the remaining N-1 block forever on `wait()`, and the main thread blocks forever on `join()` — pytest hangs rather than failing, with no traceback. Cheap hardening: pass a timeout (`threading.Barrier(n_threads, timeout=30)`), which raises `BrokenBarrierError` in the survivors instead of deadlocking. Minor, but it's the difference between a clean red and a stuck CI job.

Separately: exceptions raised inside `worker` (after the barrier) are swallowed by the `threading.Thread` machinery — `results[i]` stays `None` and `assert_array_equal(None, expected)` will surface it, so that path is at least visible. The pre-barrier hang is the real gap.

---

mortie/tests/test_threading.py
```diff
@@ -0,0 +1,59 @@
...
+def test_morton_buffer_concurrent():
+    cells = mortie.morton_coverage(
+        np.array([40.0, 42.0, 42.0, 40.0]),
+        np.array([46.0, 46.0, 48.0, 48.0]),
+        order=10,
+    )
+    expected = mortie.morton_buffer(cells, k=1)
+    for got in _run_concurrent(lambda: mortie.morton_buffer(cells, k=1)):
+        np.testing.assert_array_equal(got, expected)
```
espg, line 71: 🤖 *from Claude (review)*

**Coverage gap: 3 of 15 wrapped bindings are exercised.** The PR wraps all 15 `#[pyfunction]`s, but only `geo2mort` (→ `rust_geo2mort`), `morton_coverage` (→ `rust_polygon_coverage`), and `morton_buffer` (→ `rust_morton_buffer`) are tested here. Notably untested under concurrency:

- the **MOC + `warnings` path** (`rust_polygon_coverage_moc` / `rust_multipolygon_coverage_moc`) — this is the one place where a Python call (`py.import_bound("warnings")`) must stay *outside* `allow_threads`; it's also the highest-risk wrapping and gets zero concurrent coverage. A test hitting the `max_cells`-too-small warning from many threads would actually exercise the GIL-reacquire-then-warn boundary.
- `rust_multipolygon_coverage`, `rust_vec2ang`, `rust_pix2ang`, `rust_ang2pix`, `rust_boundaries`, `rust_linestring_coverage`, `rust_moc_normalize`, `rust_moc_to_order`, `split_children_rust`, `fast_norm2mort`.

Given `_run_concurrent` already abstracts the harness, adding a few more `lambda`s is cheap. Since there's no `--cov-fail-under` gate in `pyproject.toml`, nothing forces this — flagging so @espg can decide how much breadth is warranted for a "smoke test" tier.

---

src_rust/src/lib.rs
```diff
@@ -564,7 +579,7 @@ fn rust_polygon_coverage_moc(
         } else {
             (coverage::polygon_to_morton_moc(&lat_data, &lon_data, order), None)
         }
-    });
+    }));
```
espg, line 582: 🤖 *from Claude (review)*

Confirmed correct (no change requested): the `py.import_bound("warnings")` / `warn(...)` call stays in the post-compute `match` arm at L587-594, **outside** `allow_threads` — so the GIL is held when it runs. Same for `rust_multipolygon_coverage_moc` at L671. The closure here captures only `tolerance`/`max_cells`/`order` (Copy) and `&lat_data`/`&lon_data` (owned `Vec`s extracted at L565-566), all `Send + Ungil`; no `Bound`/`Py`/`Python` crosses the boundary. The `catch_unwind` returning `(Vec, Option<...>)` is `Send`, so moving it inside is fine. The one nit is purely stylistic: the `py.allow_threads(|| std::panic::catch_unwind(|| { ... }))` double-closure on one line (L568) reads a bit denser than the multi-line form used in the multipolygon variant at L662-666 — not worth churning, just noting the two MOC functions now format the same wrapping slightly differently.

---

src_rust/src/lib.rs
```diff
@@ -671,15 +690,17 @@ fn rust_multipolygon_coverage_moc(
 #[pyfunction]
 fn rust_moc_normalize(py: Python<'_>, morton: PyReadonlyArray1<i64>) -> PyResult<PyObject> {
     let data = morton.to_vec()?;
-    Ok(moc::normalize(&data).into_pyarray_bound(py).into_any().unbind())
+    let normalized = py.allow_threads(|| moc::normalize(&data));
```
espg, line 693: 🤖 *from Claude (review)*

Minor / judgment call: `rust_moc_normalize` and `rust_moc_to_order` are **serial** (`moc::normalize` / `moc::to_order` are not rayon-parallel, unlike the `into_par_iter` bindings). For small MOC inputs the GIL release+reacquire round-trip can cost more than the compute it overlaps. `allow_threads` is cheap (no thread spawn — just lock state toggling), so this is unlikely to regress anything measurable, and uniform wrapping is defensible for consistency. But it does mean the PR's "release the GIL across *all* bindings" framing includes a couple where the win is marginal-to-negative on the common small-input case. Not a blocker — flagging since the PR is explicitly a perf change and #34 is a perf audit; @espg may want to confirm these two are worth it or leave them serial+GIL-held.

(Note: `fast_norm2mort`'s all-scalar fast path at L62-74 correctly stays *un*wrapped — good, that's the trivially-small case where wrapping would be pure overhead.)

---

espg (commented):
🤖 *from Claude (review)*

Adversarial fresh-eyes review of the `allow_threads` wrapping. **No correctness defects found in the Rust change** — the wrapping is sound. The findings are about test quality and a couple of judgment calls, not bugs.

**What's correct (verified, no action needed):**
- All **15** `#[pyfunction]`s registered in the `_rustie` pymodule (lib.rs L751-765) are wrapped — scope matches "#5: release GIL across all bindings." Nothing missed.
- Every `allow_threads` closure captures only owned `Vec`s (extracted via `.to_vec()` *before* the wrapped region), `Copy` scalars, or `&Vec` — all `Send + Ungil`. No `Bound`/`Py`/`Python` crosses the boundary, so this compiles and is memory-safe.
- The subtle bit is handled right: the Python `warnings.warn` calls in both MOC functions stay **outside** `allow_threads`, in the post-compute match arm (lib.rs L587, L671), where the GIL is held.
- `catch_unwind` correctly moved inside `allow_threads`; the returned `std::thread::Result` is `Send`. No unwind-across-FFI concern — the panic is still caught in Rust and converted to `PyValueError` before any boundary is recrossed.
- lib.rs is 767 lines, under the ~1000-line limit. Style matches surrounding code.

**Findings (inline):**
1. **Tests are largely vacuous w.r.t. the change** (`test_threading.py`): correctness-only assertions pass identically on unmodified `main` with no `allow_threads`, so the suite would green-light a full revert. Useful as a concurrent crash/corruption smoke test, but doesn't prove GIL release. Consider one timing/overlap assertion if guarding the path matters.
2. **`barrier.wait()` has no timeout** — a failure before the barrier deadlocks the suite (pytest hangs, no traceback) instead of failing red.
3. **Coverage gap**: only 3 of 15 bindings exercised; the highest-risk one (MOC + `warnings` GIL-reacquire path) and the multipolygon/vec2ang/etc. variants get no concurrent coverage. No `--cov-fail-under` gate, so nothing forces it.
4. **Marginal wins**: `rust_moc_normalize` / `rust_moc_to_order` are serial (not rayon), so GIL release/reacquire may not pay off on small inputs — judgment call, not a bug.

**Scope note:** I confirmed via the GitHub API (`changed_files: 2`) and `git show 496fb42` that this PR contains only `lib.rs` + `test_threading.py`. The relicense (BSD→MIT), the new `.github/workflows/lint.yml`, and the ruff config in `pyproject.toml` that appear in a local `git diff main...HEAD` are **already on the base** (`main` @ `2e4deac`), **not** part of this PR — so they're out of scope here and I'm *not* flagging them against #41. (If the local `main` ref looked like it diverged, that's a stale-ref artifact, not this PR introducing workflow/license changes.)

Net: the implementation is clean and does what #34 item #5 asks. The actionable gap is test strength, not the Rust. Leaving for @espg to adjudicate.

---

espg:
🤖 *from Claude*

A fresh-context adversarial review was run against this diff; its findings are inline above for your adjudication. It confirmed the Rust change is correct (all 15 registered bindings wrapped, every closure captures only owned/`Copy`/`&Vec` data so it's `Send + Ungil`, the Python `warnings` import correctly stays outside `allow_threads` in both MOC paths, `catch_unwind` safely inside).

I acted on the one clearly-correct trivial defect and left the judgment calls standing:

- **Fixed (commit `9e88f75`):** the `barrier.wait()` had no timeout — a worker dying before the barrier would deadlock the whole suite instead of failing red. Added `timeout=30`. Also added a concurrent case for `morton_coverage_moc` (the highest-risk path, since its binding also touches Python `warnings` just outside the GIL-released region).
- **Left for your call (standing review comments):** (1) the tests are correctness-only and would also pass on unmodified `main`, so they don't *prove* the GIL is released — I avoided a timing-based assertion deliberately because it flakes under CI load; (2) breadth — still only 4 of 15 bindings under concurrent test. Both are test-philosophy judgments rather than defects, so I've left them as open review feedback rather than resolving them.

---

espg:
> (1) the tests are correctness-only and would also pass on unmodified main, so they don't prove the GIL is released

That's fine

> breadth — still only 4 of 15 bindings under concurrent test. 

Why only 4 or 15 bindings? Do you mean only four bindings test, or only four bindings test concurrency?

espg:
> Consider one timing/overlap assertion if guarding the path matters.

this sounds reasonable and cheap to add

espg:
🤖 *from Claude*

Pushed `149f61e` addressing both of your points.

> Why only 4 of 15 bindings? Do you mean only four bindings test, or only four bindings test concurrency?

The latter — **only four bindings were tested *under concurrency*** (8 threads hitting the same call). All 15 registered `#[pyfunction]`s already have **functional (single-threaded) coverage** in the rest of the suite; the gap the review flagged was purely *concurrency* breadth. The original four were the highest-risk: `geo2mort`, `morton_coverage`, `morton_coverage_moc` (the `warnings`-GIL-reacquire path), and `morton_buffer` — all rayon-parallel.

I've now broadened concurrent coverage to **six** bindings, adding the decode path and linestring:
- `test_mort2geo_concurrent` — exercises `rust_pix2ang` / `rust_vec2ang` under threads.
- `test_linestring_coverage_concurrent` — exercises `rust_linestring_coverage`.

(The remaining serial bindings — `rust_moc_normalize`/`rust_moc_to_order`/`rust_ang2pix`/`fast_norm2mort`/`split_children_rust`/the multipolygon variants — are single-threaded-tested elsewhere; happy to add concurrent cases for any you specifically want guarded, but they're the lower-risk, non-rayon ones.)

> Consider one timing/overlap assertion if guarding the path matters — this sounds reasonable and cheap to add

Added `test_gil_released_during_rust_compute`, which actually **proves the GIL is released** (the other tests pass even on unmodified `main`, as you noted). A pure-Python counter thread spins while a heavy `geo2mort` (2M points) runs; the test brackets the call and asserts the counter advanced by `> 1000` during it:

```python
pre = counter[0]
mortie.geo2mort(lats, lons, order=12)
progressed = counter[0] - pre
assert progressed > 1000
```

This isn't a flaky wall-time ratio — it's presence/absence. With the GIL **held** for the whole C call (no `allow_threads`), CPython cannot preempt it, so the Python thread is frozen and `progressed ≈ 0`; with it **released**, the thread runs freely and `progressed` is in the hundreds of thousands. The huge separation makes it robust to CI load. It would fail on a revert of this PR — closing the "tests would green-light a full revert" gap.

**Testing:** `maturin develop --release`, then `pytest mortie/tests/test_threading.py` → **7 passed**; full suite **226 passed, 8 skipped**; `flake8 --select=E9,F63,F7,F82` clean.

**Status:** your test feedback is addressed. On open-question 1 (fold the descent hot-loop wins #1/#2/#3 into this branch vs separate PRs — you left it to me): I'll take the next perf phase (**#1, cache cell corners/circumradius on `Node`**) as a follow-up commit on this branch in the next run, since the audit ranks it the biggest descent win and it's `coverage.rs`-local. Leaving this PR `implement` (continuing next run) rather than `waiting` — shout if you'd rather those land as separate PRs.

---

espg (commented):
🤖 *from Claude (review)*

Reviewed `149f61e` (test-only: `mortie/tests/test_threading.py`). This commit addresses both pieces of feedback well — the two new concurrency tests are correct, and the GIL-release test is genuinely load-bearing (it fails on unmodified `main` and passes here). No blocking issues; one robustness caveat worth a note.

**GIL-release assertion — robust.** The bracketing is correct: `pre`/`progressed` are read by the *main* thread immediately around `geo2mort`, so the window is defined by the main thread's own execution position (blocked inside the Rust call), not by wall-clock around the busy thread — busy-thread iterations outside the call can't leak in. The `>1000` gate has huge margin both ways: GIL-held → `progressed ≈ 0` (no false pass), GIL-released → a 2M-pt order=12 `geo2mort` runs long enough for a Python `while` loop to climb into the hundreds of thousands (no false fail). `counter[0] += 1` is fine: only the busy thread writes it; the main thread only reads it twice; the GIL keeps the list access atomic, so there's no race.

**Outputs compared correctly.** `mort2geo` returns a `(lat, lon)` tuple — both arrays unpacked and compared. `linestring_coverage` returns an ndarray — compared with `assert_array_equal`. Both right.

**Style.** Matches the surrounding helpers; no `# noqa`, no E9/syntax issues, line lengths within 88.

One non-blocking inline note on the core-count dependency below.

---

espg:
🤖 *from Claude (review)*

Follow-up to my review of `149f61e` — the one non-blocking caveat on `test_gil_released_during_rust_compute`:

**Core-count dependency.** The test needs the busy thread to actually run *concurrently* with the GIL-released Rust thread. On GitHub-hosted runners (2+ vCPU) this holds comfortably. On a single-core runner — or one CPU-pinned via `taskset`/cgroups — the OS must time-slice the two OS threads even with the GIL released; the busy thread would still advance (a GIL-released order=12 2M-pt `geo2mort` runs long enough that it should clear `>1000` even on one core), but it's the least-robust axis and the failure mode would be a confusing red rather than a clear signal. Not blocking, and I wouldn't gate the PR on it — just flagging it as the one place the `>1000` margin is environment-dependent rather than huge. If single-core CI ever becomes a thing, a `pytest.mark.skipif(os.cpu_count() < 2)` guard would make the intent explicit. Leaving for @espg to decide.

---

espg:
...if it's not blocking, why did you stop implementing??

> ## Phases — the "before #35" perf cluster (#34)
>
- [x] **#5** release the GIL across all bindings *(this PR)*
- [ ] **#1** cache cell corners / circumradius on `Node` in the coverage descent
- [ ] **#2** precompute per-edge great-circle normal `cross(a,b)` on `Edge`
- [ ] **#3** `SmallVec` for `Node.relevant`
- [ ] **#10** batch `vec2ang` + vectorize antimeridian fix-up in `mort2bbox`/`mort2polygon`
- [ ] **#6** shift/mask + `ilog10` micro-wins

Are all the above items done yet??

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the phase 2/3/4 commits (`149f61e..HEAD`). I rebuilt (`maturin develop --release`), ran `cargo test` (86 pass, incl. the new `test_arcs_cross_n_matches_arcs_cross`), `cargo clippy`, and `pytest` (227 pass / 8 skipped). No blocking defects found — the work is solid. Details below.

**Phase 2 (output-invariant) — verified.** The corner/`cos_cr` cache is correct: each `Node` caches *its own* corners, recomputed per node in both `base_node` (coverage.rs:296) and `node_children` (coverage.rs:419), so no stale-corner aliasing. `node_radius` now reads `node.cos_cr`, which is exactly the value `cell_cos_radius` produced at construction — same number, no recompute. `arcs_cross_n` is the dot-product form of `arcs_cross` (`orient(a,b,x) == (a×b)·x`), and full coverage output was byte-identical across the suite.

**Phase 3 (shift/mask) — bit-exact for all valid inputs.** Max order is 18 (lib.rs:113/181, morton.rs:68), so `shift = 2*order ≤ 36` — no overflow in either `1_i64 << shift` (geo2mort) or `1u64 << shift` (morton). Both `nest` (HEALPix hash ≥ 0) and `nested` (u64) are non-negative, so `>>`/`&` reproduce `/`/`%` exactly, and `<<`/`|` reproduce `*`/`+` exactly (no carry: `normed` occupies exactly the low `shift` bits). One subtle behavior change noted inline on `decimal_digit_count`.

**Phase 4 (batched vec2ang + bbox fix) — correct, and the fix is real.** Confirmed against the Rust source: multi-cell `boundaries` is shape `(N, 3, ncols)` with the cell on **axis 0** (lib.rs:380 `(n, 3, 4)`, lib.rs:411 `(n, 3, ncols)`). So the old `mort2bbox` `boundaries[:, :, i]` was genuinely indexing the wrong axis (it only dodged a shape error at N==4, the corner count) — the new `np.transpose(boundaries,(0,2,1)).reshape(-1,3)` correctly reproduces the per-cell `boundaries[i].T`, and the regression test catches it. Good catch.

Inline notes are nits / judgment calls, not blockers.

---

src_rust/src/morton.rs
```diff
@@ -173,17 +176,8 @@ fn decimal_digit_count(val: u64) -> usize {
...
-    while val >= threshold {
-        count += 1;
-        if count >= 19 {
-            break;
-        }
-        threshold *= 10;
-    }
-    count
+    // `ilog10` is a single intrinsic (stable since 1.67); digits = floor(log10)+1.
+    val.ilog10() as usize + 1
```
espg, line 180: 🤖 *from Claude (review)*

Nit / behavior-change note (non-blocking). For all *valid* morton indices this is bit-identical to the old loop — at order ≤ 18 the value has ≤ 19 digits, so `ilog10()+1 ≤ 19`, and the old loop's `count >= 19` cap was never reached, so parity holds for the contract.

The one divergence is on *invalid* >19-digit input: the old loop **capped** `digit_count` at 19 (→ `order_usize = 18`, a valid `POWERS_OF_10` index, returning garbage), whereas `ilog10()` on a 20-digit `u64` returns 19 → `order = 19` → `POWERS_OF_10[19]` panics at morton.rs:121 (the array has 19 entries, indices 0–18). That's arguably an improvement (panic over silent garbage), but it *is* a behavior change outside the "bit-exact" envelope, and `mort2nested` has no explicit `order > 18` guard. Worth a one-line mention in the PR body, or a guard if you'd rather keep the old non-panicking behavior.

---

src_rust/src/sphere.rs
```diff
@@ -91,6 +91,26 @@ pub fn arcs_cross(a: &Vec3, b: &Vec3, c: &Vec3, d: &Vec3) -> bool {
...
+pub fn arcs_cross_n(a: &Vec3, b: &Vec3, n_ab: &Vec3, c: &Vec3, d: &Vec3, n_cd: &Vec3) -> bool {
+    let d1 = dot(n_ab, c);
+    let d2 = dot(n_ab, d);
+    if (d1 > 0.0) == (d2 > 0.0) {
+        return false; // c, d on the same side of AB
+    }
+    let d3 = dot(n_cd, a);
+    let d4 = dot(n_cd, b);
+    (d3 > 0.0) != (d4 > 0.0) // a, b on opposite sides of CD
+}
```
espg, line 112: 🤖 *from Claude (review)*

Judgment call (non-blocking) — the float-sign degeneracy vs `arcs_cross`. `arcs_cross` computes `orient(a,b,c) = dot(a, cross(b,c))` while `arcs_cross_n` computes `dot(cross(a,b), c)`. These are the same scalar triple product `det[a b c]` in exact arithmetic, but FP rounding can flip the *sign* near a tie (det ≈ 0, i.e. a grazing/coplanar touch). The new unit test correctly skips those (`|orient| < 1e-9`), so the test only asserts agreement away from ties — which is honest but means the substitution is *not* proven identical exactly where the two could differ.

I think skipping is safe in the real descent, for the reason `arcs_cross` already documents (sphere.rs:81–82): an exactly-touching arc reports `false` and is paired with the vertex-in-cell check (`edges[i].leaf >> shift == node.pixel`, coverage.rs:370) plus the center→corner parity recheck. So a sign flip at a tie can't silently drop a real straddle. I verified the full coverage suite output is byte-identical after the swap, which backs this up. Flagging only so you can confirm the polar/antimeridian coverage tests actually exercise a near-tie cell edge; if they don't, that's the one residual gap.

---

mortie/tests/test_mort_inverse.py
```diff
@@ -112,6 +112,26 @@ def test_array_input(self):
...
+        for n in (2, 3, 4, 5):
+            mortons = tools.geo2mort(lats[:n], lons[:n], order=6)
+            bbox_arr = tools.mort2bbox(mortons)
+            bbox_ref = [tools.mort2bbox(int(m)) for m in mortons]
+            assert bbox_arr == bbox_ref, f"mort2bbox mismatch at n={n}"
+
+            poly_arr = tools.mort2polygon(mortons)
+            poly_ref = [tools.mort2polygon(int(m)) for m in mortons]
+            assert poly_arr == poly_ref, f"mort2polygon mismatch at n={n}"
+
```
espg, line 134: 🤖 *from Claude (review)*

Good regression test — and nicely chosen: `n=4` is exactly the length that masked the old `boundaries[:, :, i]` shape bug (cell count == corner count), and you include an antimeridian cell (lon 179.9 / -179.9) so the voting branch in `mort2bbox` is exercised.

Minor coverage gap (non-blocking): `mort2polygon` is only tested at the default `step=1` here. The phase-4 reshape (`np.transpose((0,2,1)).reshape(n, ncols)`) is `step`-dependent (`ncols = 4*step`), and the `step > 1` multi-cell path is a *different* Rust branch (lib.rs:394–413). Consider one extra assertion with `step=2` (or 3) over the same array to lock down that the reshape is correct for `ncols != 4`. The code looks right to me — `(N, 3, ncols)` → transpose → reshape reproduces per-cell `boundaries[i].T` for any `ncols` — but it's currently untested.

---

src_rust/src/coverage.rs
```diff
@@ -377,9 +401,10 @@ fn node_straddles(node: &Node, edges: &[Edge], order: u8) -> bool {
     }
     let bnd = boundaries_step_scalar(node.depth, node.pixel, near_pole_step(node.depth));
     let n = bnd.len();
+    let n_bnd: Vec<Vec3> = (0..n).map(|ci| cross(&bnd[ci], &bnd[(ci + 1) % n])).collect();
```
espg, line 408: 🤖 *from Claude (review)*

Nit (non-blocking). This near-pole densified-boundary path allocates two `Vec`s per call — `bnd` (already there) and now `n_bnd` (the per-segment normals). It's correctly gated behind `near_pole_curved`, so it's off the common path and the allocation cost is negligible, but if you ever want it allocation-free you could fold the `cross` into the inner `arcs_cross_n` call site (compute `cross(&bnd[ci], &bnd[(ci+1)%n])` inline) and drop the `n_bnd` collect — the precompute only pays off when a normal is reused across iterations, which it isn't here (each `n_bnd[ci]` is used once). The quad path above (coverage.rs:378) genuinely reuses normals across `relevant` edges, so the precompute there is well-motivated; this one is borderline. Purely a judgment call — fine to leave as-is for symmetry with the quad path.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the phase 2/3/4 commits (`149f61e..HEAD`). I rebuilt (`maturin develop --release`) and ran `cargo test` (86 pass, incl. the new `test_arcs_cross_n_matches_arcs_cross`), `cargo clippy`, and `pytest` (227 pass / 8 skipped). **No blocking defects** — the work is correct and well-grounded. Four non-blocking inline notes; details below.

**Phase 2 (claimed output-invariant) — verified.** The corner/`cos_cr` cache is sound: each `Node` caches *its own* corners, recomputed per node in both `base_node` (coverage.rs:296) and `node_children` (coverage.rs:419) — no stale-corner aliasing across the descent. `node_radius` now reads `node.cos_cr`, the exact value `cell_cos_radius` produced at construction (same number, just not recomputed). `arcs_cross_n` is the dot-product form of `arcs_cross` (`orient(a,b,x) == (a×b)·x`). Full coverage output was byte-identical across the suite after the swap, confirming invariance in practice.

**Phase 3 (shift/mask + ilog10) — bit-exact for all valid inputs.** Max order is 18 (lib.rs:113/181, morton.rs:68), so `shift = 2*order ≤ 36` — no overflow in `1_i64 << shift` (geo2mort) or `1u64 << shift` (morton). `nest` (HEALPix hash ≥ 0) and `nested` (u64) are non-negative, so `>>`/`&` reproduce `/`/`%` and `<<`/`|` reproduce `*`/`+` exactly. One out-of-envelope behavior change on `decimal_digit_count` for invalid >19-digit input — see inline.

**Phase 4 (batched vec2ang + bbox fix) — correct, and the fix is real.** Confirmed against the Rust source that multi-cell `boundaries` is `(N, 3, ncols)` with the cell on **axis 0** (lib.rs:380/411). So the old `mort2bbox` `boundaries[:, :, i]` did index the wrong axis (only dodging a shape error at `N==4`), and the new `np.transpose((0,2,1)).reshape(-1,3)` correctly reproduces per-cell `boundaries[i].T`. The regression test catches it well (it includes the `n=4` masking case + an antimeridian cell). One coverage gap: `step > 1` isn't exercised — see inline.

**Conventions:** no new dependencies (✓ — `ilog10` is std, stable since 1.67). `coverage.rs` is now 914 lines and `tools.py` 983 — both still under the ~1000-line limit but worth keeping an eye on. The clippy warnings present are all pre-existing in `lib.rs`/`prefix_trie.rs`, none in the diff-touched files.

---

espg:
🤖 *from Claude*

You asked ["why did you stop implementing?? Are all the above items done yet?"](https://github.com/espg/mortie/pull/41#issuecomment-4705019405) — so this run worked the rest of the cluster. **4 of the 5 items are now done** (the 5th needs a dependency call from you). One commit per phase, all on this branch:

| item | phase | commit | what |
|---|---|---|---|
| #1 | 2 | `7d2f906` | cache the 4 corners + cos-circumradius on `Node` (already computed for culling) so `node_straddles`/`node_radius` stop refetching from HEALPix |
| #2 | 2 | `7d2f906` | `arcs_cross_n` (normal-aware crossing: `orient(a,b,x)=(a×b)·x` → a dot product) + precomputed per-edge `n_ab=a×b` on `Edge`; probe-arc normal once per fan, cell-edge normals once per cell → inner crossing tests are dot-products only |
| #6 | 3 | `89909cb` | shift/mask for the power-of-two split in `geo2mort_scalar`/`nested2mort`/`mort2nested`; `decimal_digit_count` → `u64::ilog10` |
| #10 | 4 | `2e7148f` | one batched `vec2ang` over all cells in `mort2bbox`/`mort2polygon` instead of one Rust round-trip per cell — **also fixes a latent bug** (next ¶) |

**Bug fixed in #10:** `mort2bbox` indexed the boundary array on the wrong axis (`boundaries[:, :, i]`) for multi-cell input — it raised `ValueError` for any array length ≠ 3 and returned *wrong* boxes at length 3, which the old array test never caught (it only checked `len`/keys). The batched path reshapes correctly; a new regression test (`test_bbox_polygon_array_matches_scalar`) asserts array results equal the per-element scalar reference at N ∈ {2,3,4,5} incl. an antimeridian cell and (phase 5) `step=2`.

**Tested:** `cargo test --release` **86 passed** (adds an `arcs_cross_n`↔`arcs_cross` parity test); `maturin develop --release` + `pytest` **227 passed / 8 skipped** — same pass set as baseline plus the new test, so phases 2–3 are output-invariant; `flake8 --select=E9,F63,F7,F82` clean. **CI is green** on the branch (test 3.10/3.11/3.12, ruff, benchmarks, all wheel builds).

**Self-review (fresh context, [inline above](https://github.com/espg/mortie/pull/41)):** no blocking defects. I folded the one concrete gap — finding #3 (the `step>1` boundary branch wasn't covered) — into phase 5 (`41e839f`). The other three I left standing with rationale:
- *`ilog10` on >19-digit input* — bit-identical to the old loop for every valid morton (order ≤ 18 ⇒ ≤ 19 digits). It diverges only on *invalid* >19-digit input, where the old code produced silent garbage and the new code panics (clearer failure, on input that can't occur by construction). Not a regression; flagging rather than adding an invalid-input guard.
- *`arcs_cross_n` FP-sign at a tie* — the two formulations can only disagree on an exact-zero triple product (grazing/coplanar), which the descent already treats as `false` + a vertex-in-cell check; the polar (#28) and polar-bulge (#32) regression tests pass and the full coverage output is byte-identical, so the descent's tie handling absorbs it.
- *cold-path `n_bnd` alloc* — a near-pole-only `Vec` whose normals are used once; left for symmetry with the quad path (negligible, off the common path).

### Questions for review
1. **#3 needs your sign-off — it adds a dependency.** `SmallVec<[usize; 8]>` for `Node.relevant` (to kill the per-child `Vec` reallocs the audit flags) means adding the **`smallvec`** crate. Per §4 I won't add it silently. It's tiny, ubiquitous, MIT/Apache-2.0, no transitive deps. **OK to add `smallvec` and land #3?** If you'd rather not, I can skip it or hand-roll a fixed-capacity inline type (more code, less clean).

Marking **`waiting`** on that one decision (and your merge) — the implementable cluster is done. Say the word on `smallvec` and I'll land #3 next run.

---

espg:
> ilog10 on >19-digit input — bit-identical to the old loop for every valid morton (order ≤ 18 ⇒ ≤ 19 digits). It diverges only on invalid >19-digit input, where the old code produced silent garbage and the new code panics (clearer failure, on input that can't occur by construction). Not a regression; flagging rather than adding an invalid-input guard.

what's the impact with #35 here? we're implementing that soon, so order > 19 will soon be a regular occurrence

> SmallVec<[usize; 8]> for Node.relevant (to kill the per-child Vec reallocs the audit flags) means adding the smallvec crate. Per §4 I won't add it silently. It's tiny, ubiquitous, MIT/Apache-2.0, no transitive deps. OK to add smallvec and land item 3? If you'd rather not, I can skip it or hand-roll a fixed-capacity inline type (more code, less clean).

Can you decribe in more detail what this optimization is, how it works, and how it's expected to improve performance?

espg:
🤖 *from Claude*

Answering both in turn.

## Q1 — `ilog10` and #35 (order > 19)

**Short version: this is a non-issue for #35, because #35 deletes the code path that `ilog10` lives on.** No change is warranted on this branch, and no guard is needed — here's the reasoning, grounded in both encodings.

**Where `ilog10` is actually used.** It's only reached via `decimal_digit_count`, which is called in exactly one place — `mort2nested` ([`src_rust/src/morton.rs#L117-L121`](https://github.com/espg/mortie/blob/claude/friendly-volta-mbsm1c/src_rust/src/morton.rs#L117-L121)):

```rust
// Count decimal digits to determine order
let digit_count = decimal_digit_count(abs_val);
let order = (digit_count - 1) as u8;
```

That is, in **today's** encoding the order is *implicit*: a morton index is a base-10 string whose digits are each in `1..=4` (plus a leading base-cell digit), so the order is recovered by counting decimal digits. At order ≤ 18 the index is ≤ 19 digits, and `decimal_digit_count` = `val.ilog10() + 1` ([`#L170-L177`](https://github.com/espg/mortie/blob/claude/friendly-volta-mbsm1c/src_rust/src/morton.rs#L170-L177)) is bit-identical to the old loop across that whole range. `u64::ilog10` is well-defined for *all* `u64` (it's the old loop's `count >= 19` break that mattered, and it never trips for valid input). So on this branch, for every index that can be constructed, `ilog10` and the old loop agree exactly.

**What #35 does to this path.** #35 replaces the decimal-string encoding with the packed-binary layout we settled on the issue thread — `[4-bit base | 54-bit body (27×2-bit) | 6-bit suffix]`, with **order carried *explicitly* in the 6-bit suffix** ([#35 body](https://github.com/espg/mortie/issues/35), [the "fast path read" note](https://github.com/espg/mortie/issues/35#issue) — "read the prefix and suffix to determine base cell and MOC order"). Order is then a mask/shift of the suffix, not a count of anything. There is **no `decimal_digit_count` / `ilog10` in the new datatype** — the whole "infer order from digit count" idea is exactly what the explicit-order suffix removes.

And the `mort2nested`/`nested2mort` bridge that calls `decimal_digit_count` today is itself in scope for #35: in the settled plan, the `decimal_morton ↔ (depth, nested_idx)` bridge is #35 Phase 1, and #34 item #4 ("gut `mort2norm`") "lands here" with golden fixtures replacing the old asserts (see the [#35 plan, Phases 1–2](https://github.com/espg/mortie/issues/35#issuecomment-4701428087)). So #35 *rewrites* `mort2nested`, and `decimal_digit_count` goes away with it.

**Net.** The "order > 19 becomes regular" concern is real for the *datatype*, but it lands on the **new suffix-based order field, which never touches `ilog10`** — and the digit-count path that uses `ilog10` is deleted by #35, not extended past 19 digits. So:

- On **this** branch: `ilog10` stays correct for every valid (≤ order-18) index; the >19-digit divergence is only on input that can't be constructed today. No guard, no change.
- In **#35**: order must be read from the explicit suffix (it already is, by design) — *not* re-derived from decimal digits. The one thing to make sure of in #35's review is simply that the rewrite of `mort2nested` reads the suffix and does **not** carry forward any `decimal_digit_count`-style inference. That's a #35-scope correctness check, and it's already what the plan specifies; adding a guard here would be speculative work on a path #35 removes.

If you'd like, I'll add an explicit acceptance-criterion line to the #35 plan: *"order is decoded from the 6-bit suffix only; no decimal-digit-count inference survives the rewrite."* That's where any hardening belongs — not on this PR.

## Q2 — `SmallVec<[usize; 8]>` for `Node.relevant` (item #3)

**What `Node.relevant` is.** During the polygon coverage descent, each cell is a `Node` carrying the **subset of polygon edge indices whose bounding caps reach that cell** ([`src_rust/src/coverage.rs#L263-L270`](https://github.com/espg/mortie/blob/claude/friendly-volta-mbsm1c/src_rust/src/coverage.rs#L263-L270)):

```rust
struct Node {
    pixel: u64,
    depth: u8,
    center: Vec3,
    fill: bool,
    relevant: Vec<usize>,   // ← edge indices still relevant to this cell
}
```

It's the working set the boundary classification iterates over: `node_straddles` loops `node.relevant` for vertex-in-cell, cell-edge-crossing, and parity tests ([`#L346-L386`](https://github.com/espg/mortie/blob/claude/friendly-volta-mbsm1c/src_rust/src/coverage.rs#L346-L386)), and `arc_crossing_parity` loops it too ([`#L252-L259`](https://github.com/espg/mortie/blob/claude/friendly-volta-mbsm1c/src_rust/src/coverage.rs#L252-L259)).

**Where the cost is.** Each straddle cell refines into 4 children, and **every child allocates a fresh heap `Vec<usize>`** by filtering the parent's relevant set ([`node_children`, `#L389-L406`](https://github.com/espg/mortie/blob/claude/friendly-volta-mbsm1c/src_rust/src/coverage.rs#L389-L406)):

```rust
let relevant: Vec<usize> = node
    .relevant
    .iter()
    .copied()
    .filter(|&i| edge_relevant(&edges[i], &center, cos_cr, sin_cr))
    .collect();   // ← one heap allocation per child
```

The descent goes down to the target order along the polygon boundary, so this `collect()` runs for **4 children at every straddle node, at every level, down the whole boundary tree** — i.e. the hot inner loop produces a large number of tiny, short-lived heap allocations (`malloc`/`free` per child, each holding only a handful of `usize`). As the descent refines, edges get culled fast, so the relevant set is almost always small — typically well under 8 entries near a leaf.

**What `SmallVec<[usize; 8]>` does.** `SmallVec<[T; N]>` stores up to `N` elements **inline (on the stack / inside the `Node`)** and only spills to the heap if the set exceeds `N`. Swapping `relevant: Vec<usize>` → `relevant: SmallVec<[usize; 8]>` means: for the common case (≤ 8 relevant edges), the per-child `collect()` writes into inline storage with **zero heap allocation**, and the per-element iteration in `node_straddles` / `arc_crossing_parity` becomes a contiguous inline scan (better cache locality, no pointer-chase to a heap buffer). Cells with > 8 relevant edges (coarse cells, dense boundaries) transparently spill to the heap exactly as `Vec` does today — so it's strictly a fast-path win with no behavioral change. It's purely allocation/locality; **output is bit-identical**.

**Expected impact.** It targets the same descent hot loop that phase 2 (#1/#2) already sped up — eliminating the dominant remaining per-child allocation. The gain is in fast/repeated coverage calls (many small polygons, deep boundary descent); it won't move a single large-polygon wall-clock much, but it cuts allocator pressure in the tight loop, which is what the audit flagged.

**The crate.** `smallvec` is tiny, ubiquitous (it's a Servo/rustc-ecosystem staple), **MIT/Apache-2.0, no transitive dependencies**. Low risk by any measure — but per **§4** it is still a new dependency, so I won't add it without your explicit OK.

**Options (your call — this is the only remaining cluster item):**
1. **Add `smallvec`** and land #3 as `SmallVec<[usize; 8]>` — cleanest, ~one-line type change in `Node` + the `collect()` target.
2. **No dependency:** hand-roll a fixed-capacity inline type (a small `enum { Inline([usize; 8], len) , Heap(Vec) }` or `[usize; N]` + overflow `Vec`). Same inline-storage benefit, no dep, but more code to write/maintain and review — and it re-implements what `smallvec` already does well.
3. **Skip #3** — phases 1/2/4/6 already deliver the bulk of the cluster's win (CodSpeed shows +23–45% on the coverage benches); #3 is incremental.

Since this is the dependency decision and it's the last open cluster item, I'm leaving it with you (`waiting`). Tell me which of the three and I'll land it next run.

---

espg:
> If you'd like, I'll add an explicit acceptance-criterion line to the https://github.com/espg/mortie/issues/35 plan: "order is decoded from the 6-bit suffix only; no decimal-digit-count inference survives the rewrite." That's where any hardening belongs — not on this PR.

make a note (i.e., 'proposal: order is decoded from the 6-bit suffix only; no decimal-digit-count inference survives the rewrite') in #35 ; the suffix implementation itself is under debate, so something we should think about as an option rather than a directive at this point.

Let's go with option 1 , i.e., add smallvec as a crate dependency. Seems like it's worth the addition, and is the last perf opt to implement, which will let us merge this after next pass.

espg:
🤖 *from Claude*

**Item #3 (the last cluster item) landed** in `c23bf56` — `phase 6 of issue #34: SmallVec for Node.relevant`.

Per your [go-ahead on option 1](https://github.com/espg/mortie/pull/41#issuecomment-4713466555) I added the `smallvec` crate (`smallvec = "1"`, MIT/Apache-2.0, no transitive deps — §4 sign-off recorded) and changed `Node.relevant` from `Vec<usize>` to `SmallVec<[usize; 8]>`:

- Field type + both construction sites (`base_node`, the per-child `collect()` in `node_children`) updated; all read sites (`.iter()`, `.is_empty()`, `&node.relevant` → `arc_crossing_parity(&[usize])`) work unchanged via `SmallVec`'s deref-to-slice.
- Purely an allocation/locality win: inline storage for the common ≤8-relevant-edges case, transparent heap spill beyond. **Bit-identical** — no Python API surface change.

**Verification (cargo gates — the relevant ones for a Rust-internal change):**
- `cargo test` → **86 passed, 0 failed** (incl. the polar-polygon coverage regression).
- `cargo clippy --all-targets` → warning count **identical before/after (28)**; zero new warnings, none reference the SmallVec change.
- `cargo fmt --check` → my edited regions are clean: diffing the per-file fmt output before vs. after my change shows **only line-number shifts**, no new hunks. (The crate isn't globally fmt-clean against this toolchain's `rustfmt`, and `clippy` reports the same pre-existing `useless_conversion`/`needless_range_loop` warnings on the untouched baseline — pre-existing drift already noted under "Questions for review" #2, left untouched.)
- `maturin`/`pytest` were **not available in this environment**, so I could not rebuild the extension or run the Python suite. This change is Rust-internal with no binding/API change, so the cargo gates above cover it; CI will run the full `pytest` on the PR.

**The "before-#35" perf cluster is now complete** — all six phases (#5/#1/#2/#10/#6/#3) are landed. Marking the PR `waiting` for your merge review. Not marking ready-for-review or merging per workflow — that's your call.

---

espg:
looks like everything is faster in codspeed -- merging

PR #43, opened 2026-06-15
decimal_morton: full-resolution 64-bit Morton MOC kernel (phase 1)

espg:
Refs #35

## What this lands

The `decimal_morton` datatype: the **packed-i64 encode/decode kernel** in Rust (the shared foundation every later skin sits on), the monotonic-suffix + point/area work, a **criterion benchmark** vs the existing order-≤18 path, and the **healpix-crate bridge** (`from_nested`/`to_nested`). Encoding-coupled #34 items (#4/#7/#8/#9) have landed; the Python surface for the `decimal_morton` skin itself is deferred to the skin phases (see "Follow-up").

Module `src_rust/src/decimal_morton.rs` (wired into the crate via a one-line `pub mod` in `lib.rs`); bench `src_rust/benches/decimal_morton_bench.rs` (wired in `Cargo.toml`).

## Bit-layout approach (as settled on #35)

64-bit word, MSB → LSB:

```
[ 4-bit prefix ][ 54-bit body (27 x 2-bit) ][ 6-bit suffix ]
  63 .. 60        59 ..  6                     5 ..  0
```

- **prefix** — base cell stored as `base_id + 1` → `1..=12`; `0` empty/null sentinel; `13..=15` invalid. Monotonic, so raw unsigned sort stays Z-order.
- **body** — 27 two-bit tuples, order 1 in the highest pair, order 27 in the lowest. Stored `0..=3`, interpreted `1..=4`.
- **suffix** — one monotone preorder code `0..=63`: `0..=27` variable-length area (order == suffix); `28..=47` order 28/29 area (preorder, div/mod 5); `48..=63` order-29 max-encoded **point** (`Kind::Point`, no area claim). Raw unsigned sort == Z-order across the full `0..=29` seam.

Storage is **unsigned**; signed "negative = southern" is a later-skin presentation detail. Encode **zero-fills** below the element's order (canonical — bit-equal dups, working equality/hash/dedup/sort).

### API
- `encode` / `encode_point` / `decode` — canonical pack / max-encoded point / full unpack.
- `order_of` / `base_cell_of` / `kind_of` — fast-path reads.
- `coarsen(word, k)` — body-mask + suffix rewrite for order `k`.
- **`from_nested(nested, depth)` / `to_nested(word)` — the healpix-crate bridge** (NESTED ↔ packed word; the cross-library interop foundation for UNIQ/`center`/`vertices`).

## Phases checklist (this PR — complete)

[...]

PR #49, opened 2026-06-18
remove MORTIE_FORCE_PYTHON parity fallbacks

### LINKED ISSUE #37: Deprecate pure-Python parity fallbacks (MORTIE_FORCE_PYTHON) for Rust functions
🤖 *from Claude*

## Deprecate the pure-Python parity fallbacks for the Rust-accelerated functions

Four functions still carry `MORTIE_FORCE_PYTHON=1`-gated pure-Python implementations whose only remaining job is to be diffed against the Rust path:

- `fastNorm2Mort`, `VaexNorm2Mort` — `mortie/tools.py` (asserted **bit-identical** in `mortie/tests/test_rust_vs_python.py`)
- `geo2mort` — `mortie/tools.py` (asserted **within <0.01% tolerance** in the same test; the Rust `healpix` crate and the Python path differ on boundary cases)
- `split_children` — `mortie/prefix_trie.py` (parity covered in `mortie/tests/test_prefix_trie.py`)

The Rust extension is the default and the only supported runtime path. Carrying the `_python_*` references plus the `MORTIE_FORCE_PYTHON` plumbing adds maintenance surface (every change to one of these functions must be made twice and kept in lockstep) for a fallback no shipped code uses.

### Proposal (for discussion)
Deprecate and eventually remove the Python fallbacks + the `MORTIE_FORCE_PYTHON` env switch. Open questions to settle before doing it:

1. Is the Python path ever needed as an escape hatch for an environment where the Rust wheel won't build/import? If so, keep but document it as unsupported rather than remove.
2. If removed, what replaces the parity tests — golden/expected-output fixtures captured from the current Rust output?
3. Staging: one release with a `DeprecationWarning` when `MORTIE_FORCE_PYTHON=1` is set, then removal in the following minor.

Flagging as `small-fix` to scope the doc/deprecation-warning step; the actual removal should be its own phased PR.

### Note
`CLAUDE.md` §7 was just corrected to describe the (non-uniform) parity contract accurately — this issue tracks retiring it rather than documenting it.

espg:
Closes #37

## What and why

Removes the `MORTIE_FORCE_PYTHON=1`-gated pure-Python parity twins whose only job was Rust-vs-Python diffing, per @espg's direction to "remove directly, no deprecation warning":

- `mortie/tools.py`: deletes `_python_fastNorm2Mort{,_scalar}`, `_python_VaexNorm2Mort{,_scalar}`, and `geo2mort`'s inline Python fallback block. `fastNorm2Mort`, `VaexNorm2Mort`, and `geo2mort` now call the Rust path unconditionally. The now-dead `RUST_AVAILABLE` / `RUST_GEO2MORT` flags and their `try/except` import guards are dropped (the Rust symbols are imported directly; a missing extension now fails loudly at import, which is the intended hard dependency).
- `mortie/prefix_trie.py`: deletes `_split_children_python` and the `RUST_AVAILABLE` / `FORCE_PYTHON` gate; `split_children` calls Rust unconditionally. The `os` import and module-level `FORCE_PYTHON` switch are removed from this module.

The public `geo2uniq` / `unique2parent` / `heal_norm` helpers (exported, independently tested) are untouched; they were used by `geo2mort`'s old fallback but are part of the public API.

### Note on #43 / the encoding layout

#43 landed only the Rust-internal `decimal_morton` kernel, so the **public** encoding is still the order-≤18 decimal path. These fallbacks were therefore not yet "wrong" (they still matched the live Rust output) — but their maintenance burden (every change made twice, kept bit-identical) is retired now. Parity coverage is replaced by golden fixtures captured from the **current** Rust output.

## How the parity tests were replaced

Rust-vs-Python parity is replaced by golden/expected-output assertions captured from the live Rust path on this branch (HEAD = `origin/main` a9f5b37):

- `mortie/tests/test_rust_vs_python.py` — `TestFastNorm2Mort` / `TestVaexNorm2Mort` / `TestGeo2Mort` now pin exact Rust output (the same input arrays the parity tests used). The 1000-element `fastNorm2Mort` array and the 1.24M-point Antarctic `geo2mort` order-18 output are pinned with a `sha256` content hash plus end spot-checks; the small cases pin the full expected arrays inline.
- `mortie/tests/test_prefix_trie.py` — `TestRustParity` → `TestRustGolden`: the recursively-collected characteristics, root-level `(characteristic, len)` pairs, and total covered count are pinned per case; the 500-element random case (`seed=42`) pins counts plus `sha256` of the characteristics and root-lens.

`TestMort2Norm` (and its `_python_mort2norm` reference) is left intact — see Questions for review.

## Phases

- [x] Phase 1: remove the four functions' `MORTIE_FORCE_PYTHON` fallbacks + dead flags; replace parity tests with golden fixtures captured from current Rust output.

## How it was tested

Built and run in a venv in the worktree (`maturin develop --release`, `pip install -e ".[test]"`). No Rust source changed, so `cargo` steps are N/A.

```
flake8 mortie --select=E9,F63,F7,F82   ->  clean (exit 0)
pytest -q                              ->  245 passed, 8 skipped, 2 warnings
```

The 8 skips are all pre-existing/environmental (shapely not installed, parquet fixture absent, coarse-cell buffer cases, the issue #20 hole-detection placeholder). The slow Antarctic golden test (`test_antarctic_polygon_stable`) passes when the dataset is present (`pytest -m slow` -> 1 passed).

## Questions for review

1. **`mort2norm` / `_python_mort2norm` left intact (scope).** The issue and `CLAUDE.md` §7 enumerate exactly four `MORTIE_FORCE_PYTHON` functions; `mort2norm` is not among them, yet it shares the `tools.py` `FORCE_PYTHON` global and has its own Python reference + `TestMort2Norm` parity class. I left it fully intact (the `FORCE_PYTHON` switch and `os` import remain in `tools.py` solely for it) to stay within the stated scope. If you'd like `mort2norm`'s fallback retired too in this PR, say so and I'll fold it in (replacing its parity test with a golden fixture the same way).
2. **`MortonChild.__init__` / `_compact` retained.** These build the trie from the string-grid char array — the path the removed `_split_children_python` used. They are *not* dead: `test_prefix_trie.py` constructs `MortonChild` directly and `_compact` recurses into them, so they stay.
3. **Benchmark scripts.** `benchmarks/benchmark_rust_vs_python.py` and `benchmark_morton_bbox.py` still set `MORTIE_FORCE_PYTHON`; for the three removed functions that env var is now a no-op (they always run Rust). They're not tests and not run by CI, so I left them out of scope — flag if you want them updated/removed.

https://claude.ai/code/session_011s41CmMVwHT9TsKyjrRS8W

---
_Generated by [Claude Code](https://claude.ai/code/session_011s41CmMVwHT9TsKyjrRS8W)_


mortie/tests/test_rust_vs_python.py
```diff
@@ -21,95 +30,114 @@ def _clean_env():
...
+    @pytest.mark.parametrize("order,expected", [
+        (6, 3112321),
+        (10, 31111112321),
+        (14, 311111111112321),
+        (18, 3111111111111112321),
+    ])
+    def test_different_orders(self, order, expected):
+        from mortie import tools
+        result = np.asarray(
+            tools.fastNorm2Mort(order, np.array([100]), np.array([2])),
```
espg: 🤖 *from Claude (review)*

Minor robustness nit: the old `test_different_orders` built its inputs as `np.array([100], dtype=np.int64)` / `np.array([2], dtype=np.int64)`, but the rewrite drops the explicit dtype (`np.array([100])`, `np.array([2])`). On the Linux CI matrix the platform default int is int64 so this passes, but `np.array([100])` is int32 on Windows, which could feed an unexpected dtype into the Rust FFI. The other rewritten cases in this file keep `dtype=np.int64`, so this one is inconsistent — suggest restoring the explicit dtype here for parity with the rest. Not blocking.

---

mortie/tests/test_rust_vs_python.py
```diff
@@ -122,6 +150,24 @@ def _valid_mortons(order, n=64):
...
+    return getattr(tools, func_name)(*args, **kwargs)
+
+
+def _get_python_result(func_name, *args, **kwargs):
+    """Call a mortie.tools function with pure Python."""
+    os.environ['MORTIE_FORCE_PYTHON'] = '1'
+    from mortie import tools
+    import importlib
+    importlib.reload(tools)
+    return getattr(tools, func_name)(*args, **kwargs)
```
espg: 🤖 *from Claude (review)*

Confirmation (not a defect): `_get_rust_result` / `_get_python_result` were moved down here rather than deleted, and that's correct — they're still the only mechanism `TestMort2Norm` below uses to exercise the retained `_python_mort2norm` reference via `MORTIE_FORCE_PYTHON`. Grepped the tree: no other caller, and no dangling reference to any removed `_python_*` / `_split_children_python` / `RUST_AVAILABLE` / `RUST_GEO2MORT` / `_assert_parity` / `_get_python_roots` symbol survives. So these two helpers are correctly scoped to mort2norm and not dead.

---

mortie/tests/test_prefix_trie.py
```diff
@@ -373,75 +375,76 @@ def test_child_mantissa_subset_of_parent(self):
...
-
-    def test_parity_positive(self):
+    def test_golden_positive(self):
         arr = np.array([1231, 1232, 1233], dtype=np.int64)
-        self._assert_parity(arr, max_depth=None)
+        self._assert_golden(
+            arr, None,
+            ['123', '1231', '1232', '1233'],
+            [('123', 3)], 3,
+        )
```
espg, line 411: 🤖 *from Claude (review)*

The golden conversion looks sound and is a genuine (non-tautological) guard: the expected `chars` / `rootlens` / `total` are hardcoded literals, not recomputed from the same `split_children` call, so a future Rust change to the trie shape would actually fail here. Verified by building (`maturin develop --release`) and running `pytest mortie/tests/test_prefix_trie.py mortie/tests/test_rust_vs_python.py -q` -> 71 passed, 4 skipped on this branch; all pinned values and both sha256 hashes match the live Rust output. Input coverage is preserved one-to-one from the old parity cases (positive / negative / mixed-sign / max_depth 0 / max_depth 2 / identical / single / 500-elem seed=42), so nothing the old `_assert_parity` exercised was dropped.

---

mortie/tools.py
```diff
@@ -5,24 +5,15 @@
 import numpy as np
 import os
 from . import _healpix as hp
+from . import _rustie
 
-# Allow forcing pure Python for testing/comparison
+# Allow forcing pure Python for the mort2norm parity reference
 FORCE_PYTHON = os.environ.get('MORTIE_FORCE_PYTHON', '0') == '1'
```

Confirmation: retaining `os` + the module-level `FORCE_PYTHON` here is correct and not dead code — `mort2norm` (line 242, `if FORCE_PYTHON or not RUST_MORT2NESTED:`) still consumes it, and `mort2norm` was intentionally left out of scope (issue #37 / CLAUDE.md §7 enumerate only the four removed functions). The three public encoders (`fastNorm2Mort`, `VaexNorm2Mort`, `geo2mort`) now call the Rust path unconditionally with no behavioral change beyond dropping the fallback branch — the Rust call args and the `geo2mort` contiguous-array / `np.atleast_1d` post-processing are byte-for-byte the same as the old Rust branch. The `try/except` import guards becoming direct imports (failing loudly on a missing extension) matches the stated hard-dependency intent. flake8 E9/F63/F7/F82 is clean (exit 0); the only F841 hits in these files (prefix_trie.py:103/106, tools.py:517) are pre-existing and outside this diff's hunks.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of phase 1 (remove `MORTIE_FORCE_PYTHON` fallbacks for the four enumerated functions + replace parity tests with golden fixtures). Built and ran the suite on this branch to verify; the diff is clean and does what the PR body claims. Summary:

- **No dead code / dangling refs.** Every removed symbol (`_python_fastNorm2Mort*`, `_python_VaexNorm2Mort*`, `_split_children_python`, `RUST_AVAILABLE`, `RUST_GEO2MORT`, `_assert_parity`, `_get_python_roots`) has zero surviving references in the tree.
- **No behavior change in the public functions** beyond dropping the fallback branch — the Rust call args and `geo2mort` post-processing are unchanged.
- **Golden fixtures are real regression guards** (hardcoded literals + sha256, not recomputed from the asserted call) and preserve the old parity cases' input coverage one-to-one. 71 passed / 4 skipped locally.
- **Scope is contained** to the four functions + their tests; `mort2norm` and its `os`/`FORCE_PYTHON` plumbing are correctly left intact.
- **No flake8 E9/F-class issues introduced** (the few F841 in these files are pre-existing and outside the diff).

One non-blocking nit inline (dropped explicit `dtype=np.int64` in `test_different_orders`); rest are confirmations.

---

mortie/prefix_trie.py
```diff
@@ -14,10 +14,11 @@
 """
 
 import math
-import os
 import numpy as np
 
-FORCE_PYTHON = os.environ.get('MORTIE_FORCE_PYTHON', '0') == '1'
+from . import _rustie
+
+_rust_split_children = _rustie.split_children_rust
```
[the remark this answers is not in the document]
espg: 🤖 *from Claude*

Fixed in cb35c48 — import block reordered to stdlib / third-party / first-party with `_rustie` moved to the top with the other imports.

---

mortie/tests/test_rust_vs_python.py
```diff
@@ -1,17 +1,26 @@
...
+parity check remains below.
 """
 
+import hashlib
 import numpy as np
 import os
 import pytest
 from pathlib import Path
 
+# Captured from the Rust geo2mort output at order 18 on the Antarctic dataset.
```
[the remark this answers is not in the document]
espg: 🤖 *from Claude*

Fixed in cb35c48 — top-level imports reordered (stdlib `hashlib`/`os`/`pathlib`, then `numpy`/`pytest`).

---

mortie/tests/test_rust_vs_python.py
```diff
@@ -122,6 +150,24 @@ def _valid_mortons(order, n=64):
     return np.asarray(tools.fastNorm2Mort(orders, normed, parents), dtype=np.int64)
 
 
+def _get_rust_result(func_name, *args, **kwargs):
+    """Call a mortie.tools function with Rust enabled."""
+    os.environ.pop('MORTIE_FORCE_PYTHON', None)
+    from mortie import tools
+    import importlib
+    importlib.reload(tools)
```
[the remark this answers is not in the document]
espg: 🤖 *from Claude*

Fixed in cb35c48 — `import importlib` hoisted above the `from mortie import tools` line in `_get_rust_result`.

---

mortie/tests/test_rust_vs_python.py
```diff
@@ -122,6 +150,24 @@ def _valid_mortons(order, n=64):
...
+    return getattr(tools, func_name)(*args, **kwargs)
+
+
+def _get_python_result(func_name, *args, **kwargs):
+    """Call a mortie.tools function with pure Python."""
+    os.environ['MORTIE_FORCE_PYTHON'] = '1'
+    from mortie import tools
+    import importlib
+    importlib.reload(tools)
```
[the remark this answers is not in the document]
espg: 🤖 *from Claude*

Fixed in cb35c48 — same reorder applied in `_get_python_result`.

---

mortie/tools.py
```diff
@@ -5,24 +5,15 @@
 import numpy as np
 import os
 from . import _healpix as hp
+from . import _rustie
 
-# Allow forcing pure Python for testing/comparison
+# Allow forcing pure Python for the mort2norm parity reference
```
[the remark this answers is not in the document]
espg: 🤖 *from Claude*

Fixed in cb35c48 — `tools.py` imports reordered to stdlib / third-party / first-party.

---

mortie/tests/test_prefix_trie.py
espg, line 1: 🤖 *from Claude*

Fixed in cb35c48 — `numpy`/`pytest` reordered, the `from mortie.prefix_trie import (...)` names sorted, and the extra blank line removed. `ruff check --select I001` is now clean on all four changed files.

---

mortie/tests/test_rust_vs_python.py
espg: 🤖 *from Claude*

Addressed in b416ef7 — restored explicit `dtype=np.int64` on both input arrays in `test_different_orders`, matching the other rewritten cases and avoiding the int32-on-Windows path into the Rust FFI.

---

espg:
> Benchmark scripts. benchmarks/benchmark_rust_vs_python.py and benchmark_morton_bbox.py still set MORTIE_FORCE_PYTHON; for the three removed functions that env var is now a no-op (they always run Rust). They're not tests and not run by CI, so I left them out of scope — flag if you want them updated/removed.

updated as appropriate -- if they just touch the depreciated python path (i.e., that's what they test), they should be removed.

> MortonChild.__init__ / _compact retained. These build the trie from the string-grid char array — the path the removed _split_children_python used. They are not dead: test_prefix_trie.py constructs MortonChild directly and _compact recurses into them, so they stay.

Yup, makes sense, keep it.

> mort2norm / _python_mort2norm left intact (scope). The issue and CLAUDE.md §7 enumerate exactly four MORTIE_FORCE_PYTHON functions; mort2norm is not among them, yet it shares the tools.py FORCE_PYTHON global and has its own Python reference + TestMort2Norm parity class. I left it fully intact (the FORCE_PYTHON switch and os import remain in tools.py solely for it) to stay within the stated scope. If you'd like mort2norm's fallback retired too in this PR, say so and I'll fold it in (replacing its parity test with a golden fixture the same way).

the _python_mort2norm doesn't support the full order scope (i.e., to order 29), and is also matched to the prior negative basenumber convention in the southern hemisphere. I also can't imagine it being more performant than rust. We should cut out _python_mort2norm too, and make sure that the mort2norm excises the 'force python' code and logic... and also verify that the rust code matches the new order capacity that we've built in to #35 work.

espg:
🤖 *from Claude*

Phase 2 (`27db50e`) — folds in the directives from your [review comment](https://github.com/espg/mortie/pull/49#issuecomment-4737450912).

**1. Cut `_python_mort2norm` + the force-python switch (`mortie/tools.py`)**
- Deleted `_python_mort2norm` and the `if FORCE_PYTHON or not RUST_MORT2NESTED:` branch in `mort2norm`; it now calls `_rust_mort2nested` unconditionally.
- Removed the module-level `FORCE_PYTHON` global and the now-unused `import os` (grepped: their only consumers were the removed branch). The `try/except` import guard / `RUST_MORT2NESTED` flag are gone — `_rust_mort2nested` is a direct import that fails loudly on a missing extension, matching phase 1's pattern. Dropped the imported-but-never-used `_rust_nested2mort` alias with it.
- `mortie/tests/test_rust_vs_python.py`: replaced `TestMort2Norm`'s Rust-vs-Python parity (and the `_get_rust_result`/`_get_python_result`/`_valid_mortons`/`_clean_env`/`os` plumbing it needed) with golden fixtures captured from the **current** Rust output — hardcoded `(normed, parent, order)` literals, not recomputed from the asserted call. Cases: order-1 all-12-parents (both hemispheres), order-6 spread, order-18 spread, scalar north/south, plus a `fastNorm2Mort`-roundtrip guard at order 14.

**2. Order-29 verification — FINDING: the public `mort2norm` Rust path does NOT support order 29; it is capped at order ≤18.**
- `src_rust/src/morton.rs`: `POWERS_OF_10`/`POWERS_OF_4` are length-19 arrays (orders 0–18); `fast_norm2mort_scalar` `panic!`s if `order > 18`; `mort2nested` indexes `POWERS_OF_10[order_usize]` (out-of-bounds → panic for order > 18, surfaced as a `ValueError` via the `catch_unwind` in `rust_mort2nested`).
- Empirically confirmed: a 30-digit order-29 morton (`int("1"+"1"*29)`) raises `OverflowError: Python int too large to convert to C long` in `mort2norm` (it can't even fit i64).
- This matches the PR body's note that #43 landed only the Rust-internal `decimal_morton` kernel — the public decimal encoding is still order ≤18. Per your steer, I did **not** expand scope to implement order-29 (that's the #48 v1.0 work) and kept the golden fixtures at the currently-supported orders, with order 18 as the explicit max case (a comment in `TestMort2Norm` records this).

**3. Benchmark scripts — both removed.**
- `benchmarks/benchmark_rust_vs_python.py`: its sole purpose is timing Rust vs the `MORTIE_FORCE_PYTHON=1` Python path of `fastNorm2Mort`/`geo2mort`. With those fallbacks gone the env var is a no-op and "Python" == "Rust", so the comparison is meaningless. Removed.
- `benchmark_morton_bbox.py` (repo root): benchmarks `split_children` Rust vs the removed `_split_children_python`, and additionally imports `mortie.morton_bbox` — a module that no longer exists (it's `mortie.prefix_trie` now), so it was already broken. Removed.
- Both fall squarely under "if they just touch the deprecated python path … they should be removed." (Verified no other tracked file references either script or `MORTIE_FORCE_PYTHON` except `CLAUDE.md` — see Questions.)

**4. Dropped-dtype nit** ([discussion_r3432378582](https://github.com/espg/mortie/pull/49#discussion_r3432378582)) — already restored in `b416ef7`; `test_different_orders` carries explicit `dtype=np.int64` on both input arrays. No change needed this phase; noting for completeness.

**Tested** (no Rust source changed → `cargo` N/A):
```
flake8 mortie --select=E9,F63,F7,F82   -> clean (exit 0)
ruff check --select I001 (both files)  -> All checks passed
pytest -q                              -> 245 passed, 8 skipped, 2 warnings
```
Same 245/8 totals as the pre-phase baseline; the 8 skips are pre-existing/environmental. `TestMort2Norm` went from 6 parity tests to 6 golden tests.

## Questions for review
1. **`CLAUDE.md` §7 is now stale but I left it untouched.** It still lists `fastNorm2Mort`/`VaexNorm2Mort`/`geo2mort`/`split_children` as `MORTIE_FORCE_PYTHON`-gated functions with Python references — all removed in phase 1, and `mort2norm`'s reference is now gone too. The auto-mode classifier reads that file, and editing conventions/scope wording reads as side-effecting, so I did not touch it. Want me to update §7 (drop the fallback list / note "no Python fallbacks remain") in this PR, or handle it separately?
2. **Order-29 is out of scope here** per the finding above — confirming you're happy with the golden fixtures capped at order 18 and order-29 left to #48.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of phase 2 (commit `27db50e`, "phase 2 of issue #37") only. I verified the claims rather than trusting them; the phase is in good shape and I found no blocking issues. Summary of what I checked, then per-finding notes (each led with file:line).

**Verification performed (worktree, editable install):**
- `flake8 mortie --select=E9,F63,F7,F82` → clean (exit 0).
- `ruff check --select I001 mortie/tools.py mortie/tests/test_rust_vs_python.py` → all checks passed.
- `python -m pytest mortie/tests/test_rust_vs_python.py -q` → 16 passed.
- Style pass `--max-line-length=88` on the two changed py files reports E501/F841 only on **pre-existing** lines (166, 182, 444, 461, 472, 630) — none introduced by this diff; correctly left alone per §4 ("don't fix unrelated pre-existing failures").
- Grep across the tree for `_python_mort2norm`, `FORCE_PYTHON`, `MORTIE_FORCE_PYTHON`, `RUST_MORT2NESTED`, `_rust_nested2mort`, `benchmark_rust_vs_python`, `benchmark_morton_bbox` → no dangling code references; only hit is CLAUDE.md prose (see finding 3).
- No dependency added. Order-29 correctly NOT implemented (out of scope, #48).

**Findings**

1. `mortie/tests/test_rust_vs_python.py:144-193` — CONFIRMATION (not actionable). The golden literals are genuinely pinned, not tautological: they are hardcoded `assert_array_equal` targets, not recomputed from the asserted call. I independently re-encoded every fixture with `fastNorm2Mort` and confirmed each roundtrips back to the input mortons (order1/6/18 all `True`), and that the order6 array values for parent 2 / parent 7 are exactly the `test_scalar_north` (2120,2,6) and `test_scalar_south` (1533,7,6) literals — so the array and scalar fixtures are internally consistent. Good regression guards.

2. `mortie/tools.py:214-216` vs `:231-236` — CONFIRMATION. The empty-input edge case the old parity test never exercised is still handled correctly: `mort2norm` returns early on `morton.size == 0` before reaching `_rust_mort2nested`, so dropping the Python fallback is behavior-preserving. Mixed-order (`:226-227`) and scalar (`:238-239`) paths are likewise untouched and still covered by the new scalar tests. No edge case was lost relative to the removed `_valid_mortons`-based parity test (that test only fed valid same-order mortons).

3. `CLAUDE.md` §7 — NON-BLOCKING / for @espg (out of scope for this diff, flagging not directing). After this phase, `FORCE_PYTHON` is entirely gone from `tools.py` and `fastNorm2Mort`/`VaexNorm2Mort`/`geo2mort`/`mort2norm` all call Rust unconditionally — yet CLAUDE.md §7 still describes `fastNorm2Mort`/`VaexNorm2Mort`/`geo2mort` (and `split_children`) as "`MORTIE_FORCE_PYTHON=1`-gated" with bit-identical/tolerance parity contracts. That governance text is now stale for the tools.py functions. Updating CLAUDE.md may be intentional final-phase work or may need an explicit decision (it's a hard-requirement file); raising it rather than acting. Note: §7 never listed `mort2norm` as gated, so this phase's removal doesn't contradict §7 directly — it's the cumulative PR that does.

4. Module docstring `mortie/tests/test_rust_vs_python.py:2-10` and class docstring `:137-142` — CONFIRMATION. Docstrings were updated to drop the "mort2norm still carries a Python reference" claim and now accurately describe golden-only testing; the order-18 cap rationale (30-digit order-29 overflows i64, deferred to #48) is correctly documented and matches the test contents.

5. Commit message `phase 2 of issue #37` — CONFIRMATION. Title-only, matches repo log style (§3). Benchmark deletions are appropriate dead-code removal (both scripts only existed to compare the now-removed Python paths). No speculative abstraction introduced.

No diff-scoped changes are required from my side; finding 3 is the only item standing for @espg.

---

espg:
> Updating CLAUDE.md may be intentional final-phase work or may need an explicit decision (it's a hard-requirement file); raising it rather than acting.

Strike the offending text from CLAUDE.md , and I'll review it

espg (commented):
🤖 *from Claude (review)*

Adversarial review of `5beff14` ("phase 3 of issue #37") — the §7 CLAUDE.md strike. Verified the new text against the actual post-PR code state. Everything checks out; no actionable findings.

**1. "No remaining Python fallbacks" — accurate.** Grepping `mortie/tools.py` and `mortie/prefix_trie.py` finds zero occurrences of `MORTIE_FORCE_PYTHON`, `FORCE_PYTHON`, `RUST_AVAILABLE`, `RUST_GEO2MORT`, `RUST_MORT2NESTED`, or any `_python_*` twin. (The remaining hits in the tree are confined to a stale `.claude/worktrees/` snapshot and to `benchmarks/`, not the shipped package.)

**2. Affected-function list — complete and correct.** `fastNorm2Mort`, `VaexNorm2Mort`, `geo2mort`, `mort2norm` all live in `tools.py` and call Rust unconditionally (`_rust_fast_norm2mort` / `_rust_geo2mort` / `_rust_mort2nested`); `split_children` lives in `prefix_trie.py` and calls `_rust_split_children` unconditionally. The set named in the text matches the code exactly — `mort2norm` is correctly included (it was the addition over the original four).

**3. No contradiction left elsewhere in CLAUDE.md.** Grepping the whole file for `FORCE_PYTHON` / `fallback` / `parity` / `Python twin` / `pure-Python` returns only line 82 itself (the edited line). No other section still asserts a parity contract or `MORTIE_FORCE_PYTHON` gate.

**4. "Fails loudly at import" — true.** Both modules do a bare `from . import _rustie` at module top with no try/except; every other `_rustie` import across the package is likewise unguarded. The only try/except in `__init__.py` is for version metadata (`PackageNotFoundError`), unrelated to the extension. A missing extension therefore raises `ImportError` at import time, exactly as stated.

**5. Scope — contained.** The edit is the single directed §7 paragraph (1 line changed). It does not touch other sections. The added parentheticals ("the inverse/encode API", "golden/pinned fixtures captured from the Rust path") are accurate and minor; not an overreach.

Confirmation only — nothing to fold in.

---

espg:
🤖 *from Claude*

Phase 3 (`5beff14`) — struck the stale `MORTIE_FORCE_PYTHON` text from `CLAUDE.md` §7, per [your directive](https://github.com/espg/mortie/pull/49#issuecomment-4743847912).

The old §7 paragraph described `fastNorm2Mort`/`VaexNorm2Mort`/`geo2mort`/`split_children` (and implicitly `mort2norm`) as carrying `MORTIE_FORCE_PYTHON`-gated Python references with bit-identical/tolerance parity contracts — all removed in this PR. Replaced it with an accurate statement: the Rust extension is the sole runtime path, no Python fallbacks remain, those functions call Rust unconditionally and a missing extension fails loudly at import, and functionality is tested against golden/pinned fixtures. Single-paragraph change, scope contained to the directed strike.

Fresh-context self-review posted above (COMMENT) — confirmed accurate on all checks: no surviving `MORTIE_FORCE_PYTHON`/`FORCE_PYTHON`/`_python_*` refs for those five functions in the shipped package, the function list is complete and correct, no other part of `CLAUDE.md` is left contradictory, and "fails loudly at import" holds (both modules do an unguarded `from . import _rustie`). No diff-scoped findings to fold.

**This completes the PR's phases** (Phase 1: remove the four functions' fallbacks + golden fixtures; Phase 2: cut `_python_mort2norm` + the `FORCE_PYTHON` switch + dead benchmark scripts; Phase 3: `CLAUDE.md` §7 strike). The one open item — order-29 support for the public `mort2norm` path — was found out of scope here (the public decimal path is capped at order ≤18) and is now under discussion on #48 Tier 0, where I've laid out the packed-`uint64` kernel vs. 128-bit/Nx2/string options. Marking `waiting` — the PR is complete and awaiting your review/merge.

Local gate on this branch: `flake8 mortie --select=E9,F63,F7,F82` clean; `pytest -q` 245 passed / 8 skipped (skips pre-existing/environmental). No Rust source changed this phase, so `cargo` steps are N/A. Codecov shows all modified lines covered; the CodSpeed `coverage_triangle[4]` regression flagged on the PR is noise (this PR changes no Rust/coverage code — note the "different runtime environments" warning).

---

Issue #50, opened 2026-06-18
RangeMOC / `moc`-crate adoption decision (split from #35 Phase 3)


espg:
🤖 *from Claude*

Split out from #35 (Phase 3) and #48 per direction, so the `moc`/RangeMOC adoption call has its own thread to land on. Detailed options + impact below, grounded in an actual `cargo tree`.

## The blocking `cargo tree` check — resolved (and it's the bad case)
Ran a throwaway crate with `moc = "0.19"` alongside our `healpix = "0.3.2"`:

```
├── healpix v0.3.2          (mortie's current HEALPix dep)
└── moc v0.19.2
    └── cdshealpix v0.9.1    (moc's HEALPix dep)
```

**`healpix` and `cdshealpix` are two distinct published crates**, not the same crate under one name. So pulling `moc` adds a **second, independent HEALPix implementation** (`cdshealpix 0.9.1`) *on top of* our existing `healpix 0.3.2` — they do **not** unify. (The earlier research note guessed they might be the same crate; the tree disproves that.)

**Impact of the duplicate:**
- Two HEALPix crates compiled in → larger binary / **bigger Lambda layer** (the cost you flagged).
- Two implementations of nested/uniq/cone logic. The *values* still interoperate (both follow the HEALPix spec, indices are plain `u64`), so conversions are integer-level and functionally safe — but it's redundant compiled code and two deps to track on version bumps.
- `moc` is pre-1.0 (0.19.2, ~13 breaking releases) — pin exact, expect upgrade churn.

## Options
**A. Status quo — keep the in-house `moc.rs`.** We already have the single-pass `moc::normalize` (landed #43). No new dep, no duplicate HEALPix. Cost: no FITS/IVOA serialization, no range storage, no MOCPy/Aladin interop.

**B. `moc` for serialization/interop ONLY, behind a Cargo feature.** Pull `moc` (+`cdshealpix 0.9.1`) solely for `RangeMOC` ↔ **FITS/ASCII/JSON** (IVOA-standard interchange with astropy / MOCPy / Aladin — the one thing neither our code nor `healpix`'s BMOC provides). Keep the descent + in-house normalize. **Feature-flag it** so the default/Lambda build stays lean and only opt-in builds carry the 2nd HEALPix crate.

[...]

PR #53, opened 2026-06-18
MOC boolean set algebra via the patched healpix BMOC fork

espg:
Refs #50

## What this does

Implements fast MOC boolean set-algebra over mortie's signed decimal-morton covers — the **Option D** decision in #50 — backed by the `healpix` crate's BMOC (`Bmoc::or` / `and` / `minus`). No `moc` crate, no `cdshealpix`, no FITS/IVOA serialization.

**Dependency swap (approved in https://github.com/espg/mortie/issues/50#issuecomment-4743717025):** the root `Cargo.toml` `healpix` dep is moved from crates.io `0.3.2` to the @espg-patched fork, pinned by rev:

```toml
healpix = { git = "https://github.com/espg/healpix-rs", rev = "9118d3834b6c2a8860dc7a375d46df2d51516df9" }
```

The fork is `healpix` v0.3.3 with the BMOC `minus` fix present. The crate name is still `healpix`, so the existing `use healpix::...` imports across the tree (`geo2mort.rs`, `buffer.rs`, `linestring.rs`, `decimal_morton.rs`, benches) are unaffected — they compile unchanged against the fork.

### Public API added (`mortie/coverage.py`, exported from `__all__`)
- `moc_or(a, b)` — union; `moc_or(a, b) == compress_moc(concatenate([a, b]))`
- `moc_and(a, b)` — intersection
- `moc_minus(a, b)` — difference `a \ b`

Names and design inherited from #52 (@espg chose the terser BMOC verbs over union/intersection/difference in https://github.com/espg/mortie/pull/52#issuecomment-4737665127). Docstrings note the equivalences and carry `See Also` cross-refs. Rust internals and PyO3 bindings are named consistently: `moc::moc_or/moc_and/moc_minus` → `rust_moc_or/rust_moc_and/rust_moc_minus`, registered in the module init and wrapping the compute in `py.allow_threads`, mirroring the existing `rust_moc_normalize` binding.

### Approach
Signed-morton handling stays at the boundary, mirroring the existing decode→op→encode pattern: decode each morton i64 → `(depth, nested u64)` via `mort2nested`, build a packed BMOC per input, run the BMOC op, extract cells, re-encode via `nested2mort`, then `normalize`. The decimal sign carries the hemisphere through the nested hash, so southern covers need no special-casing.

One correctness subtlety worth flagging: the BMOC ops require a **valid, non-overlapping** input cover, and `pack()` only merges complete sibling quartets — it does not prune a descendant of a coarser cell in the same set. A raw mortie cover can contain overlaps (e.g. a coarse cell plus a finer descendant), so `build_bmoc` runs mortie's `normalize` first to deliver the canonical, ancestor-pruned, sibling-merged cover the BMOC ops assume. Building straight from the raw cover produced wrong results in testing (spurious cells); normalizing first makes every op match the brute-force reference.

## Supersedes #52

This replaces #52's **in-house sorted-range merge** with the upstream BMOC. #52 went in-house because upstream `healpix` 0.3.2's `Bmoc::minus` **infinite-loops** on mixed-order inputs: in the `Greater` arm (right cell coarser than left), when the finer left cell is fully covered by a coarser *full* right cell, the original code fell through without advancing either cursor. The fork adds the terminal `else` arm (`bmoc/ops.rs::minus`):

```rust
} else {
    // hl_at_dr == r.hash && r.is_full && l.is_full:
    // the finer left cell is fully covered by the coarser full
    // right cell, so `l - r` is empty. Drop `l` and advance left;
    // keep `r`, since it may still cover further fine left cells.
    // (Without this branch neither cursor advanced => infinite loop.)
    left = it_left.next();
}
```

#52 stays **blocked** (per #50); this PR does not touch #52 or its branch.

## How tested (real local counts)
- `cargo test --release` — **154 passed, 0 failed** (21 of them in `moc::`, including the new set-ops tests).
- New Rust tests in `moc.rs`: `or` bit-for-bit == `normalize(concat(a,b))`; `and`/`minus` vs a brute-force deepest-order leaf-set reference; disjoint covers; self-minus empty; mixed-order; both hemispheres; empty inputs; and a 300-iteration randomized fuzz over mixed-order cover pairs in both hemispheres comparing all three ops to the reference.
- **Regression test (`test_minus_mixed_order_terminates_regression`)** reproduces the exact 0.3.2 trigger — a cover whose many finer (depth-5) cells are descendants of the other's coarse full (depth-2) cells — and asserts `minus` now *terminates* and matches the brute-force reference. On 0.3.2 this hangs forever; reaching the assert proves the fork fix.
- `cargo clippy` / `cargo fmt --check` — clean on the changed files (`moc.rs`, the new `lib.rs` bindings). The pre-existing tree-wide `useless_conversion` lint and pre-existing `lib.rs` fmt drift are left alone (the new bindings match the surrounding `.to_vec()?` idiom exactly).
- `flake8 mortie --select=E9,F63,F7,F82` — clean.
- `pytest -v` — **290 passed, 9 skipped, 0 failed** (93% coverage). New `TestMOCSetOps` in `test_coverage.py` covers `moc_or`/`moc_and`/`moc_minus`: ==compress(concat) for or, brute-force via `moc_to_order` for and/minus, disjoint, self-minus empty, mixed-order, empty, southern hemisphere. (The 3 `test_arrow.py` failures seen on a first pass were a missing-`pandas` environment gap — a docs/notebook extra, not this change — and all pass once `pandas` is installed.)

## Phases
- [x] Phase 1 — dependency swap to the pinned fork rev in `Cargo.toml`; confirm the tree still builds against it.
- [x] Phase 2 — Rust `moc_or`/`moc_and`/`moc_minus` backed by BMOC + PyO3 bindings in `lib.rs`; Rust tests incl. the mixed-order `minus` regression and fuzz.
- [x] Phase 3 — Python wrappers in `coverage.py`, exports in `__init__.py`, `TestMOCSetOps` in `test_coverage.py`.

## Questions for review
1. **Pin by `rev` vs `tag`** — pinned to the fork's commit `9118d383` for reproducibility. Prefer a tagged release on the fork (e.g. a `0.3.3-mortie` tag) instead, so the dep reads more clearly and survives force-pushes to the fork?
2. **Also expose `not` / `xor`?** The fork's BMOC implements `not` (complement) and `xor` (symmetric difference) too. Out of scope for #50 (Option D names only `or`/`and`/`minus`); add them in this PR, or leave for a follow-up?
3. **Rust internal fn naming** — kept the internal Rust fns as `moc_or`/`moc_and`/`moc_minus` (matching the public Python names and avoiding #52's `union` vs `moc_or` mismatch). Confirm that's the preferred convention over bare `or`/`and`/`minus` (which would shadow the BMOC trait verbs).
4. **Normalize-on-build** — `build_bmoc` normalizes each input before constructing the BMOC (rationale above). That's an extra sort/pass per operand; acceptable, or should callers be required to pass already-normalized covers and skip it?

https://claude.ai/code/session_01Fggjz7kP1WyxkWQbUdQahH

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fggjz7kP1WyxkWQbUdQahH)_

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ path = "src_rust/src/lib.rs"
 pyo3 = { version = "0.22", features = ["extension-module", "abi3-py310"] }
 numpy = "0.22"
 rayon = "1.10"
-healpix = "0.3.2"
+healpix = { git = "https://github.com/espg/healpix-rs", rev = "9118d3834b6c2a8860dc7a375d46df2d51516df9" }
 smallvec = "1"
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ path = "src_rust/src/lib.rs"
 pyo3 = { version = "0.22", features = ["extension-module", "abi3-py310"] }
 numpy = "0.22"
 rayon = "1.10"
-healpix = { git = "https://github.com/espg/healpix-rs", rev = "9118d3834b6c2a8860dc7a375d46df2d51516df9" }
+healpix = "0.3.3"
 smallvec = "1"
 
 [dev-dependencies]
--- a/mortie/__init__.py
+++ b/mortie/__init__.py
@@ -42,6 +42,9 @@ from .coverage import (
     morton_coverage_moc,
     compress_moc,
     moc_to_order,
+    moc_or,
+    moc_and,
+    moc_minus,
 )
 from .linestring import linestring_coverage
 
@@ -81,6 +84,9 @@ __all__ = [
     'morton_coverage_moc',
     'compress_moc',
     'moc_to_order',
+    'moc_or',
+    'moc_and',
+    'moc_minus',
     'linestring_coverage',
     'MortonChild',
     'split_children',
--- a/mortie/coverage.py
+++ b/mortie/coverage.py
@@ -254,3 +254,84 @@ def moc_to_order(morton, order):
 
     morton = np.asarray(morton, dtype=np.int64).ravel()
     return np.asarray(_rustie.rust_moc_to_order(morton, order))
+
+
+def moc_or(a, b):
+    """Union of two morton covers (the cells in ``a`` or ``b``).
+
+    Equivalent to ``compress_moc(concatenate([a, b]))``, but computed by the
+    healpix-crate BMOC ``or`` rather than a concatenate-then-compress pass.
+
+    Parameters
+    ----------
+    a, b : array_like
+        Morton covers (mixed order allowed).
+
+    Returns
+    -------
+    numpy.ndarray
+        Sorted, compacted union (``int64``).
+
+    See Also
+    --------
+    moc_and : intersection of two covers.
+    moc_minus : difference ``a \\ b``.
+    compress_moc : ``moc_or(a, b) == compress_moc(concatenate([a, b]))``.
+    """
+    from . import _rustie
+
+    a = np.asarray(a, dtype=np.int64).ravel()
+    b = np.asarray(b, dtype=np.int64).ravel()
+    return np.asarray(_rustie.rust_moc_or(a, b))
+
+
+def moc_and(a, b):
+    """Intersection of two morton covers (the cells in both ``a`` and ``b``).
+
+    Parameters
+    ----------
+    a, b : array_like
+        Morton covers (mixed order allowed).
+
+    Returns
+    -------
+    numpy.ndarray
+        Sorted, compacted intersection (``int64``).
+
+    See Also
+    --------
+    moc_or : union of two covers.
+    moc_minus : difference ``a \\ b``.
+    """
+    from . import _rustie
+
+    a = np.asarray(a, dtype=np.int64).ravel()
+    b = np.asarray(b, dtype=np.int64).ravel()
+    return np.asarray(_rustie.rust_moc_and(a, b))
+
+
+def moc_minus(a, b):
+    """Difference of two morton covers (the cells in ``a`` but not ``b``).
+
+    Computes ``a \\ b``.
+
+    Parameters
+    ----------
+    a, b : array_like
+        Morton covers (mixed order allowed).
+
+    Returns
+    -------
+    numpy.ndarray
+        Sorted, compacted difference (``int64``).
+
+    See Also
+    --------
+    moc_or : union of two covers.
+    moc_and : intersection of two covers.
+    """
+    from . import _rustie
+
+    a = np.asarray(a, dtype=np.int64).ravel()
+    b = np.asarray(b, dtype=np.int64).ravel()
+    return np.asarray(_rustie.rust_moc_minus(a, b))
--- a/src_rust/src/moc.rs
+++ b/src_rust/src/moc.rs
@@ -58,6 +58,10 @@ fn build_bmoc(morton: &[i64]) -> MutableBmoc {
 }
 
 /// Re-encode a BMOC's cells back to a normalized mortie morton cover.
+///
+/// Discarding each cell's full/partial flag is safe here: every cell mortie
+/// pushes is `is_full = true`, so the BMOC only ever emits full cells — the flag
+/// carries no information a `(hash, depth)` cover would lose.
 fn bmoc_to_morton(bmoc: MutableBmoc) -> Vec<i64> {
     let cells: Vec<i64> = bmoc
         .into_cells()
@@ -526,25 +530,36 @@ mod tests {
     /// whose finer cells are descendants of the other's coarse full cells.
     #[test]
     fn test_minus_mixed_order_terminates_regression() {
-        // a: depth-2 full cells (coarse) spanning two base cells.
-        let a: Vec<i64> = (0..3)
-            .map(|n| nested2mort(n, 2))
-            .chain((0..3).map(|n| nested2mort((4u64 << 4) + n, 2)))
+        // The hang is in `minus`'s `Greater` arm, which fires only when a *finer*
+        // LEFT (minuend) cell is a descendant of a coarser *full* RIGHT
+        // (subtrahend) cell — both `is_full`. So the minuend must be the FINER
+        // cover and the subtrahend the coarse full ancestor. (An earlier cut had
+        // the operands reversed, so only the always-advancing `Less`/`Equal` arms
+        // ran and the test would have passed on buggy 0.3.2 too.)
+        //
+        // minuend: scattered depth-5 + depth-3 cells inside base-cell-0's depth-1
+        // cell 0 (no complete quartets, so they stay finer than it through
+        // normalize), plus one depth-1 cell in base cell 5 so the result is
+        // non-empty.
+        let mut fine: Vec<i64> = [0u64, 5, 17, 42, 130]
+            .iter()
+            .map(|&n| nested2mort(n, 5))
             .collect();
-        // b: many depth-5 descendants of a's coarse cells (finer than a), mixed
-        // with a few coarser cells — the exact "finer-cell-inside-coarse-full"
-        // pattern that hung 0.3.2.
-        let mut b: Vec<i64> = Vec::new();
-        for n in 0..200u64 {
-            // descendants of cell 0@2
-            b.push(nested2mort(n, 5));
-        }
-        // a coarse cell shared with a
-        b.push(nested2mort(1, 2));
-        // If the fork fix is absent this loops forever; reaching the assert proves
-        // termination. Correctness is checked against the brute-force reference.
-        let got = moc_minus(&a, &b);
-        assert_eq!(got, ref_minus(&a, &b, 5));
+        fine.push(nested2mort(3, 3)); // depth-3 cell, also inside depth-1 cell 0
+        let outside = nested2mort((5u64 << 2) + 1, 1); // depth-1 cell in base cell 5
+        fine.push(outside);
+        // subtrahend: the coarse FULL depth-1 ancestor (base-cell-0 cell 0) of
+        // every "inside" cell above.
+        let coarse = vec![nested2mort(0, 1)];
+        // Minuend finer, subtrahend coarser-full => exercises the buggy `Greater`
+        // arm. On 0.3.2 this loops forever; reaching the assert proves the fork
+        // fix. Correctness is checked against the brute-force reference.
+        let got = moc_minus(&fine, &coarse);
+        assert_eq!(got, ref_minus(&fine, &coarse, 5));
+        assert!(
+            !got.is_empty(),
+            "the `outside` cell is not subtracted, so the result must be non-empty"
+        );
     }
 
     #[test]
--- a/src_rust/src/moc.rs
+++ b/src_rust/src/moc.rs
@@ -16,12 +16,92 @@
 //! issue #28 bug).
 
 use crate::morton::{mort2nested, nested2mort};
+use healpix::bmoc::{Bmoc, MutableBmoc};
 
 /// Maximum HEALPix depth mortie supports (order 18 → 64-bit morton limit).
 /// Cells are mapped to half-open ranges over the uniform grid at this depth so a
 /// single sort linearizes ancestry: a coarser cell's range strictly contains its
 /// descendants'.
 const MAX_DEPTH: u8 = 18;
+
+// ── BMOC-backed boolean set algebra (issue #50, Option D) ──────────────────
+//
+// `moc_or`/`moc_and`/`moc_minus` are thin wrappers around the `healpix` crate's
+// BMOC `or`/`and`/`minus`.  A mortie cover is a `Vec<i64>` of signed decimal
+// morton indices at mixed orders; the BMOC works on `(depth, nested hash)`
+// pairs.  So each op follows the same decode → op → encode → normalize shape:
+//
+//   1. decode each morton i64 → `(depth, nested u64)` via `mort2nested`,
+//   2. build a `MutableBmoc` (packed/valid) per input,
+//   3. run the BMOC op,
+//   4. re-encode the result cells → morton i64 via `nested2mort`,
+//   5. `normalize` to mortie's canonical compact form.
+//
+// The decimal morton sign carries the hemisphere, which round-trips through the
+// nested hash, so southern covers need no special handling at this boundary.
+
+/// Build a packed, valid BMOC at `MAX_DEPTH` from a morton cover.
+///
+/// The BMOC ops require a canonical, non-overlapping input cover (a valid MOC):
+/// `pack()` only merges complete sibling quartets, it does not prune a cell that
+/// is a descendant of another in the same set.  `normalize` first delivers
+/// exactly that — sorted, sibling-merged, ancestor-pruned — so a raw mixed-order
+/// cover with overlaps is made valid before it reaches the BMOC.
+fn build_bmoc(morton: &[i64]) -> MutableBmoc {
+    let canonical = normalize(morton);
+    let mut builder = MutableBmoc::<false>::with_capacity(MAX_DEPTH, canonical.len());
+    for &m in &canonical {
+        let (nested, depth) = mort2nested(m);
+        builder.push_unchecked(depth, nested, true);
+    }
+    builder.into_packed()
+}
+
+/// Re-encode a BMOC's cells back to a normalized mortie morton cover.
+fn bmoc_to_morton(bmoc: MutableBmoc) -> Vec<i64> {
+    let cells: Vec<i64> = bmoc
+        .into_cells()
+        .map(|c| nested2mort(c.hash, c.depth))
+        .collect();
+    normalize(&cells)
+}
+
+/// Union (logical OR) of two morton covers.
+///
+/// Equivalent to `normalize(concat(a, b))`; backed by BMOC `or`.
+pub fn moc_or(a: &[i64], b: &[i64]) -> Vec<i64> {
+    if a.is_empty() {
+        return normalize(b);
+    }
+    if b.is_empty() {
+        return normalize(a);
+    }
+    bmoc_to_morton(build_bmoc(a).or(&build_bmoc(b)))
+}
+
+/// Intersection (logical AND) of two morton covers.
+///
+/// Backed by BMOC `and`.
+pub fn moc_and(a: &[i64], b: &[i64]) -> Vec<i64> {
+    if a.is_empty() || b.is_empty() {
+        return Vec::new();
+    }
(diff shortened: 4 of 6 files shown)


src_rust/src/moc.rs
```diff
@@ -16,13 +16,93 @@
...
+///
+/// The BMOC ops require a canonical, non-overlapping input cover (a valid MOC):
+/// `pack()` only merges complete sibling quartets, it does not prune a cell that
+/// is a descendant of another in the same set.  `normalize` first delivers
+/// exactly that — sorted, sibling-merged, ancestor-pruned — so a raw mixed-order
+/// cover with overlaps is made valid before it reaches the BMOC.
+fn build_bmoc(morton: &[i64]) -> MutableBmoc {
+    let canonical = normalize(morton);
+    let mut builder = MutableBmoc::<false>::with_capacity(MAX_DEPTH, canonical.len());
+    for &m in &canonical {
```
espg, line 53: 🤖 *from Claude (review)*

Confirmed correct. `MutableBmoc::with_capacity(max_depth, capacity)` sets the BMOC's `max_depth` field; cells are then encoded relative to it (`encode_raw_value` does `hash <<= 1 + ((max_depth - depth) << 1)`), so the only hard requirement is `depth <= max_depth` for every pushed cell — `max_depth - depth` must not underflow. Mortie's morton caps at order 18 (`fast_norm2mort_scalar` panics for order > 18, `mort2nested` decodes <= 18 digits), so `MAX_DEPTH = 18` is both correct and the tightest legal choice: it dominates every cell depth and matches across both operands so the merge aligns. It does **not** cap or misplace deeper cells — there are none. The `or`/`and`/`minus` results carry `max(lhs, rhs).max_depth == 18`, so `into_cells()` decodes back to the right depths. No issue.

---

src_rust/src/moc.rs
```diff
@@ -16,13 +16,93 @@
...
+
+/// Build a packed, valid BMOC at `MAX_DEPTH` from a morton cover.
+///
+/// The BMOC ops require a canonical, non-overlapping input cover (a valid MOC):
+/// `pack()` only merges complete sibling quartets, it does not prune a cell that
+/// is a descendant of another in the same set.  `normalize` first delivers
+/// exactly that — sorted, sibling-merged, ancestor-pruned — so a raw mixed-order
+/// cover with overlaps is made valid before it reaches the BMOC.
+fn build_bmoc(morton: &[i64]) -> MutableBmoc {
+    let canonical = normalize(morton);
```
espg, line 51: 🤖 *from Claude (review)*

The normalize-on-build rationale checks out, and I verified the BMOC precondition is fully satisfied. After `normalize`, the cover is sorted, ancestor-pruned, sibling-merged, and disjoint (the range-sweep prunes any cell contained in another, and the stack-merge collapses complete quartets) — which is exactly a valid MOC. The doc's point that `pack()` alone is insufficient is correct: `pack()` only merges complete sibling quartets and would not prune a descendant of a coarser cell in the same set, so feeding a raw overlapping mixed-order cover straight to `into_packed()` could leave an invalid (overlapping) BMOC. Normalizing first is therefore load-bearing, not belt-and-suspenders.

One worth-noting invariant (not a defect): every cell is pushed with `is_full = true`, so both operands are all-full. That is what lets `bmoc_to_morton` discard the per-cell full/partial flag safely — `and`/`or`/`minus` only ever emit partial cells when an input cell is partial, which never happens here. If a future caller ever built operands with partial cells, dropping the flag in `into_cells().map(|c| nested2mort(c.hash, c.depth))` would silently treat a partial cell as full. Fine as-is given the all-full contract; a one-line comment on that assumption in `bmoc_to_morton` would help the next reader.

---

src_rust/src/moc.rs
```diff
@@ -16,13 +16,93 @@
...
+}
+
+/// Re-encode a BMOC's cells back to a normalized mortie morton cover.
+fn bmoc_to_morton(bmoc: MutableBmoc) -> Vec<i64> {
+    let cells: Vec<i64> = bmoc
+        .into_cells()
+        .map(|c| nested2mort(c.hash, c.depth))
+        .collect();
+    normalize(&cells)
+}
```
espg, line 71: 🤖 *from Claude (review)*

The trailing `normalize` in `bmoc_to_morton` is load-bearing, not redundant — but only in one direction. The BMOC ops already return a packed, sibling-merged, disjoint result, so re-normalizing won't *prune* or *merge* anything further. What it does provide is the **sort/canonical ordering mortie's contract promises** (`normalize` ends with `out.sort_unstable()`), since `into_cells()` yields cells in BMOC entry order, not sorted morton order. The empty short-circuits are each correct and consistent with this non-empty path: every early return also routes through `normalize(...)` (or returns the canonical empty `Vec::new()`), so an empty operand produces exactly what the full path would for that case. Verified all six branches across `moc_or`/`moc_and`/`moc_minus` match the BMOC-path semantics (`a ∪ ∅ = a`, `a ∩ ∅ = ∅`, `a \ ∅ = a`, `∅ \ a = ∅`). Good.

---

src_rust/src/moc.rs
```diff
@@ -16,13 +16,93 @@
 //! issue #28 bug).
 
 use crate::morton::{mort2nested, nested2mort};
+use healpix::bmoc::{Bmoc, MutableBmoc};
```
espg, line 19: 🤖 *from Claude (review)*

`Bmoc` here is **not** an unused import (no clippy `unused_imports` risk): it is the trait that provides the `.or()` / `.and()` / `.minus()` methods used in `moc_or`/`moc_and`/`moc_minus`. Those methods live on `trait Bmoc` (with `where Self: Sized`), implemented for `MutableBmoc<true>` — the type returned by `into_packed()`. The trait must be in scope for the method calls to resolve, so the two-name import is correct and necessary. Confirmed against the pinned fork source.

---

src_rust/src/moc.rs
```diff
@@ -323,4 +403,186 @@ mod tests {
             assert_eq!(normalize(&cells), normalize_reference(&cells));
         }
     }
+
+    // ── BMOC set-algebra tests (issue #50) ─────────────────────────────────
+
```
espg, line 412: 🤖 *from Claude (review)*

**Finding (missing-coverage / mislabeled test) — `test_minus_mixed_order_terminates_regression` does not construct the trigger condition it claims to guard.**

The 0.3.2 hang lives in `minus`'s `Ordering::Greater` arm (`l.depth > r.depth` — i.e. **left operand finer, right operand coarser**, left a descendant of a coarse full right cell). I verified the fix in the pinned fork is exactly the new terminal `else { left = it_left.next(); }` in that `Greater` arm (`src/bmoc/ops.rs::minus`, ~L425), and the stray `println!("not()")` is also gone.

But in this test, `a` (the **left** operand of `moc_minus(a, b)`) holds only depth-2 cells, while `b` (right) after `normalize` holds depth-2 cells plus a handful of depth-5 leaves. So at every step `l.depth <= r.depth`: equal-depth pairs hit `Ordering::Equal` and finer-right pairs hit `Ordering::Less` — both of which always advance a cursor and were never buggy. The `Greater` arm is **never reached**, so this test passes on *buggy 0.3.2 too* and proves nothing about the fix. To actually trigger the hang, the finer operand must be the **left** argument — e.g. swap to `moc_minus(b, a)`, or make `a` the deep cover and `b` the coarse full cover. Suggest fixing the construction so it fails (hangs) without the fork patch.

Mitigating: `test_setops_match_reference_fuzz` (300 random mixed-order pairs, independent `a`/`b`) *does* hit the `Greater` arm with high probability and is the real, if probabilistic, guard against the hang — so the fix is not untested, just not deterministically pinned by the dedicated regression test.

On the oracle itself: `setop_reference` is sound — densify both to a common deep `order`, do the set op on flat leaf sets, then `normalize`. Because `to_order` coarsens any cell finer than `order` (lossy), the comment's contract "`order` must be >= the deepest input cell" is the real correctness condition, and every caller honors it (fixed-`order` tests use a literal >= their max depth; the fuzz uses `da.max(db)`). Exact. Edge cases covered well: cross-base-cell (southern + fuzz over 12 bases), single-element and disjoint covers, empty inputs, self-ops, negative-morton/southern round-trip. Two gaps worth a line each if you want belt-and-suspenders: an **order-0 (whole base cell) operand** is never exercised in the set-op tests (it is in morton roundtrip tests, so encoding is fine), and no assertion that the deterministic regression case reaches the `Greater` arm (covered by the fix above).

---

Cargo.toml
```diff
@@ -15,7 +15,7 @@ path = "src_rust/src/lib.rs"
 pyo3 = { version = "0.22", features = ["extension-module", "abi3-py310"] }
 numpy = "0.22"
 rayon = "1.10"
-healpix = "0.3.2"
+healpix = { git = "https://github.com/espg/healpix-rs", rev = "9118d3834b6c2a8860dc7a375d46df2d51516df9" }
```
espg: 🤖 *from Claude (review)*

Dependency swap confirmed in-scope and correct. This is the same `healpix` crate (not a second/duplicate HEALPix), repointed from the published `0.3.2` to @espg's fork `github.com/espg/healpix-rs`, git-pinned by `rev = 9118d38…` — explicitly approved in issue #50 ("modifying python and cargo toml's ... dependency swaps for the patched 0.3.3 rust crate; this is expected and approved"). I checked out the pinned rev: it is version `0.3.3`, contains the `minus` `Greater`-arm fix, and the `println!("not()")` debug line is removed. No *other* new dependency was introduced (only this rev change in `[dependencies]`). Pinning by exact rev (rather than a tag/branch) is the right call for an unmerged fork — reproducible and immune to fork-branch movement. One forward-looking note (non-blocking, @espg's call): once the upstream PR to `matt-cornell/healpix-rs` merges, this should move back to a crates.io version pin to drop the git dependency.

---

mortie/coverage.py
```diff
@@ -254,3 +254,84 @@ def moc_to_order(morton, order):
...
+    ----------
+    a, b : array_like
+        Morton covers (mixed order allowed).
+
+    Returns
+    -------
+    numpy.ndarray
+        Sorted, compacted union (``int64``).
+
+    See Also
```
espg, line 275: 🤖 *from Claude (review)*

Public API surface checks out. The three Python wrappers mirror the existing `moc_to_order`/`compress_moc` pattern exactly (`np.asarray(..., dtype=np.int64).ravel()` → `_rustie.rust_moc_*` → `np.asarray`), names line up cleanly Python `moc_or/and/minus` ↔ Rust `rust_moc_or/and/minus` ↔ `moc::moc_or/and/minus`, and both the `from .coverage import (...)` block and `__all__` in `mortie/__init__.py` are updated consistently. Docstrings are accurate: `moc_or`'s stated equivalence `moc_or(a, b) == compress_moc(concatenate([a, b]))` is exactly what `test_or_equals_compress_concat` / the Rust `test_or_equals_normalize_concat` assert, and `moc_minus` correctly documents `a \ b`.

One minor nit (non-blocking): each wrapper does a function-local `from . import _rustie`. That matches `moc_to_order` immediately above, so it is consistent with the surrounding file — keep it. Flagging only so it is a conscious choice rather than an oversight.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the BMOC-backed `moc_or`/`moc_and`/`moc_minus` (issue #50, Option D, patched `healpix` fork). I read the diff, the full `moc.rs`/`morton.rs` on the branch, and the pinned fork rev (`9118d38`, v0.3.3) source. Overall the implementation is correct and well-tested; one test is mislabeled and doesn't guard what it claims.

**Verified correct**
- **normalize-on-build** is load-bearing and fully establishes the BMOC's valid-MOC precondition (sorted, ancestor-pruned, sibling-merged, disjoint). `pack()` alone would not prune a descendant of a coarser cell, so the doc's rationale is sound.
- **`MutableBmoc::with_capacity(MAX_DEPTH, len)`** — `MAX_DEPTH = 18` is correct and tightest: it dominates every cell depth (morton caps at order 18) and matches across operands; no deeper-cell capping/misplacement.
- **decode→op→encode→normalize bridge** round-trips both hemispheres; the trailing `normalize` is load-bearing for canonical sort order; all six empty short-circuits match the non-empty path.
- **Fork dependency swap** is the approved patched `healpix` (same crate, git-pinned rev, v0.3.3) — fix present, `println!("not()")` removed, no other new dep.
- **Public API** (Python ↔ Rust naming, `__all__`, docstrings) consistent and accurate.
- `Bmoc` import is the trait providing `.or/.and/.minus` — not unused.

**Findings to fold in (diff-scoped, author can address)**
1. **`test_minus_mixed_order_terminates_regression` does not trigger the bug it guards.** The hang is in `minus`'s `Greater` arm (finer **left**, coarser full **right**); this test's left operand (`a`) is the coarse one, so only the never-buggy `Less`/`Equal` arms run — it passes on buggy 0.3.2 too. Reconstruct so the finer cover is the left arg (e.g. `moc_minus(b, a)`). The fuzz test does cover the arm probabilistically, so the fix isn't untested — but this dedicated regression is ineffective as written. (inline)
2. **Minor, optional:** add a one-line note in `bmoc_to_morton` that dropping the full/partial flag is safe only because all built cells are `is_full = true`; and consider an order-0 (whole base cell) operand in the set-op tests. (inline)

No finding implies a side-effecting directive — all are ordinary author fixes within this PR's scope. The fork-pin → crates.io migration after the upstream PR merges is a future @espg decision, noted only.

Event = COMMENT (no approve/request-changes).

---

espg:
🤖 *from Claude*

Phase 4 (`6b01e18`) — folds the fresh-context self-review's findings (the review posted above as a COMMENT).

**Finding 1 (real defect, fixed) — the `minus` regression test wasn't actually triggering the 0.3.2 hang.** The hang lives in `minus`'s `Greater` arm, which fires only when a *finer* LEFT (minuend) cell is a descendant of a coarser *full* RIGHT (subtrahend) cell. The original `test_minus_mixed_order_terminates_regression` had the operands reversed — it called `moc_minus(coarse, fine)`, so only the always-advancing `Less`/`Equal` arms ran; it would have passed on **buggy** 0.3.2 too, proving nothing about the fork fix. Rewrote it so the minuend is the finer cover (scattered depth-5 + depth-3 cells with no complete quartets, so they stay finer than the subtrahend through `normalize`, plus one depth-1 cell in another base cell so the result is non-empty) and the subtrahend is their coarse full depth-1 ancestor. That construction exercises the formerly-buggy `Greater` arm; the test now genuinely pins the fix. Added a `!got.is_empty()` assertion alongside the brute-force-reference equality.

```rust
let got = moc_minus(&fine, &coarse);   // minuend finer, subtrahend coarser-full
assert_eq!(got, ref_minus(&fine, &coarse, 5));
assert!(!got.is_empty(), "the `outside` cell is not subtracted ...");
```

Verified: `cargo test --release moc::` → **21 passed, 0 failed**, including the corrected regression test (it terminates and matches the reference) and the 300-iteration fuzz. (The fuzz already hit the `Greater` arm probabilistically; the regression test now pins it deterministically.)

**Finding 2(a) (tightening, done)** — added a comment in `bmoc_to_morton` documenting that discarding each cell's full/partial flag is safe because every cell mortie pushes is `is_full = true`, so the BMOC only emits full cells.

**Finding 2(b) (deferred, standing for your call)** — the reviewer suggested adding an order-0 (whole base cell) operand to the set-op tests. I left this out of this fold: order-0 base-cell encoding round-tripping interacts with the order-range standardization under discussion on #48, and I'd rather not expand this PR's test surface into that until the order/encoding direction lands there. Flagging it rather than acting.

`moc.rs` is `cargo fmt --check`-clean (the only fmt drift in the tree is pre-existing in `benches/coverage_bench.rs`, left for the #34 sweep). The four **Questions for review** in the PR description (pin by `rev` vs a fork tag; whether to also expose `not`/`xor`; Rust internal fn naming; normalize-on-build) are still open for your call — marking `waiting` on those. PR stays draft.

---

mortie/__init__.py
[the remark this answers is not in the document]
espg: 🤖 *from Claude*

Leaving this `I001` as-is (same call as the identical finding on #52, https://github.com/espg/mortie/pull/52#discussion_r3432792735):

1. mortie's lint contract is **flake8**, not ruff/isort (CLAUDE.md §7: "There is no `black`/`ruff`/`mypy` configured"). The gating flake8 job passes; reviewdog's `I001` is advisory and not the project's style authority.
2. This PR did not introduce the disorder — `mortie/__init__.py` uses intentional *logical* grouping with section comments (`# Import coverage functions`, etc.), not alphabetical isort order, and that predates this PR tree-wide. #53 only **appends** `moc_or`/`moc_and`/`moc_minus` to the existing `.coverage` import block (and `__all__`) in their natural or → and → minus order.
3. Satisfying `I001` would require reordering the whole import section, churning unrelated lines and breaking the logical grouping/comments — out of scope for a set-algebra PR.

Flagging rather than silencing — a tree-wide import-sort pass would pair naturally with the queued fmt/clippy sweep on #34 if you want one.

---

espg:
> One forward-looking note (non-blocking, @espg's call): once the upstream PR to matt-cornell/healpix-rs merges, this should move back to a crates.io version pin to drop the git dependency.

fixed upstream, and posted on [crates.io at version 0.3.3](https://crates.io/crates/healpix)

swap us back to the upstream version

espg:
🤖 *from Claude*

Phase 5 (`90c7a62`) — swapped the `healpix` dependency from the git-fork pin back to the published crate, per your note that the `minus` fix landed upstream and shipped as crates.io `healpix` 0.3.3.

```diff
-healpix = { git = "https://github.com/espg/healpix-rs", rev = "9118d38..." }
+healpix = "0.3.3"
```

This resolves **Question (1)** (rev-vs-tag pin) — the git dependency is gone entirely; we're back on a normal crates.io semver pin.

**Verified the published 0.3.3 carries the fix** (not just the fork):
- `cargo update -p healpix` → resolves to `healpix v0.3.3` from crates.io, fork rev removed from the lockfile.
- `cargo test --release` → **154 passed, 0 failed**, including `test_minus_mixed_order_terminates_regression` (the mixed-order `minus` that hung on 0.3.2 now terminates and matches the brute-force reference) and the 300-iteration fuzz.
- Rebuilt the extension against 0.3.3 and ran the full Python suite → **290 passed, 9 skipped**; `moc_minus` on a finer-vs-coarser mixed-order pair returns correctly (no hang).

No source changed — only the dependency source moved from the fork to the identical published crate. `Cargo.lock` is untracked, so the diff is the single `Cargo.toml` line. PR stays draft; the remaining open Questions are (2) expose `not`/`xor`? and (4) normalize-on-build (I recommend keeping it) — both also captured in the new follow-up tracking issue.

---

PR #58, opened 2026-06-19
packed-u64 migration (Option A) + norm2mort consolidation

### LINKED ISSUE #56: Cut fastNorm2Mort, VaexNorm2Mort
...seems like we probably don't need these anymore. We have geo2mort & mort2norm... so presumably we just need one function that's norm2mort ... which will probably be pretty fast given that we it's going to be rust backed.

espg:
Refs #48
Closes #56

## What this does

Implements the locked **Option A** decision from #48: make the packed-`uint64` `decimal_morton` kernel (`src_rust/src/decimal_morton.rs`, `MAX_ORDER = 29`) the canonical storage for the **entire** bare morton channel, with the human-readable decimal string as a render-only repr produced by decoding the word. Folds in #56 (the forward encoder is now `norm2mort(normed, parent, order)`, the exact inverse of `mort2norm`; `VaexNorm2Mort` deleted; hard-cut, no shim).

Phase 5 carries the decision on #58 to its conclusion: the public morton **value dtype is now native `uint64`** end-to-end, not an `int64` bit-reinterpretation of the word.

## The atomic flip (why one channel, one move)

The legacy decimal morton and the packed word share one untyped 64-bit morton channel with the whole coverage / MOC / buffer / linestring / prefix-trie subsystem, so the format had to flip everywhere at once. The flip is concentrated at two bridge functions — `morton::mort2nested` / `morton::nested2mort` (now thin pass-throughs over `decimal_morton::to_nested` / `from_nested`) — through which `geo2mort`, coverage, MOC, buffer, and linestring all already routed; plus the prefix-trie parser onto the kernel's repr.

## Phases

- [x] **Phase 1 — converter, decimal repr, golden fixtures** (`3241623`/`0f00a5f`).
- [x] **Phase 2 — the atomic whole-system flip** (`4414c14`).
- [x] **Phase 3 — delete legacy machinery** (`f8a8e58`).
- [x] **Phase 4 — conformance** (`b0a23b6`).
- [x] **Phase 5 — native `uint64` public morton dtype** (`cbfa4e3`/`26ceb33`): switched the morton **value** dtype from `int64`-bit-reinterpretation to native `uint64` everywhere on the public bare-morton channel and the skins' numpy interop. Base cells 7-11 (prefix 8-12) now read as large positive `u64` words (bit 63 set) instead of negative ints. Because the word is unsigned, **the kernel's "raw word sort == Z-order" property now holds for free**, so `moc.rs`'s `out.sort_unstable()` on `Vec<u64>` is the unsigned Z-order with no sign special-casing — the signed-sort wrinkle is gone (a new `moc.rs` test asserts a cover spanning northern and southern base cells comes back in ascending unsigned/Z-order). Concretely:
  - **`src_rust/src/lib.rs`**: morton-array params `PyReadonlyArray1<i64>` → `PyReadonlyArray1<u64>`; morton outputs returned as `uint64` arrays — `rust_mort2nested`, `rust_nested2mort` (output), `rust_geo2mort`, `norm2mort` path, the inverse-API bindings (`rust_mi_*`), coverage/MOC/buffer/linestring bindings, `split_children_rust`. **Exception:** `rust_mi_from_legacy`'s INPUT stays `i64` (legacy decimal values were genuine signed i64); only its OUTPUT becomes `u64`.
  - **`src_rust/src/morton.rs`**: `mort2nested`/`nested2mort` take/return `u64` directly — the `i64`↔`u64` reinterpret layer is removed.
  - **`src_rust/src/{moc,buffer,linestring,coverage,prefix_trie}.rs`** and **`geo2mort.rs`**: carry morton as `u64` (e.g. `Vec<u64>`, `geo2mort_scalar -> u64`). Non-morton arrays unchanged (lat/lon `f64`, HEALPix nested ids `u64`, depths `u8`, permutation buffers `i64`).
  - **Python (`tools.py`, `coverage.py`, `linestring.py`, `prefix_trie.py`, `morton_index.py`, `arrow.py`)**: morton arrays use `np.uint64`; scalars return `np.uint64`. The `morton_index` pandas skin now stores `uint64` (`type=np.uint64`, `kind="u"`, sort/compare on the raw word) and the Arrow skin's storage type is `pa.uint64()`.

## Representation decision

The public bare morton channel is now the packed word as a **native `uint64`** everywhere. Base cells 7-11 read as large positive `u64` values (bit 63 set), replacing the previous "southern = negative `i64`" presentation. The frozen decimal *repr* (`to_decimal_repr`, including its leading `-` for southern cells) is **unchanged** — see Question (4) for the now-settled decision to keep it signed.

## How it was tested

Built in the worktree venv (`maturin develop --release`):

```
cargo test --lib                       -> 157 passed (155 baseline + 2 new u64 Z-order tests)
cargo clippy --all-targets             -> no new warnings (36 == baseline; pre-existing useless_conversion + test-only lints unchanged)
cargo fmt --check                      -> clean
flake8 mortie --select=E9,F63,F7,F82   -> clean
pytest -q                              -> 388 passed, 8 skipped
```

The packed golden fixture (`packed_u64_golden.json`) was regenerated: the **only** record field that changed is `word` (450/1080 records — exactly base cells 7-11 — flip from negative `i64` to the positive `u64` value); `decimal_repr`, `nested`, `depth`, `base_cell`, `order`, `tuples` are all unchanged. The `Ant_…_morton.npz` regression fixture was refreshed `int64` → `uint64` (bit pattern identical; `old.view(uint64) == new` verified).

## Questions for review

1. **`generate_morton_children` / `clip2order` semantics** — unchanged from Phase 2 (`clip2order(..., print_factor=True)` returns the level count `18 - clip_order`).
2. **`norm2mort` argument order** — `norm2mort(normed, parent, order)` per #56.
3. **`morton_index` pandas storage** — switched to native `uint64` (`type=np.uint64`, `kind="u"`) rather than keeping `int64`-bits and converting at the numpy boundary; the all-zero NA sentinel and pandas round-trips are preserved. Flag if you'd prefer the internal storage stay `int64`-bits.
4. **Decimal repr sign — RESOLVED: keep the signed scheme.** The decimal repr keeps its leading `-` for southern cells (north `1`–`6`, south `-1`–`-6`); `to_decimal_repr` is left unchanged. Rationale (it's functional, not cosmetic): the repr doubles as an S3/Hive partition key, and its single-character tokens make the *lexicographic string* order Z-ordered within a hemisphere, prefix-scannable (a cell's string is a prefix of its descendants'), and unambiguously parseable. An unsigned `1`–`12` lead would break all three — two-digit leads for bases 9-11 (`10`/`11`/`12`) interleave into base-0's subtree lexicographically and reintroduce parse ambiguity. Full design record: https://github.com/espg/mortie/pull/58#issuecomment-4755980331


espg (commented):
🤖 *from Claude (review)*

Adversarial review of Phase 1 (additive: `to_decimal_repr` + `from_legacy_decimal` kernel functions, the two PyO3 bindings, the Python skin, and the 360-record golden freeze). I checked every claim against the surrounding kernel (`decode`, `from_nested`, `order_of`, `morton::mort2nested`, `fast_norm2mort_scalar`) and against issue #48.

**Correctness — verified sound, no bug found.**
- `to_decimal_repr` (decimal_morton.rs:494): `decode` returns exactly `order` tuples, order-1 first (MSB-first), which is the same digit order `fast_norm2mort_scalar` emits (its `for i in (1..=order).rev()` loop), so the per-order digits line up. The leading-digit math (`base+1` north / `base-5` south, `-` for bases 6-11) is the exact inverse of `mort2nested`'s parent extraction (morton.rs:136-142: north `first_digit-1`, south `first_digit+5`). Order 0 correctly degrades to the bare base-cell digit. No off-by-one at the 5/6 boundary. This is anchored by an *independent* check against `str(legacy_i64)` over all 12 bases × orders 0-18, in both the Rust test (`decimal_repr_matches_legacy_decimal_orders_0_to_18`) and the Python test — that is the right way to pin it.
- `from_legacy_decimal` (decimal_morton.rs:539): pure composition `mort2nested` → `from_nested`; sign/hemisphere handling lives entirely in `mort2nested` (no special-casing here, correctly). The "never lossy for order ≤ 18" claim holds — legacy tops out at order 18 (`fast_norm2mort_scalar` panics above 18), well inside the 27-tuple body, so the result is always `Kind::Area`, asserted in `from_legacy_decimal_matches_kernel_encode`. The `legacy == 0` panic is documented and delegated to `mort2nested`'s existing panic.

**Findings (all non-blocking) are in the inline comments below.** Summary:
1. The headline `test_golden_fixture_matches_kernel` is, by construction, a drift detector — it regenerates from the same kernel it asserts against, so it cannot catch a *logic* error in encode/repr, only an accidental future change. The real correctness anchor is the legacy cross-check, which only covers orders 0-18. Orders 19-29 have *no* independent oracle (no legacy path exists), only a structural-shape check. That's inherent, but the fixture's framing as a "frozen contract" overstates what the 19-29 rows guarantee.
2. Freeze breadth: one tuple-seed per (base, order) = 360 rows. For a 1.x freeze this is thin — it does not pin boundary tuples (all-1s, all-4s, mixed 28/29 tail). Worth widening before this is truly "frozen."
3. Minor: `& 3` mask in the digit push is dead (decode already guarantees 0..=3); the `into_any().unbind()` on the PyList may trip the `useless_conversion` clippy lint the issue mentions.

No convention violations: decimal_morton.rs stays well under ~1000 lines of new code, docstrings match the surrounding freeform style, no tests weakened, no new deps, no dead/speculative code beyond the nit below. Posting as COMMENT only.

---

espg (commented):
🤖 *from Claude (review)*

Inline detail for Phase 1 (the MCP tooling could not submit these as line-anchored comments — `submit_pending` kept resolving to the already-submitted summary review — so the four findings are inlined here against their file:line. All non-blocking; no correctness bug.)

**1. `mortie/tests/test_packed_golden.py:80` `test_golden_fixture_matches_kernel` — drift detector, not a correctness oracle.**
`_regenerate_records()` (line 43) calls the exact same kernel bindings (`rust_mi_from_nested`, `rust_mi_decimal_repr`, `rust_mi_to_nested`) that produced the committed JSON, then asserts they still match. If the kernel's encode/repr logic were *wrong*, both sides would be wrong identically and this passes green. It catches accidental future kernel change, not incorrect values. The real correctness anchor is `test_legacy_converter_matches_repr_orders_0_to_18` (independent `fastNorm2Mort` path) — but that only reaches order 18. Orders 19-29 have no independent oracle anywhere in this PR (`test_golden_decimal_repr_shape` checks length/sign/digit-range only). Inherent (no legacy path exists up there), but the fixture's "frozen 1.x contract" framing overstates what the 19-29 rows guarantee; recommend saying so in the fixture `note`.

**2. `mortie/tests/test_packed_golden.py:31` `_sample_tuples` — freeze breadth is thin.**
One deterministic tuple vector per (base, order) = 360 rows pins a single path per cell. It never exercises the values most likely to expose a packing bug: an all-`0` (digit `1`) path, an all-`3` (digit `4`) path, and the order-28/29 tail tuples in both positions (the only orders where tuples live in the suffix, not the body). A future kernel change could corrupt 28/29 tail handling and still match this fixture if the one sampled seed doesn't populate those pairs distinctly. Suggest widening to {all-1s, all-4s, current pseudo-random} per (base, order) before calling it frozen.

**3. `src_rust/src/decimal_morton.rs:510` `char::from(b'1' + (t & 3))` — dead mask.**
`t` comes from `decode`, which already extracts every tuple as `(word >> shift) & 3` (line 307) and the 28/29 tail via `decode_tail`, so every element is already `0..=3`. The `& 3` can never change the value — dead code per the convention. Either drop it (`b'1' + t`) or note in the comment it is a defensive assertion. Mapping itself is correct: max `t==3` → `b'4'`, always ASCII `'1'..'4'`.

**4. `src_rust/src/lib.rs:1116`/`1137` the two bindings — three confirmations (all fine).**
(a) Asymmetry is correct: `rust_mi_from_legacy` wraps the map in `catch_unwind` because `from_legacy_decimal` → `mort2nested`/`from_nested` can panic (zero/malformed legacy); `rust_mi_decimal_repr` does not, because `to_decimal_repr` returns `Option` and never panics — empty/invalid flows through `ok_or_else(...)` → `ValueError`, matching its docstring. (If `decode` ever gains a panic path, the repr binding would need the guard too.)
(b) `rust_mi_decimal_repr` serial `.iter()` over `par_iter()` is a reasonable deliberate choice: tiny per-element work, and the `Result` collect short-circuits deterministically on the first bad word in index order (rayon would error nondeterministically).
(c) `into_any().unbind()` (and `into_any()` in `rust_mi_from_legacy`) are the `useless_conversion` clippy warnings issue #48 enumerates; since the fmt/clippy sweep is sequenced as a separate post-merge PR, leaving them is consistent — just confirming these are the known lint, not new ones.

---

espg:
🤖 *from Claude*

Folded the Phase-1 review (self-review + ruff bot) into `3241623`. What I addressed:

- **ruff `I001` (unsorted imports, `test_packed_golden.py`)** — fixed: imports are now sorted (stdlib, then `numpy`, then the two `mortie` imports). flake8 `--max-line-length=88` clean on the file.
- **Dead `& 3` mask in `to_decimal_repr` (`decimal_morton.rs`)** — removed; `decode` already guarantees every tuple is `0..=3`, so the digit push is now `char::from(b'1' + t)` with a clarifying comment.
- **Freeze breadth too thin (one tuple-path per cell)** — widened the golden fixture from 360 to **1080 records**: each (base cell, order) is now pinned for three tuple variants — `all1` (digit 1), `all4` (digit 4), and a pseudo-random seed — so the digit extremes and the order-28/29 suffix-tail tuples are exercised, not just one path.
- **"Frozen contract" framing overstated the 19-29 rows** — clarified in both the test module docstring and the fixture `note`: `test_golden_fixture_matches_kernel` is a *drift* detector; the independent correctness anchor (the legacy cross-check) covers orders 0-18 only, and 19-29 are pinned to the kernel's natural extension with a structural-shape check (length / sign / 1-4 digits), no external oracle.

Left as-is (documented in the PR body): the two new bindings' `useless_conversion` clippy false-positive on `?` over a pyo3 `Result`, matching the crate-wide pattern (#57) — no blanket `#[allow]` added.

Local after the fold: `cargo test --lib` 161 passed; `pytest -q` 303 passed / 8 skipped; flake8 clean.

This PR stays `waiting` on the Phase-2 scope decision (Question (1) in the PR body — the atomic whole-system `i64` wire-format flip).

---

espg:
> Therefore the flip (Phase 2 below) must land in one atomic PR that re-points geo2mort, norm2mort, mort2norm, the whole tools.py inverse API, and coverage/MOC/buffer/linestring + a rewrite of the prefix-trie's decimal-digit parser onto the kernel, all together. 

YES. Do this

espg:
🤖 *from Claude*

**Build fixed (it was a staleness break, not a logic bug), and Phase 2 is now unblocked by your "YES."**

### Why CI went red
The branch was cut from the pre-#57 `main`. While it sat in review, #57 merged to `main` (`4469649`), which reworked `panic_msg` to take a second `fallback: &str` argument. CI builds the PR *merged into current `main`*, so the 2-arg definition from `main` met this PR's new `rust_mi_from_legacy` binding, which still called the old 1-arg `panic_msg(e)` → `error[E0061]` at `lib.rs`. That single compile error cascaded to every job that builds the extension (all three `test` jobs, all wheel builds, both benchmark jobs); only `ruff` and the sdist build (which don't compile) stayed green.

### Fix
Merged `origin/main` into the branch (clean auto-merge, no force-push — `86312d3`) and updated the new binding to `panic_msg(e, "mi_from_legacy panicked")`, matching the convention #57 established (`0f00a5f`). Local on the merged tree:
- `cargo test --lib` → **164 passed**
- `flake8 mortie --select=E9,F63,F7,F82` → clean
- `pytest -q` → **313 passed, 8 skipped**

CI is re-running on `0f00a5f`.

### Next
Per your [go-ahead](https://github.com/espg/mortie/pull/58#issuecomment-4755334236) on Question (1), I'm clearing `waiting` and proceeding with **Phase 2 — the atomic whole-system `i64` wire-format flip** (`geo2mort`/`norm2mort`/`mort2norm`/the `tools.py` inverse API **plus** coverage/MOC/buffer/linestring + the prefix-trie decimal-digit parser, all onto the packed-u64 kernel in one commit), then Phase 3 (delete the legacy `POWERS_OF_10`/string-scan machinery) and Phase 4 (conformance fixtures). Staying a draft; per-phase self-review continues.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of the atomic packed-u64 wire flip (commits 4414c14, f8a8e58, b0a23b6). Overall the flip looks **correct**: `mort2nested`/`nested2mort` are a faithful reinterpret over `decimal_morton::{to,from}_nested` (the `morton as u64` / `as i64` round-trip is total and lossless for every bit pattern, including the sign-bit-set southern words); the inlined `legacy_to_nested` reproduces the deleted digit-scan decoder exactly (order = `ilog10`, same digit→2-bit `digit-1` LSB-first loop, same parent/sign fold) and is cross-checked against the new test-only `legacy_encode`; `mort2nested(0)` still raises as before; `norm2mort`'s `uint64` shift is overflow-safe at order 29 (max ≈ 3.5e18 < u64 max); and `generate_morton_children`'s NESTED-space descent is sound (Python-int `<<` + `uint64` arange stays `uint64` — no float coercion or overflow in range).

A few genuine concerns below, all low-severity (one doc/test off-by-one repeated in three places, plus one design observation about sort order). Nothing blocks correctness of the flip.

**Off-by-one in the "negative base cell" boundary (3 spots, inline).** The prefix stores `base + 1`, so bit 63 (the i64 sign bit) is set when `base + 1 >= 8`, i.e. **base cells 7–11**, not 8–11. Base 6 → prefix 7 (positive i64) but its *decimal repr* still leads with `-` (southern, base ≥ 6); base 7 → prefix 8 (negative i64) but is northern in the repr sense. The docs/tests that say "8–11" are inaccurate, and the sign assertions silently skip base 7. Not a logic bug (the reinterpret is total), but the dual sign conventions — repr-sign keys on base ≥ 6, i64-sign keys on base ≥ 7 — are a latent footgun worth getting right in the prose. Flagged inline on morton.rs and the conformance test.

**moc.rs `normalize()` / `to_order()` end with a *signed* `out.sort_unstable()` on the packed i64.** This is not a correctness problem for the set operations (the BMOC path and the ancestor-prune/sibling-merge sweep both re-derive everything from unsigned `cell_range`, so dedup/canonicalization are order-independent), and it matches the legacy behavior (southern was negative there too, so no regression). But it does mean the public output is **not** in Z-order: the southern hemisphere (bit 63 set) sorts to the front. The kernel module doc sells "a raw unsigned sort is preserved as a Z-order curve" as the paramount #35 property; sorting the i64 form quietly defeats that for any consumer expecting ascending Z-order. If Z-ordered output is part of the contract, these final sorts want `sort_unstable_by_key(|&m| m as u64)`. (moc.rs isn't in this PR's diff, so raising here rather than inline — a question for review, not a blocker.)

**Test quality:** the rewritten tests largely hold their weight — `TestNorm2MortRoundTrip` / `TestMortHealpixRoundTrip` are genuine `(normed, base, order)` identities across all 12 cells and orders 0–29, and `test_morton_validation`'s switch from the now-meaningless digit check to order-mismatch + empty-sentinel checks is appropriate. One pre-existing weak test (`test_morton_sign_consistency`, untouched by this PR) asserts `np.any(m > 0) or np.any(m < 0)`, which is a tautology for any non-zero morton array — out of scope here, but worth tightening to north>0 / south<0 in a later pass now that the sign semantics changed.

---

src_rust/src/morton.rs
```diff
@@ -1,350 +1,156 @@
...
+//! tables were removed with the flip; the one-way legacy *decode* needed by the
+//! converter now lives inlined in `decimal_morton::from_legacy_decimal`.
+
+/// Decode a packed-u64 morton word (bit-reinterpreted to `i64`) into its
+/// HEALPix NESTED cell ID and depth.
 ///
-/// This is a direct port of the Python fastNorm2Mort function.
+/// The bare-`i64` morton channel carries the canonical packed word
+/// (`decimal_morton`, issue #48); this is a thin reinterpret-and-decode over
+/// [`crate::decimal_morton::to_nested`]. Base cells 8-11 set bit 63, so the word
```
espg: 🤖 *from Claude (review)*

Off-by-one in the doc: "Base cells 8-11 set bit 63" is inaccurate. The prefix stores `base + 1`, so bit 63 (the i64 sign bit) is set whenever `base + 1 >= 8`, i.e. **base cells 7-11**, not 8-11. (base 6 → prefix 7, positive i64; base 7 → prefix 8, negative i64.) Worth correcting since this is the canonical description of the sign convention and the same "8-11" claim is repeated in the PR body and the conformance test. Note the latent footgun this creates: the *decimal repr* sign keys on base ≥ 6 (southern, leading `-`) while the *i64* sign keys on base ≥ 7, so base 6 has a negative repr but a positive i64 word — the prefix_trie correctly uses the repr sign, but the two conventions disagree and the prose should make that explicit.

---

src_rust/src/morton.rs
```diff
@@ -1,350 +1,156 @@
...
+                let normed = if order == 0 {
+                    0
+                } else {
+                    0b1110 & ((1u64 << shift) - 1)
+                };
+                let nested = (parent << shift) | normed;
+                let morton = nested2mort(nested, order);
+                // Southern base cells (8..=11) set bit 63 -> the i64 is negative.
+                if parent >= 8 {
+                    assert!(morton < 0, "base {} should set the sign bit", parent);
```
espg: 🤖 *from Claude (review)*

This sign assertion only fires for `parent >= 8`, so base cell **7** — which also sets bit 63 (prefix 8) and is therefore a negative i64 — is never checked for negativity. The roundtrip identity below still covers base 7 fully, so this isn't a correctness gap in coverage, but the guard's `>= 8` perpetuates the "8-11 are the negative ones" misconception. Tightening to `parent >= 7` would both fix the boundary and document the real one.

---

mortie/tests/test_packed_conformance.py
```diff
@@ -0,0 +1,144 @@
...
+    """norm2mort <-> mort2norm is the identity for every base cell / order."""
+
+    @pytest.mark.parametrize("order", range(MAX_ORDER + 1))
+    def test_all_base_cells_both_hemispheres(self, order):
+        for base in range(12):
+            for normed in _normed_samples(order):
+                m = tools.norm2mort(normed, base, order)
+                # Southern base cells (8-11) set the i64 sign bit.
+                if base >= 8:
+                    assert int(m) < 0, f"base {base} should be negative"
```
espg: 🤖 *from Claude (review)*

Same off-by-one as in morton.rs: "Southern base cells (8-11) set the i64 sign bit" and `if base >= 8` are both narrower than reality — base cell **7** also sets the sign bit (prefix `base+1 = 8`, bit 63 set), so its negativity goes unasserted here. The `(int(n2), int(p2), o2) == (normed, base, order)` roundtrip below still covers base 7, so no correctness hole, but `if base >= 7` would assert the true sign boundary and stop propagating the "8-11" framing into the conformance contract.

---

espg (commented):
🤖 *from Claude (review)*

Inline line-anchored detail for the three findings called out in my summary review above (the off-by-one "negative base cell" boundary, repeated in morton.rs doc + test and the conformance test). All low-severity; no correctness bug in the flip itself.

---

espg:
🤖 *from Claude*

Phases 2-4 of the atomic flip have landed (`4414c14` flip, `f8a8e58` legacy deletion, `b0a23b6` conformance), plus `d4c49af` (ruff bot) and `d2b125a` (self-review fold). CI is green on the test matrix (3.10/3.11/3.12), ruff, wheels, and both benchmark jobs. Local: `cargo test --lib` 155 passed, `pytest -q` 388 passed / 8 skipped, `cargo clippy`/`cargo fmt --check`/`flake8 --select=E9,F63,F7,F82` clean.

**What I folded from the fresh-context self-review** (it judged the flip correct; findings were low-severity):

- **base-cell sign boundary was off-by-one (real doc/test bug, fixed).** The prefix stores `base+1`, so the i64 sign bit (bit 63) is set for prefix >= 8, i.e. **base cells 7-11**, not 8-11. The conformance test's `if base >= 8: assert m < 0` silently skipped base 7; it now asserts `>= 7`, and the southern-roundtrip Rust test now asserts `parent >= 7`. Corrected the "8-11" prose everywhere it appeared in this PR's surface (`morton.rs`, `tools.py` `norm2mort` docstring, `USAGE.md`, `test_rust_vs_python.py`) and in the pre-existing contract docs that describe exactly this wire format (`morton_index.py`, `lib.rs` batch-wrapper header — comments only; the code already sorted on the `uint64` view, so behavior was always correct). Empirically confirmed: bases 7-11 decode to negative `i64`, 0-6 to positive.

**Addressed the ruff bot** (`d4c49af`): fixed every inline reviewdog finding on the diff — import sorting (I001) across `__init__.py`, the rewritten test files, and the benchmarks; stray f-strings-without-placeholders (F541); and the mid-file `_rustie` import in the regeneration script (kept with `# noqa: E402`, as it must follow the morton compute). Remaining ruff annotations are on pre-existing lines this PR did not touch.

**One finding I left standing for your call** (a design question, not a bug): `moc.rs` `normalize()`/`to_order()` end with a *signed* `out.sort_unstable()` over the packed `i64`, so the returned cover is not in unsigned Z-order — base cells 7-11 (bit 63) sort to the front. This is **not** a correctness problem (the set algebra re-derives everything from unsigned `cell_range`; dedup/canonicalization are order-independent) and it's no regression (the legacy southern words were negative too, so the old output had the same property). But it does mean the public cover ordering doesn't match the kernel's "raw unsigned sort == Z-order" property. `moc.rs` is outside this PR's diff, so I did not change it. If Z-ordered cover output is part of the 1.x contract, the fix is `sort_unstable_by_key(|&m| m as u64)` in those two spots — happy to do it here or in a follow-up; flagging for your decision.

Continuing to leave this a draft; not marking ready-for-review.

---

mortie/tests/test_polygon_regression.py
```diff
@@ -227,12 +230,14 @@ def test_statistics_summary(self, polygon_coordinates, reference_morton):
...
-        n_negative = np.sum(morton < 0)
-        print(f"\nSign distribution:")
-        print(f"  Positive:         {n_positive:,} ({100*n_positive/len(morton):.1f}%)")
-        print(f"  Negative:         {n_negative:,} ({100*n_negative/len(morton):.1f}%)")
+        # Hemisphere distribution: the word is uint64 (issue #58), so southern
+        # cells (base 7-11) set bit 63 rather than reading as negative.
+        bit63 = np.uint64(1) << np.uint64(63)
+        n_north = np.sum(morton < bit63)
+        n_south = np.sum(morton >= bit63)
+        print(f"\nHemisphere distribution (bit 63):")
```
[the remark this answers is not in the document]
espg: 🤖 *from Claude*

Fixed in 26ceb33 — dropped the unused `f` prefix (`print("\nHemisphere distribution (bit 63):")`).

---

espg:
🤖 *from Claude*

Phase 5 (native `uint64` morton dtype) is pushed: `cbfa4e3` (the conversion) + `26ceb33` (ruff + self-review fixes).

**ruff bot:** both F541 findings on `test_polygon_regression.py` (the `print(f"...")` lines with no placeholders, at the new "Hemisphere distribution" line and the pre-existing "Sample morton indices" line) are fixed in `26ceb33` by dropping the unused `f` prefix. (The inline reply to the second thread 422'd — its line had shifted after the push — hence this thread-level note.)

**self-review (fresh-context, comment-only):** no blockers/majors. It confirmed the flip is consistent and complete, and that it actually *removes* a latent bug — directly constructing a southern word (`> i64::MAX`) as a Python int previously hit `OverflowError` on the old `np.int64()` path and is now lossless on `np.uint64`. Folded the one actionable nit into `26ceb33`: reworded a stale "base cells 8-11 set the i64 sign bit" comment in `test_morton_index.py` to the correct bit-63 / **7-11** boundary (prefix 8-12) with unsigned-`u64` phrasing.

Local green on `26ceb33`: `cargo test --lib` 157 passed, `cargo fmt --check` clean, `cargo clippy --all-targets` no new warnings (36 == baseline), `flake8 mortie --select=E9,F63,F7,F82` clean, `pytest -q` 388 passed / 8 skipped. Golden-fixture regen changed **only** the `word` field (450/1080 records — base cells 7-11 — flip negative→positive `u64`); `decimal_repr`/`nested`/`depth`/`base_cell`/`order`/`tuples` unchanged. New `moc.rs` test asserts a north+south cover returns in ascending unsigned/Z-order.

Leaving the PR in **draft** for review.

---

espg:
🤖 *from Claude*

**Design record: morton representation — `uint64` storage + signed decimal repr (Refs #48).** Capturing the rationale here so it's documented before it freezes at 1.0, since both choices are part of the 1.x contract.

There are **three distinct orderings** a morton index participates in, and they're served by two different representations:

1. **`uint64` numeric sort of the packed word → Z-order.** The kernel layout (`[4-bit prefix | 54-bit body | 6-bit suffix]`, prefix = `base+1`) is built so a raw *unsigned* sort is the Z-order curve — `decimal_morton.rs` calls this "the property #35 calls paramount."
2. **`int64` numeric sort → broken.** Base cells 7-11 set prefix bit 63, so reinterpreted as signed they go negative and sort to the front. This is why the public morton dtype is moving to **`uint64`** (this PR's in-flight commit): it removes the negative raw values and makes ordering (1) hold for free — which also means `moc.rs`'s `to_order`/`normalize` sorts need no signed-key special-casing.
3. **Lexicographic *string* sort of the decimal repr → S3/Hive partition keys.** This is independent of the numeric sort (string order ≠ numeric order) and is the reason the repr is **functional, not cosmetic**.

### Why the decimal repr stays *signed* (north `1`–`6`, south `-1`–`-6`)
The repr is single-character tokens throughout: an optional `-`, a lead digit `1`–`6`, then one digit `1`–`4` per order. That single-char property is load-bearing for ordering (3) and for parseability:

- **Lexicographic == Z-order within a hemisphere**, with the **prefix-scan property** (a cell's string is a prefix of its descendants' strings: `"-1"` → `"-11"`,`"-12"`,…), so S3 prefix listing / Hive partition pruning group a cell with its subtree. `"-"` (`0x2D`) sorts below digits, so the southern block sorts as one contiguous group ahead of the northern block.
- **Unambiguous round-trip parse** from the string.

A fully *unsigned-direct* scheme (encode the base cell as `1`–`12`) breaks all of this, because there is no single decimal digit for 12 cells — bases 9-11 need two-digit leads (`"10"`/`"11"`/`"12"`). That makes the repr ambiguous: **`"11"` could be base 11, or base 1 + child `1`; `"111"` could be base 11 + `1`, or base 1 + `11`** — unrecoverable without a delimiter or fixed-width padding. Lexicographically it also interleaves base-9's `"10…"` into base-0's subtree, destroying both the Z-order and the prefix grouping for string keys.

The `-` resolves this by splitting the 12 base cells into **two groups of 6**, each a single digit `1`–`6`, with the sign as the group selector — effectively a 7th bit of base-cell identity. So it simultaneously buys single-char tokens (lexicographic Z-order + prefix scans for S3/Hive), disambiguation, and round-trip parseability. The only fully-unsigned alternatives that preserve string-sortability are a single-char *non-decimal* lead (hex/letter for bases 7-12) or fixed-width zero-padding — both uglier for humans than the sign.

### Net decision
- **Storage / in-memory arrays → `uint64`** (no negative raw values; free Z-order numeric sort).
- **Decimal repr → keep the signed single-digit scheme** (deliberate, for S3/Hive lexicographic ordering + disambiguation — not legacy cruft).

These are complementary, each correct in its domain. The golden fixtures reflect this: the `uint64` switch flips only the `word` field for base cells 7-11 (negative → positive); `decimal_repr` stays signed and unchanged.

---

PR #72, opened 2026-06-26
moc_min / common_ancestor: deepest-common-ancestor reduction

### LINKED ISSUE #61: v1.0: mixed-order MOC semantics

espg:
Refs #61.

Implements the new `min` / deepest-common-ancestor reduction requested on #61 ([comment](https://github.com/espg/mortie/issues/61#issuecomment-4804033539)), per the [approved plan](https://github.com/espg/mortie/issues/61#issuecomment-4804156053) and @espg's [answers to the open questions](https://github.com/espg/mortie/issues/61#issuecomment-4804257689).

The other parts of #61 (the mixed-order MOC *semantics* decision and the depth-ceiling lift 18→29) were already discharged — the ceiling landed in #59/#70 and the storage/normalize/set-op-composition semantics were settled in the [plan comment](https://github.com/espg/mortie/issues/61#issuecomment-4804156053). This PR is the last open *code* deliverable on the issue. Left as `Refs` (not `Closes`) so the close stays @espg's call.

## What it does
Adds a reduction over an array of packed morton words that returns the **highest-order common parent** (deepest common ancestor) of the whole input, and **raises** when the words span more than one base cell (no common ancestor above the root).

The op is cheap because ancestry is already a prefix relation in the packed word: the deepest common ancestor is the longest shared path prefix (after the common base cell), capped at each word's own order. Decisions captured from the thread:

- **Name:** public Python `common_ancestor`, with `moc_min` kept as the MOC-family **alias** (`moc_min = common_ancestor`).
- **Return:** a single packed morton index — scalar `numpy.uint64`, not a length-1 array.
- **Single element** returns itself; **empty** raises `ValueError`; **irreducible** (mixed base cell) raises `ValueError`.
- **Mixed orders accepted** — each member is capped at its own order when computing the shared prefix.
- Documented as the **reduce-by-common-coarsening** sibling of `clip2order` / the deferred `//` operator.

## Approach
- `decimal_morton::common_ancestor(words) -> Result<u64, CommonAncestorError>` (`src_rust/src/decimal_morton.rs`): verifies one shared base cell, then lowers the candidate order `k` (starting at the shallowest word's order) until every word shares the order-`k` ancestor of `words[0]`, and returns `coarsen(words[0], k)`. Termination is guaranteed — at `k == 0` every word reduces to the (already-checked-equal) base cell. Reuses the existing `base_cell_of` / `to_nested` / `coarsen` kernel primitives; no new bit-twiddling. Built directly on the kernel (not via a `moc.rs` pass-through) since it operates on the packed-word path layout, which is `decimal_morton`'s domain.
- `rust_moc_min` PyO3 binding (`src_rust/src/lib.rs`): takes a `u64` array, returns a scalar `u64`, maps each `CommonAncestorError` variant to a `ValueError` (mirrors `rust_moc_xor` registration).
- Python `common_ancestor` + `moc_min` alias (`mortie/coverage.py`), exported from `mortie/__init__.py`.

## Phases
- [x] **Phase 1** — Rust kernel `common_ancestor` + `CommonAncestorError`, `rust_moc_min` binding + registration, and Rust unit tests (single/identical, four-children→parent, full-base-cell→order-0, mixed-order, order-29→28 suffix-tail, cross-base-cell raise, empty/invalid raise, plus a 300-case randomized containment property).
- [x] **Phase 2** — Python `common_ancestor` / `moc_min` wrapper + exports, and Python tests (`TestCommonAncestor`).
- [x] **Phase 3 (self-review fold)** — fixed a reversed `clip2order` call in the docstring example, added an underflow-safety note on the shift, documented the area/point kind behavior, and added tests pinning it (Rust: lone point returns itself, point batch → area cell; Python: order-0 input, order-29 single/batch).

## How it was tested
- `cargo test` — 173 passing (12 `common_ancestor` cases incl. the randomized containment property and the point-kind cases). `cargo fmt` clean.
- `cargo clippy` — no new findings beyond the pre-existing pyo3-macro `useless_conversion` warnings that every `#[pyfunction]` carries (deferred to the wholesale clippy sweep tracked on #48); `rust_moc_min` is consistent with its `moc_or`/`and`/`minus`/`xor` siblings.
- `maturin develop --release`, then `flake8 mortie --select=E9,F63,F7,F82` clean and the `TestCommonAncestor` suite + the `common_ancestor` doctest green. (CI runs the full `pytest` matrix on 3.10/3.11/3.12.)

## Questions for review
1. **Area/point kind of the result.** A genuine common ancestor is an enclosing region, so any real coarsening yields a `Kind::Area` cell. The one ambiguous case is when the input collapses to a *single* order-29 cell (no coarsening): the current behavior returns `words[0]` verbatim, so a lone `Kind::Point` returns itself (honoring "single element returns itself"), but a `{point, area}` pair at the same order-29 cell takes the first element's kind. The zagg consumer ([#87 context](https://github.com/espg/mortie/issues/61#issuecomment-4804175424)) casts points, so if you'd rather `common_ancestor` always canonicalize to the area form (dropping the point flag even for a single point), I can make that change — flagging rather than guessing, since it trades off against the "returns itself" contract.
2. **Close #61 on merge?** This delivers the issue's remaining code item; the semantics/ceiling parts already landed. Happy to switch `Refs` → `Closes`, or leave #61 open for the broader 1.x freeze statement.

### DIFF
--- a/mortie/coverage.py
+++ b/mortie/coverage.py
@@ -484,3 +484,55 @@ def moc_not(cover, domain=None):
         )
 
     return moc_minus(domain, cover)
+
+
+def common_ancestor(morton):
+    """Deepest common ancestor (highest-order common parent) of a morton set.
+
+    The array-reduction sibling of :func:`clip2order` (coarsen): where coarsening
+    lowers each word to a *caller-given* order, ``common_ancestor`` *discovers*
+    the deepest order at which the whole input collapses to a single enclosing
+    cell, and returns that one cell.  Because a packed morton word self-encodes
+    its order and ancestry, this is the longest shared path prefix after the
+    common base cell, capped at each word's own order — so mixed-order input is
+    fine (each word is capped at its own order).
+
+    Parameters
+    ----------
+    morton : array_like
+        Morton indices (mixed order allowed).  A single index returns itself.
+
+    Returns
+    -------
+    numpy.uint64
+        The packed morton index of the deepest cell that contains every input.
+
+    Raises
+    ------
+    ValueError
+        If ``morton`` is empty, contains an empty/invalid word, or spans more
+        than one HEALPix base cell — there is then no common ancestor above the
+        (non-existent) whole-sphere root.
+
+    See Also
+    --------
+    clip2order : coarsen each word to a fixed order (the elementwise form;
+        ``common_ancestor`` is its reduce-by-common-coarsening reduction).
+
+    Examples
+    --------
+    The four order-4 children of an order-3 cell reduce to that parent:
+
+    >>> import mortie, numpy as np
+    >>> parent = mortie.norm2mort(7, 0, 3)             # one order-3 cell in base 0
+    >>> kids = mortie.clip2order(np.full(4, parent), 4)  # doctest: +SKIP
+    >>> mortie.common_ancestor(kids) == parent           # doctest: +SKIP
+    True
+    """
+
+    morton = np.asarray(morton, dtype=np.uint64).ravel()
+    return np.uint64(_rustie.rust_moc_min(morton))
+
+
+# ``moc_min`` is the MOC set-family alias for :func:`common_ancestor` (issue #61).
+moc_min = common_ancestor
--- a/mortie/coverage.py
+++ b/mortie/coverage.py
@@ -506,10 +506,11 @@ def common_ancestor(morton):
     -------
     numpy.uint64
         The packed morton index of the deepest cell that contains every input.
-        Any genuine coarsening yields an **area** cell; the area/point kind is
-        preserved only when the whole input collapses to a single order-29 cell
-        (then the first word is returned unchanged, so a lone area or point
-        returns itself).
+        A batch (more than one input) always yields an **area** cell — even when
+        the inputs collapse to a single order-29 cell, since the shared cell is
+        an enclosing area, not any one input point.  Only a single-element input
+        is returned unchanged (its area/point kind preserved), so a lone area or
+        point returns itself.
 
     Raises
     ------
--- a/mortie/coverage.py
+++ b/mortie/coverage.py
@@ -506,6 +506,10 @@ def common_ancestor(morton):
     -------
     numpy.uint64
         The packed morton index of the deepest cell that contains every input.
+        Any genuine coarsening yields an **area** cell; the area/point kind is
+        preserved only when the whole input collapses to a single order-29 cell
+        (then the first word is returned unchanged, so a lone area or point
+        returns itself).
 
     Raises
     ------
@@ -521,12 +525,12 @@ def common_ancestor(morton):
 
     Examples
     --------
-    The four order-4 children of an order-3 cell reduce to that parent:
+    The four order-5 children of an order-4 cell reduce to that parent:
 
     >>> import mortie, numpy as np
-    >>> parent = mortie.norm2mort(7, 0, 3)             # one order-3 cell in base 0
-    >>> kids = mortie.clip2order(np.full(4, parent), 4)  # doctest: +SKIP
-    >>> mortie.common_ancestor(kids) == parent           # doctest: +SKIP
+    >>> parent = mortie.norm2mort(11, 0, 4)              # one order-4 cell in base 0
+    >>> kids = mortie.norm2mort([11 * 4 + s for s in range(4)], [0] * 4, 5)
+    >>> int(mortie.common_ancestor(kids)) == int(parent)
     True
     """
 
--- a/mortie/tests/test_coverage.py
+++ b/mortie/tests/test_coverage.py
@@ -1086,6 +1086,18 @@ class TestCommonAncestor:
         got = mortie.common_ancestor(cell)
         assert int(got) == int(cell[0])
 
+    def test_order_29_identical_batch_returns_order_29_cell(self):
+        # Two *identical* order-29 cells (no coarsening) still reduce to the
+        # order-29 cell that encloses them. Mirrors the Rust point-kind case:
+        # a batch always yields the enclosing area, never a passed-through point.
+        # (geo2mort emits area words, so the value is the same cell.)
+        cell = int(mortie.geo2mort(45.0, -100.0, 29)[0])
+        pts = np.array([cell, cell], dtype=np.uint64)
+        anc = mortie.common_ancestor(pts)
+        assert int(anc) == cell
+        _, anc_order = mortie.mort2healpix(anc)
+        assert int(anc_order) == 29
+
     def test_order_29_batch_encloses_to_common_cell(self):
         # A batch of nearby order-29 cells reduces to their common enclosing
         # cell. Check containment by *coarsening* each input to the ancestor's
--- a/mortie/tests/test_coverage.py
+++ b/mortie/tests/test_coverage.py
@@ -985,3 +985,91 @@ class TestMOCNot:
             warnings.simplefilter("error")
             got = mortie.moc_not(cover, domain=empty)
         assert len(got) == 0
+
+
+class TestCommonAncestor:
+    """Deepest common ancestor / moc_min reduction (issue #61).
+
+    `common_ancestor(words)` returns the single highest-order cell that contains
+    every input word; `moc_min` is its alias.  Containment is checked by
+    densifying the ancestor and the inputs to a common deep order with
+    `moc_to_order` and asserting the ancestor's leaves are a superset.
+    """
+
+    def test_moc_min_is_common_ancestor_alias(self):
+        assert mortie.moc_min is mortie.common_ancestor
+
+    def test_single_returns_itself(self):
+        cell = mortie.norm2mort(7, 3, 5)  # one order-5 cell in base 3
+        assert int(mortie.common_ancestor(np.atleast_1d(cell))) == int(cell)
+
+    def test_identical_returns_the_cell(self):
+        cell = mortie.norm2mort(42, 9, 6)
+        words = np.full(4, cell, dtype=np.uint64)
+        assert int(mortie.common_ancestor(words)) == int(cell)
+
+    def test_four_children_reduce_to_parent(self):
+        # The four order-5 children of an order-4 cell reduce to that parent.
+        parent = mortie.norm2mort(11, 0, 4)  # within-base norm 11 @ order 4
+        # The four order-5 children share parent's norm prefix (norm*4 + s).
+        kids = mortie.norm2mort([11 * 4 + s for s in range(4)], [0] * 4, 5)
+        assert int(mortie.common_ancestor(kids)) == int(parent)
+
+    def test_diverging_cells_reduce_to_base_cell(self):
+        # Two cells diverging at order 1 share only the base cell -> order 0.
+        base = 7
+        a = mortie.norm2mort(0, base, 3)  # nested 0 @ order 3
+        b = mortie.norm2mort(1 << 4, base, 3)  # different first tuple @ order 3
+        got = mortie.common_ancestor(mortie.norm2mort([0, 1 << 4], [base, base], 3))
+        assert int(got) == int(mortie.norm2mort(0, base, 0))
+        # sanity: a and b really are in the same base cell.
+        assert (int(a) >> 60) == (int(b) >> 60)
+
+    def test_mixed_order_input(self):
+        # A deep order-8 cell and a shallow order-5 cell sharing a common prefix
+        # reduce to the prefix cell; the ancestor contains both.
+        base = 2
+        deep = mortie.norm2mort(0b00_01_10_11_00_01_10_11, base, 9)
+        shallow = mortie.norm2mort(0b00_01_10, base, 3)  # shares first 3 tuples
+        anc = mortie.common_ancestor(
+            np.concatenate([np.atleast_1d(deep), np.atleast_1d(shallow)])
+        )
+        anc_leaves = set(int(x) for x in mortie.moc_to_order(np.atleast_1d(anc), 9))
+        for w in (deep, shallow):
+            leaves = set(int(x) for x in mortie.moc_to_order(np.atleast_1d(w), 9))
+            assert leaves.issubset(anc_leaves), "ancestor must contain every input"
+
+    def test_order_29_reduces_to_28(self):
+        # Two order-29 cells differing only at order 29 reduce to their order-28
+        # parent (exercises the suffix-tail path through the binding).
+        base = 9
+        a = mortie.norm2mort(0, base, 29)
+        b = mortie.norm2mort(1, base, 29)  # differs only in the lowest tuple
+        got = mortie.common_ancestor(
+            np.concatenate([np.atleast_1d(a), np.atleast_1d(b)])
+        )
+        _, order = mortie.mort2healpix(got)  # scalar form: (norm, order)
+        assert int(order) == 28
+        assert int(got) == int(mortie.norm2mort(0, base, 28))
+
+    def test_cross_base_cell_raises(self):
+        a = mortie.norm2mort(0, 2, 4)
+        b = mortie.norm2mort(0, 5, 4)  # different base cell
+        with pytest.raises(ValueError, match="base cell"):
+            mortie.common_ancestor(
+                np.concatenate([np.atleast_1d(a), np.atleast_1d(b)])
+            )
+
+    def test_empty_raises(self):
+        with pytest.raises(ValueError, match="empty"):
+            mortie.common_ancestor(np.array([], dtype=np.uint64))
+
+    def test_invalid_word_raises(self):
+        with pytest.raises(ValueError, match="invalid"):
+            mortie.common_ancestor(np.array([0], dtype=np.uint64))
+
+    def test_returns_scalar_uint64(self):
+        cell = mortie.norm2mort(7, 3, 5)
+        got = mortie.common_ancestor(np.atleast_1d(cell))
+        assert isinstance(got, np.uint64)
+        assert got.shape == ()
--- a/mortie/tests/test_coverage.py
+++ b/mortie/tests/test_coverage.py
@@ -1088,7 +1088,10 @@ class TestCommonAncestor:
 
     def test_order_29_batch_encloses_to_common_cell(self):
         # A batch of nearby order-29 cells reduces to their common enclosing
-        # cell, which must contain every input when densified back to order 29.
+        # cell. Check containment by *coarsening* each input to the ancestor's
+        # order (every input's ancestor at that order must equal it) rather than
+        # densifying the ancestor to order 29 — a coarse ancestor would expand to
+        # ~4**(29-order) leaves and exhaust memory.
         lats = [45.0, 45.0001, 45.0002]
         lons = [-100.0, -100.0001, -100.0002]
         pts = np.array(
@@ -1096,7 +1099,6 @@ class TestCommonAncestor:
             dtype=np.uint64,
         )
         anc = mortie.common_ancestor(pts)
-        anc_leaves = set(int(x) for x in mortie.moc_to_order(np.atleast_1d(anc), 29))
-        for p in pts:
-            leaves = set(int(x) for x in mortie.moc_to_order(np.atleast_1d(p), 29))
-            assert leaves.issubset(anc_leaves)
+        _, anc_order = mortie.mort2healpix(anc)  # scalar form: (norm, order)
+        coarsened = mortie.clip2order(int(anc_order), pts)
+        assert all(int(c) == int(anc) for c in np.atleast_1d(coarsened))
--- a/mortie/tests/test_coverage.py
+++ b/mortie/tests/test_coverage.py
@@ -1073,3 +1073,30 @@ class TestCommonAncestor:
         got = mortie.common_ancestor(np.atleast_1d(cell))
         assert isinstance(got, np.uint64)
         assert got.shape == ()
+
+    def test_order_zero_input_returns_base_cell(self):
+        # Order-0 base cells in the same base trivially reduce to that base cell.
+        base0 = mortie.norm2mort(0, 4, 0)
+        got = mortie.common_ancestor(np.full(3, base0, dtype=np.uint64))
+        assert int(got) == int(base0)
+
+    def test_order_29_single_returns_itself(self):
+        # A lone order-29 cell (a point cast to max resolution) returns itself.
+        cell = np.atleast_1d(mortie.geo2mort(45.0, -100.0, 29))
+        got = mortie.common_ancestor(cell)
+        assert int(got) == int(cell[0])
+
+    def test_order_29_batch_encloses_to_common_cell(self):
+        # A batch of nearby order-29 cells reduces to their common enclosing
+        # cell, which must contain every input when densified back to order 29.
+        lats = [45.0, 45.0001, 45.0002]
+        lons = [-100.0, -100.0001, -100.0002]
+        pts = np.array(
+            [int(mortie.geo2mort(la, lo, 29)[0]) for la, lo in zip(lats, lons)],
+            dtype=np.uint64,
+        )
+        anc = mortie.common_ancestor(pts)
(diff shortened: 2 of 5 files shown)


mortie/coverage.py
```diff
@@ -484,3 +484,55 @@ def moc_not(cover, domain=None):
...
+    clip2order : coarsen each word to a fixed order (the elementwise form;
+        ``common_ancestor`` is its reduce-by-common-coarsening reduction).
+
+    Examples
+    --------
+    The four order-4 children of an order-3 cell reduce to that parent:
+
+    >>> import mortie, numpy as np
+    >>> parent = mortie.norm2mort(7, 0, 3)             # one order-3 cell in base 0
+    >>> kids = mortie.clip2order(np.full(4, parent), 4)  # doctest: +SKIP
```
espg: 🤖 *from Claude (review)*

Docstring example has `clip2order`'s arguments reversed. The signature is `clip2order(clip_order, midx=None, ...)` — order is the **first** positional arg (`tools.py:556`), so this should read `mortie.clip2order(4, np.full(4, parent))`, not `mortie.clip2order(np.full(4, parent), 4)`. As written it would raise if it ran (it's `+SKIP`, so CI won't catch it). Since this docstring's whole framing is "`common_ancestor` is the reduction sibling of `clip2order`," an example that mis-calls `clip2order` undercuts the point. Consider fixing the call and dropping `+SKIP` so the doctest actually pins the parent↔children round-trip.

---

src_rust/src/decimal_morton.rs
```diff
@@ -365,6 +365,85 @@ pub fn coarsen(word: u64, k: u8) -> Option<u64> {
...
+            .iter()
+            .all(|&(o, n)| n >> (2 * (o - k) as u32) == target)
+        {
+            break;
+        }
+        k -= 1;
+    }
+    // `coarsen(first, k)` with `k <= order0` is the canonical order-`k` ancestor
+    // word; at `k == order0` (single / identical input) it returns `first` as-is.
+    Ok(coarsen(first, k).expect("first decoded above"))
```
espg: 🤖 *from Claude (review)*

Result is **input-order-dependent** when the reduction bottoms out at the deepest order and the inputs are a point/area mix at the same order-29 cell. `to_nested` flattens a `Kind::Point` to the same `(29, nested)` as the equivalent order-29 area cell (`:521`, "returns its order-29 nested cell just like an area cell"), so a point and the area cell of the same cell never lower `k` below 29. Then `coarsen(first, 29)` returns `first` **unchanged** (`coarsen` short-circuits `k >= native`, `:338`). So `common_ancestor(&[point, area])` returns the point while `common_ancestor(&[area, point])` returns the area cell — same logical cell, different `Kind`/suffix bits, decided purely by array order. Worth either (a) documenting that the kind of the returned word follows `words[0]` in the collapse case, or (b) canonicalizing to `Kind::Area` (an ancestor is an enclosing area, never a point). Note `coarsen(first, k)` for `k < 29` already yields an area cell, so only the all-collapse-at-29 case is affected. There's no test pinning this either way — see the test-coverage comment below.

---

src_rust/src/decimal_morton.rs
```diff
@@ -365,6 +365,85 @@ pub fn coarsen(word: u64, k: u8) -> Option<u64> {
...
+        }
+        let (o, n) = to_nested(w).expect("base_cell_of accepted this word");
+        k = k.min(o);
+        others.push((o, n));
+    }
+
+    // Lower `k` until every word shares `first`'s order-`k` ancestor.  This
+    // terminates at `k == 0` at the latest: there every word reduces to its base
+    // cell, which we have already proved equal across the input.
+    loop {
```
espg, line 441: 🤖 *from Claude (review)*

Termination/underflow reasoning checks out, but worth a one-line note for the reviewer: `k` is initialized to `min` over **all** orders incl. `order0` (`:418`/`:425`), so `order0 - k >= 0` and every `o - k >= 0`; the loop only ever decrements (never at `k == 0`, since all base cells are proven equal before the loop), so the `u8` shift exponents never underflow. The logic is correct — the comment at `:429-431` covers the `k == 0` floor but not why the per-word `o - k` shift is always in range; a half-line there would save the next reader the same derivation. (Non-blocking.)

---

mortie/tests/test_coverage.py
```diff
@@ -985,3 +985,91 @@ def test_empty_domain_returns_empty_without_warning(self):
...
+
+    def test_invalid_word_raises(self):
+        with pytest.raises(ValueError, match="invalid"):
+            mortie.common_ancestor(np.array([0], dtype=np.uint64))
+
+    def test_returns_scalar_uint64(self):
+        cell = mortie.norm2mort(7, 3, 5)
+        got = mortie.common_ancestor(np.atleast_1d(cell))
+        assert isinstance(got, np.uint64)
+        assert got.shape == ()
```
espg, line 1075: 🤖 *from Claude (review)*

Test gaps against the @espg-pinned behavior (issue #61):

1. **No point-kind input at the Python layer.** The Rust suite covers a single point (`common_ancestor_single_returns_itself`), but nothing covers a `Kind::Point` flowing through the binding, and nothing covers the **point+area-at-same-cell** mix that exposes the input-order dependence flagged on `decimal_morton.rs`. If the intended contract is "ancestor is always an area cell," add a test that pins it (e.g. an order-29 point and the area cell of the same cell, asserting a stable `kind`/value regardless of order). If order-dependence is acceptable, a test documenting it would still be worth having.

2. **No order-0 (base-cell) input.** Rust has `common_ancestor_full_base_cell_reduces_to_order_zero`, but the Python suite never feeds an order-0 word in or asserts the order-0 result decodes cleanly through `mort2healpix`.

3. **Single-base-cell-only 2-element collapse** isn't pinned at the Python layer beyond `test_identical_returns_the_cell` (which uses 4 identical words). A 2-element same-cell case and a single-point case would round out the `single → itself` / `identical → the cell` contract here rather than only in Rust.

Not blockers, but (1) is the meaty one given the determinism question.

---

mortie/coverage.py
```diff
@@ -484,3 +484,55 @@ def moc_not(cover, domain=None):
...
+    >>> mortie.common_ancestor(kids) == parent           # doctest: +SKIP
+    True
+    """
+
+    morton = np.asarray(morton, dtype=np.uint64).ravel()
+    return np.uint64(_rustie.rust_moc_min(morton))
```
espg, line 539: 🤖 *from Claude (review)*

`np.asarray(morton, dtype=np.uint64).ravel()` will silently coerce non-integer / out-of-range input (e.g. a float array, or negative values via wraparound) rather than rejecting it — the Rust side then sees garbage words and reports them as `Invalid`, which surfaces as a confusing `ValueError`. The neighboring `moc_*` wrappers in this file use the same `np.asarray(..., dtype=np.uint64)` idiom, so this matches surrounding code and is fine to leave — just flagging that a non-uint64 input doesn't fail fast. No change needed if you want to stay consistent with the family.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-eyes adversarial review of the `common_ancestor` / `moc_min` primitive (issue #61). Overall the core is **correct and well-matched to the surrounding kernel** — I verified the things most likely to be wrong and they hold up:

- **Reduction loop**: terminates (floors at `k == 0` because all base cells are proven equal before the loop), and the shift exponents (`order0 - k`, per-word `o - k`) never underflow because `k` is seeded from the `min` order across *all* words including `first`. The `coarsen(first, k)` constructor is sound: the loop proves every word shares `first`'s order-`k` prefix, so coarsening any member yields the same canonical word, and at `k == order0` it returns `first` unchanged (single/identical → itself). ✔
- **Error mapping**: empty → `Empty`, undecodable word → `Invalid`, cross-base → `MixedBaseCell`, each mapped to `ValueError` with a matchable message — matches the @espg-pinned contract (single→itself, empty/irreducible raise, mixed orders accepted). ✔
- **API surface**: `moc_min = common_ancestor` alias is sound for `__all__`/imports; scalar `np.uint64` return with `shape == ()` is pinned by `test_returns_scalar_uint64`. ✔
- Terseness, naming, docstrings, no new deps, line lengths all in line with the conventions.

Findings posted inline, none are merge-blockers:
1. **`coverage.py:528`** — docstring example calls `clip2order` with args reversed (order is the first positional). `+SKIP` hides it from CI but it's a wrong example in the very docstring that frames this as `clip2order`'s reduction sibling.
2. **`decimal_morton.rs:444`** — result is input-order-dependent for a point/area mix at the same order-29 cell (`to_nested` flattens `Kind::Point` to the area cell, so `k` stays at 29 and `coarsen` returns `first` verbatim → returned `kind` follows `words[0]`). Document or canonicalize to `Kind::Area`.
3. **`test_coverage.py`** — no point-kind input, no point+area-same-cell case (the determinism question above), no order-0 input at the Python layer; (1) is the meaty gap.
4. **`decimal_morton.rs:432`** — minor: a half-line on why the per-word shift never underflows would help the next reader.
5. **`coverage.py:534`** — minor: non-uint64 input coerces silently rather than failing fast; matches the `moc_*` family, fine to leave.

Recommendation: fold in (1) and decide (2) (with a test per (3)); the rest are optional.

---

espg:
🤖 *from Claude*

Folded the fresh-context self-review's diff-scoped findings into the `address self-review on #61` commit:

- **Docstring example fixed** — the `clip2order` call in `common_ancestor`'s docstring had its arguments reversed (signature is `clip2order(clip_order, midx)`); replaced it with a correct, doctest-checked `norm2mort`-based example.
- **Area/point determinism documented + pinned** — the reviewer flagged that a `{point, area}` pair at the same order-29 cell is `words[0]`-dependent. Documented the kind behavior in both the Rust doc comment and the Python docstring, and added tests: a lone point returns itself, and a batch of points reduces to their enclosing **area** cell (Rust); plus order-0 and order-29 single/batch cases (Python). The remaining canonicalize-to-area design choice is raised as Question (1) in the PR body for your call — I didn't guess it since it trades against the "single element returns itself" contract.
- **Underflow-safety note** added on the `o - k` shift (logic was already correct; `k` is seeded from the `min` order over all words).

The minor "silent uint64 coercion" finding I left as-is — it matches the existing `moc_*` family behavior.

All phases are complete and CI is green; this is now `waiting` on your review and on Question (1) (result kind) / Question (2) (whether to `Closes #61`).


---

espg:
🤖 *from Claude*

Correction to my earlier "CI is green" note — the first phase-3 push actually **failed** the test matrix, and I've since fixed it:

- **Root cause:** the new `test_order_29_batch_encloses_to_common_cell` checked containment by densifying the common ancestor to order 29 with `moc_to_order`. When the batch's ancestor is a *coarse* cell, that expands to ~`4**(29-order)` leaves and OOM-killed the runner (`The runner has received a shutdown signal` mid-test). Exactly the flat-cover memory hazard the codebase already warns about.
- **Fix** (`fix order-29 batch test densification on #61`): check containment the cheap way instead — coarsen each input to the ancestor's order with `clip2order` and assert they all equal the ancestor. No densification, no expansion.

The test matrix is now green on **3.10 / 3.11 / 3.12**, alongside ruff and the full wheel-build matrix. The one remaining red check, **CodSpeed Performance Analysis**, completes in the same second it starts and has no GitHub Actions log (it's the external app.codspeed.io integration) — an instant baseline/token failure on the branch, independent of this diff, not a real perf-run result. Flagging rather than touching it, since CI config is out of scope here.

Still `waiting` on your review and on Question (1) (result kind) / Question (2) (`Closes #61`?).


---

espg:
> Area/point kind of the result. A genuine common ancestor is an enclosing region, so any real coarsening yields a Kind::Area cell. The one ambiguous case is when the input collapses to a single order-29 cell (no coarsening): the current behavior returns words[0] verbatim, so a lone Kind::Point returns itself (honoring "single element returns itself"), but a {point, area} pair at the same order-29 cell takes the first element's kind. The zagg consumer (https://github.com/espg/mortie/issues/61#issuecomment-4804175424) casts points, so if you'd rather common_ancestor always canonicalize to the area form (dropping the point flag even for a single point), I can make that change — flagging rather than guessing, since it trades off against the "returns itself" contract.

This will basically never happen-- we don't have an input resolution high enough to trigger this, nor do we have mechanics for storing orders greater than order 29 (although we can reason about them). Definitionally, if there's two order 29 points that are input into `moc_min` , and they are encoded as 'points' rather than area, the correct behavior would be to emit an order 29 'area', since both 'points' exist in the same 'area'.

Don't close #61 ; update it (convert the bullet to checkboxes in the summary, and check what's been completed). There are still non-code items left in that issue (I believe).

src_rust/src/decimal_morton.rs
```diff
@@ -365,6 +365,100 @@ pub fn coarsen(word: u64, k: u8) -> Option<u64> {
...
+    }
+    // Single-element input returns `first` verbatim, preserving its kind. A batch
+    // is always an enclosing *area*: `from_nested` packs the order-`k` ancestor
+    // (the shared nested prefix) into an `Kind::Area` word, even when `k == order0`
+    // (all inputs the same order-29 cell), where `coarsen` would have kept a point.
+    if rest.is_empty() {
+        Ok(coarsen(first, k).expect("first decoded above"))
+    } else {
+        Ok(from_nested(nested0 >> (2 * (order0 - k) as u32), k))
+    }
```
espg, line 459: 🤖 *from Claude (review)*

Verified this batch path is correct on both **value** and **kind**, no off-by-one or base-offset mismatch:

- `to_nested` returns `nested = base * 4^order + within` (full nested index, base offset included). `nested0 >> (2*(order0 - k))` divides by `4^(order0-k)`, yielding `base*4^k + (within >> 2*(order0-k))` — the full order-`k` nested index, base offset intact. `from_nested` consumes exactly that (`base = nested >> 2*depth`), so the conventions line up. ✅
- The shift is always non-negative: `k` is seeded from `order0` and only decremented, and this branch uses `order0 - k` with `order0` = `first`'s order. ✅
- Value equivalence with the old `coarsen` path: the loop computes `target = nested0 >> (2*(order0-k))` and breaks when every word matches it, so the value passed to `from_nested` is the proven shared prefix. `from_nested` rebuilds the same path tuples `coarsen` would have kept, differing only in always emitting `Kind::Area`. ✅

One readability nit (non-blocking): the loop already computes this exact `target` expression each iteration, and on `break` it equals `nested0 >> (2*(order0-k))`. Recomputing it here is fine, but lifting the post-loop `target` into a named binding (or returning it from the loop via `let ancestor_nested = loop { ... break X };`) would remove the duplicated shift expression and make the "this is the shared prefix" intent self-evident. Purely cosmetic.

---

src_rust/src/decimal_morton.rs
```diff
@@ -1223,6 +1317,183 @@ mod tests {
...
+        let same = common_ancestor(&[p_a, p_a]).expect("same base => Ok");
+        assert_eq!(
+            kind_of(same),
+            Kind::Area,
+            "identical points => area, not point"
+        );
+        assert_eq!(order_of(same), 29);
+        assert_eq!(same, encode(base, &t, 29));
+    }
+
```
espg, line 1430: 🤖 *from Claude (review)*

The new Rust test is well-targeted — it's the **only** test that exercises the actually-changed branch (the `k == order0` batch path, which is where new behavior diverges from the old `coarsen` pass-through). It asserts kind (`Kind::Area`), order (29), *and* value (`encode(base, &t, 29)`), which is exactly the right coverage. ✅

The value assert is sound for a subtle reason worth recording: `p_a = encode_point(base, &t)` is a `Kind::Point`, but `to_nested` decodes a point's order-28/29 tail tuples identically to an area's (per its own docstring), so `from_nested(to_nested(p_a).1, 29) == encode(base, &t, 29)`. So the area word reconstructed from a point input lands on the same path. Good.

Nice-to-have (non-blocking): the coarsening cases (`point_kind_behavior`'s `[p_a, p_b] -> encode(base,&t,5)`, `full_base_cell_reduces_to_order_zero`) already covered the `k < order0` branch, but every one of those used inputs where the old `coarsen` path *also* returned area — so none of them would have caught a regression in the new `from_nested` value math at a *coarsening* order. They pass, and `[p_a, p_b]` does mix point-input → area-output, so coverage is adequate; just noting the identical-batch test is doing the heavy lifting for the changed line.

---

mortie/tests/test_coverage.py
```diff
@@ -985,3 +985,132 @@ def test_empty_domain_returns_empty_without_warning(self):
...
+    def test_order_29_identical_batch_returns_order_29_cell(self):
+        # Two *identical* order-29 cells (no coarsening) still reduce to the
+        # order-29 cell that encloses them. Mirrors the Rust point-kind case:
+        # a batch always yields the enclosing area, never a passed-through point.
+        # (geo2mort emits area words, so the value is the same cell.)
+        cell = int(mortie.geo2mort(45.0, -100.0, 29)[0])
+        pts = np.array([cell, cell], dtype=np.uint64)
+        anc = mortie.common_ancestor(pts)
+        assert int(anc) == cell
+        _, anc_order = mortie.mort2healpix(anc)
```
espg, line 1098: 🤖 *from Claude (review)*

This Python test asserts the right things (value `== cell` and order `== 29`), but note it does **not** independently exercise the new behavior at the kind level the way the directive frames it. `geo2mort_scalar` itself calls `from_nested(nest, order)` (src_rust/src/geo2mort.rs:24-28), so `cell` is *already* a `Kind::Area` word — there's no point input on the Python side, so `int(anc) == cell` would have held under the *old* code too (old code returned `words[0]` verbatim, and `words[0]` here is an area). The Rust test (`common_ancestor_point_kind_behavior`) is what actually pins the point→area transition. If you want the Python layer to also pin the kind, consider asserting `mort2norm`/`kind_of`-style area-ness, or constructing a point word — otherwise this test is a value/order regression guard, not a kind guard. Non-blocking.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of commit `90f0399` ("address @espg result-kind directive on #61"), reviewed against CLAUDE.md and the issue #61 result-kind directive. **The diff is correct — I found no blocking bug.**

**Correctness (verified, not assumed):**
- Batch path `from_nested(nested0 >> (2*(order0 - k)), k)` is right on both value and kind. `to_nested` returns a full nested index *including* the base offset (`base*4^order + within`); the right-shift produces the full order-`k` nested index, which is exactly `from_nested`'s input convention (`base = nested >> 2*depth`). No off-by-one, no base-offset mismatch. Shift is always ≥ 0. (inline)
- The batch always emits `Kind::Area`, including the no-coarsening `k == order0` case — matching the directive. Single-element (`words.len() == 1`) still returns `first` verbatim via `coarsen`, preserving kind. Behavior, both docstrings (Rust + Python), and tests are now mutually consistent.

**Tests:** The new Rust test `common_ancestor_point_kind_behavior` additions are the high-value coverage — they're the only test exercising the changed `k == order0` branch, and they assert kind + order + value. Adequate.

**Non-blocking notes:**
1. The Python test guards value/order but not kind: `geo2mort` already emits an area word (it calls `from_nested` internally), so that assertion would have passed under the old code too. The point→area transition is pinned only on the Rust side. (inline)
2. `nested0 >> (2*(order0-k))` is recomputed post-loop; lifting the loop's `target` into a binding would remove the duplicated shift. Cosmetic. (inline)
3. `decimal_morton.rs` is now 1513 lines — over the ~1000-line soft limit in CLAUDE.md §4. This commit adds only ~6 net lines so it doesn't *cause* the overage, but the module was already past the threshold; flagging for awareness, not as something this PR introduced.

No undiscussed dependencies, no CI/workflow/version edits, commit message is terse and title-only per §3. Looks good to fold in.

---

espg:
🤖 *from Claude*

Folded your result-kind directive (commit `90f0399`):

- **`common_ancestor` now canonicalizes a batch to `Kind::Area`.** For `len > 1` the result is emitted via `from_nested(...)` (always area) — including the no-coarsening case where all inputs are the same order-29 cell, which previously returned `words[0]` verbatim and could carry a `Kind::Point` flag. A **single-element** input still returns itself unchanged (kind preserved), honoring "single element returns itself." This matches your call: two order-29 points → the enclosing order-29 **area**.
- Doc comments (Rust + Python) rewritten to state the rule (batch → always area; single → verbatim), and a Rust test pins it (two identical order-29 points → `Kind::Area` at order 29, value `encode(base, &t, 29)`), plus a Python batch test.
- **Q2 (don't close #61):** done — left as `Refs`, and I converted #61's Scope to checkboxes and ticked the completed items (depth ceiling, storage form, normalize/compress, set-op composition, and this `moc_min`/`common_ancestor` reduction), leaving the dtype/NEP-42 interplay and the 1.x freeze statement open.

Fresh-context self-review ran on the commit (review on `90f0399`): no correctness bug — it verified the `to_nested`→`from_nested` nested-index conventions line up (no base-offset mismatch / off-by-one) and the kind directive is satisfied. Non-blocking notes: the kind transition is only pinned on the Rust side (Python has no public way to construct a `Kind::Point`, so there's nothing kind-wise to assert from Python); a cosmetic dedup of the post-loop shift; and `decimal_morton.rs` is now ~1513 lines — **over §4's ~1000-line soft limit** (pre-existing, not caused by this 6-line commit). Flagging the module size for your call rather than splitting it unilaterally (§4 wants that discussed first) — want a follow-up to carve the kernel/MOC/bridge sections into separate modules?

Tests green: `cargo test` 173 passed, `cargo fmt`/`flake8` clean, `pytest` 430 passed / 8 skipped, coverage 93%. The one red check is **CodSpeed** (a 13.38% "regression" on `fast_norm2mort_scalar`) — unrelated to this diff (`common_ancestor` doesn't touch `fast_norm2mort`); it's the same external-app baseline/noise result flagged earlier, not a real perf change, and doesn't block a draft.

All phases complete and your open questions are answered — marking `waiting` on your review / merge.

---

PR #84, opened 2026-06-27
add split_base_cells

### LINKED ISSUE #74: moc_min followup
We probably should have a companion method, something like 'split_base_cells' that when given a MOC, will return a single array if all are in the same base cell, or return a list of arrays that contain the MOC subsets paired to there respective base cells.

...the other option would be a flag on `moc_min` that allowed splitting, so users could call it. We need the invariant reducer for downstream workflows that expect a single value return... but perhaps if we do add a flag here, the raise error when that flag isn't set could be useful, and direct users to enable `base_cell_splits=True` (or similar).

refs #72

espg:
Closes #74
Refs #72, Refs #61

## What it does

Adds `split_base_cells(words, sort=False)`, the companion primitive to `moc_min` for the cross-base-cell case `moc_min` refuses. It partitions a morton set by HEALPix base cell and returns a **dict keyed by each group's `moc_min`**, exactly per the binding spec in #74:

- **Group-by, not further reduced.** The value of each entry is that group's raw array of words (a `uint64` `np.ndarray`), preserving input order within the group.
- **Key is the group's `moc_min`.** Per @espg's [answer](https://github.com/espg/mortie/issues/74#issuecomment-4814336560): "split to list of arrays, then run `moc_min` on each of the groups; the dict return then has the `moc_min` output as the dict key." Self-describing, same 64-bit width as the data, and the base-cell id is cheap to recover from the key (`key >> 60` minus the `+1` prefix offset, or `mort2healpix` / `MortonIndexArray.base_cell`).
- **`sort` flag.** Default `sort=False` (faster — preserves input order within each group); `sort=True` sorts each group's words for a canonical MOC per base cell. The `moc_min` key is order-insensitive, so both modes produce the same keys.
- **Standalone primitive, not a flag on `moc_min`.** `moc_min` stays strict; its mixed-base-cell error message now **points at `split_base_cells`** as the remedy (`src_rust/src/decimal_morton.rs`, `CommonAncestorError::MixedBaseCell` Display).
- Edge cases: empty input → `{}`; single base cell → one-entry dict; an empty/invalid word makes that group's `moc_min` raise `ValueError`.

## Approach

Implemented as a **thin Python wrapper in `mortie/coverage.py`**, next to `common_ancestor` / `moc_min`. It composes two existing Rust primitives rather than adding a new kernel:

- `_rustie.rust_mi_base_cell_of(words)` — the existing vectorized per-element base-cell extraction (already used by `MortonIndexArray.base_cells`) — so the 4-bit-prefix layout isn't re-hardcoded in Python.
- `moc_min` per group for the dict key.

```python
bases = _rustie.rust_mi_base_cell_of(words)
out = {}
for base in dict.fromkeys(bases.tolist()):   # stable, first-seen order
    group = words[bases == base]
    if sort:
        group = np.sort(group)
    out[int(moc_min(group))] = group
```

This matches @espg's acceptance of "a thin Python wrapper that groups by base-cell prefix then calls the existing `moc_min` per group … if it matches surrounding style." The only Rust change is the one-line `moc_min` error-message pointer the spec requires. Exported from `mortie/__init__.py` (`__all__`) consistent with `moc_min`.

## Phases

- [x] **Phase 1** — `split_base_cells` wrapper + docstring in `coverage.py`, `moc_min` mixed-base-cell error message points here (Rust), export in `__init__.py`, and `TestSplitBaseCells` tests.

(Single coherent phase: a thin wrapper over existing primitives plus a one-line Rust message edit; no useful sub-phase split.)

## How it was tested

- `cargo fmt --check` clean; `cargo test --release` — **173 passed** (the message edit touches only `Display`; no test pins the old string).
- `cargo clippy --release` — no new findings (only the pre-existing pyo3-macro `useless_conversion` warnings deferred to #48).
- `maturin develop --release`, then `flake8 mortie --select=E9,F63,F7,F82` clean and `pytest -q` — **440 passed, 8 skipped**. New `TestSplitBaseCells` (10 cases) covers: empty → `{}`, single-base-cell one-entry, multi-base-cell partition, the dict-key-is-`moc_min` invariant, base-cell id recoverable from each key, `sort=False` preserves input order vs `sort=True` canonical order (same keys), values are `uint64`, invalid-word raise, and that `moc_min` on mixed base cells raises with a message naming `split_base_cells`. The `split_base_cells` doctest also passes under `--doctest-modules`.

## Questions for review

1. **Container ordering of the dict.** Entries are inserted in first-seen base-cell order (stable `dict.fromkeys`), not sorted by base id. The spec pinned the within-group order (the `sort` flag) but not the order of the dict's entries; first-seen is the cheaper default and dicts are nominally unordered. Happy to sort entries by base id if you'd prefer a deterministic key order.
2. **Implementation site.** Done as a Python wrapper composing `rust_mi_base_cell_of` + `moc_min` (your "thin wrapper … acceptable" path) rather than a new Rust group-by kernel, since the partition is a one-pass vectorized mask and a kernel would duplicate `rust_mi_base_cell_of`. Say the word if you'd rather it live in Rust for a single-pass group-by.

https://claude.ai/code/session_012YMu5m9gvKY6dvjYTPRyeR

---
_Generated by [Claude Code](https://claude.ai/code/session_012YMu5m9gvKY6dvjYTPRyeR)_

### DIFF
--- a/mortie/__init__.py
+++ b/mortie/__init__.py
@@ -24,6 +24,7 @@ from .coverage import (
     moc_xor,
     morton_coverage,
     morton_coverage_moc,
+    split_base_cells,
 )
 from .linestring import linestring_coverage
 
@@ -92,6 +93,7 @@ __all__ = [
     'moc_not',
     'common_ancestor',
     'moc_min',
+    'split_base_cells',
     'linestring_coverage',
     'MortonChild',
     'split_children',
--- a/mortie/coverage.py
+++ b/mortie/coverage.py
@@ -599,8 +599,8 @@ def split_base_cells(words, sort=False):
 
     bases = _rustie.rust_mi_base_cell_of(words)
     out = {}
-    # Stable group-by preserving first-seen base-cell order; within each group
-    # the input order is preserved (np.unique keeps the original positions).
+    # Stable group-by: dict.fromkeys yields base cells in first-seen order, and
+    # the boolean mask below keeps each group's words in input order.
     for base in dict.fromkeys(bases.tolist()):
         group = words[bases == base]
         if sort:
--- a/mortie/coverage.py
+++ b/mortie/coverage.py
@@ -541,3 +541,69 @@ def common_ancestor(morton):
 
 # ``moc_min`` is the MOC set-family alias for :func:`common_ancestor` (issue #61).
 moc_min = common_ancestor
+
+
+def split_base_cells(words, sort=False):
+    """Partition a morton set by HEALPix base cell, keyed by each group's
+    :func:`moc_min`.
+
+    The companion to :func:`moc_min` for the cross-base-cell case it refuses:
+    where ``moc_min`` reduces a *single* base cell's words to one ancestor and
+    raises on mixed base cells, ``split_base_cells`` groups the words by base
+    cell and hands back each group untouched.  Every group is keyed by its own
+    ``moc_min`` — the deepest cell enclosing that group — which is self-
+    describing (a packed word the same 64 bits wide as the data) and from which
+    the base cell id is cheap to recover (e.g. ``mort2healpix`` /
+    ``MortonIndexArray.base_cell``).
+
+    Parameters
+    ----------
+    words : array_like
+        Morton indices (mixed order and mixed base cell allowed).
+    sort : bool, optional
+        If ``False`` (default, the faster path) each group keeps the input
+        order of its words.  If ``True`` each group's words are sorted, giving a
+        canonical MOC per base cell.
+
+    Returns
+    -------
+    dict[int, numpy.ndarray]
+        Maps the ``int`` of each group's ``moc_min`` word to that group's
+        ``uint64`` array of words.  Empty input returns ``{}``; a single base
+        cell returns a one-entry dict.
+
+    Raises
+    ------
+    ValueError
+        If a group's ``moc_min`` reduction fails — e.g. ``words`` contains an
+        empty/invalid word (``moc_min`` rejects it).
+
+    See Also
+    --------
+    moc_min : the single-base-cell reduction this partitions for; its mixed-
+        base-cell error points here.
+
+    Examples
+    --------
+    >>> import mortie, numpy as np
+    >>> a = np.atleast_1d(mortie.norm2mort(0, 2, 4))   # one cell in base 2
+    >>> b = np.atleast_1d(mortie.norm2mort(0, 5, 4))   # one cell in base 5
+    >>> groups = mortie.split_base_cells(np.concatenate([a, b]))
+    >>> sorted(int(np.uint64(k) >> np.uint64(60)) - 1 for k in groups)
+    [2, 5]
+    """
+
+    words = np.asarray(words, dtype=np.uint64).ravel()
+    if words.size == 0:
+        return {}
+
+    bases = _rustie.rust_mi_base_cell_of(words)
+    out = {}
+    # Stable group-by preserving first-seen base-cell order; within each group
+    # the input order is preserved (np.unique keeps the original positions).
+    for base in dict.fromkeys(bases.tolist()):
+        group = words[bases == base]
+        if sort:
+            group = np.sort(group)
+        out[int(moc_min(group))] = group
+    return out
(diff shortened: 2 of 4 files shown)


espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the single commit (`9086c72`, +165/-1). **No blocking issues** — the diff matches the binding spec from issue #74: dict keyed by each group's `moc_min`, group-by (not further reduced), base cell extracted via the existing `rust_mi_base_cell_of` (no re-hardcoded prefix), `sort=False` default preserves input order / `sort=True` sorts each group, empty input → `{}`, and `moc_min`'s mixed-base-cell error now names `split_base_cells`. The Rust change is Display-only (no behavioral change to `moc_min`). `coverage.py` is 609 lines (well under ~1000), numpy is the only dependency touched, and the `__all__` export sits next to `moc_min` consistently.

Verified the key-collision concern is a non-issue: each group's `moc_min` is computed within one base cell and its output carries that base cell's prefix, so keys are distinct per base cell — and `test_dict_key_is_moc_min_of_group` + `test_base_cell_recoverable_from_key` pin that. No uint64 signedness/overflow risk: max prefix is 12 (`0xc`), bit 63 is never set, so `>> 60` is safe, and `int(moc_min(...))` keys are unbounded Python ints.

Findings, all diff-scoped nice-to-haves (none blocking):

(1) **Stale comment** — `coverage.py:602-603` (`mortie/coverage.py`). The comment says "the input order is preserved (np.unique keeps the original positions)", but the code uses `dict.fromkeys(...)` + a boolean mask `words[bases == base]`, not `np.unique`. The boolean mask *does* preserve order, so the behavior is correct — only the justification names the wrong primitive. Suggest: "(the boolean mask keeps original positions)".

(2) **Invalid-word path is under-tested for the mixed case** — `coverage.py:600-608`. A `0` (or otherwise invalid-prefix) word maps to the sentinel base `255` via `rust_mi_base_cell_of`, lands in its own `255` group, and that group's `moc_min` raises `ValueError`. `test_invalid_word_raises` covers a lone `[0]`, but not a `0` *mixed with valid words* — there, the iteration order over `dict.fromkeys` means the raise may fire only after some valid groups are built (still raises, just later). Behavior is sensible; a one-line test asserting `[valid, 0]` also raises would lock it in. Optional.

(3) **Doctest robustness (minor)** — `coverage.py:592` (`mortie/coverage.py`). The doctest's `int(np.uint64(k) >> np.uint64(60)) - 1` open-codes the base-cell-from-key extraction the docstring elsewhere recommends doing via `mort2healpix` / `MortonIndexArray.base_cell`. Not wrong (and it documents the cheap path), just slightly at odds with the prose's suggested API. Leave as-is if intended.

On the two "Questions for review" raised in the PR body — both are genuine @espg preference calls, not correctness problems, and should stand for @espg:
- Dict entry order (first-seen vs sorted-by-base-id): dicts are insertion-ordered but the spec only pinned *within-group* order, so first-seen is a defensible default. Design choice, not a bug.
- Python wrapper vs Rust kernel: the wrapper composes two existing vectorized primitives in one pass; a kernel would duplicate `rust_mi_base_cell_of`. Reasonable as-is. Design choice.

Tests are meaningful (assertions compare against `mortie.moc_min`/`np.sort`, not tautologies) and cover the dict-key-is-moc_min invariant, base-cell recoverability, sort False/True (same keys), uint64 dtype, multi-base-cell partition + recombination, single base cell, empty, and the `moc_min`-error-names-`split_base_cells` case via `pytest.raises(match=...)`. Good coverage.

---

espg:
🤖 *from Claude*

Folded the fresh-context self-review's diff-scoped findings (commit `74b0a31`):
- **(1)** Corrected the stale comment at `coverage.py:602-603` — it claimed `np.unique keeps the original positions`, but the group-by uses `dict.fromkeys` (first-seen base-cell order) plus a boolean mask (preserves each group's input order). Comment now names the actual primitives. No behavior change.
- **(2)** Added `test_invalid_word_mixed_with_valid_raises` — pins that a `0` sentinel mixed with *valid* words still raises (its sentinel group's `moc_min` rejects the empty word), not just the lone-`[0]` case.

Green after the fold: `flake8 mortie --select=E9,F63,F7,F82` clean; `TestSplitBaseCells` 11 passed.

Left standing for review:
- The review's finding (3) (the doctest open-codes `>> 60 - 1` rather than routing through `mort2healpix`) is intentional — it keeps the example dependency-free; happy to switch if preferred.
- The two design questions in the PR body remain open for your call: dict **entry order** (first-seen vs sorted-by-base-id — only *within-group* order was pinned by the spec), and the **implementation site** (Python wrapper composing `rust_mi_base_cell_of` + `moc_min` vs a single-pass Rust kernel). Both are preferences, not correctness issues.

Marking `waiting` on these two questions; the feature itself is complete and green.

---

espg:
behavior is correct and as expected (no follow up to this PR)

Issue #73, opened 2026-06-26
Follow up items for the week of June 23, 2026


espg:
🤖 *from Claude*

Sweep of closed PRs from the last 10 days (June 16–25, 2026): #41, #43, #44, #49, #51, #52, #53, #55, #57, #58, #59, #70. Items already tracked in open issues (#54, #61, #68, #48) are excluded. This covers the gaps.

---

### Items

**(1) `on_minor_arc` FP sensitivity in `robust_crossing`** — PR #44 (hemisphere+ PIP)

The Phase-7 SoS hardening covers `robust_crossing`'s four straddle gates but not its antipodal-disambiguation `on_minor_arc`, which still uses raw f64 signs. A probe point lying exactly on a polygon edge's great circle can flip on floating-point noise, silently producing incorrect PIP results with no error raised. PR #44's author explicitly recommended opening a follow-up issue for this. Not a known active bug, but a latent correctness risk on edge-great-circle inputs.

**(2) Versioning automation in CI** — PR #55 / commit `1cf29c2`

Manual `Cargo.toml` version bumps are the current release path. Both the PR body ("...needs to get fixed and automated in a workflow...") and the merge commit message ("Update Cargo.toml... REALLY needs to be automated!!!") flag this as an explicit pain point. Risk: version drift between the tag that triggers `build-wheels.yml` and the `Cargo.toml` that sets the wheel version. A CI workflow or pre-release script that bumps the version and opens a PR would close this.

**(3) `mort2norm` `FORCE_PYTHON` fallback remnant** — PR #49 (Closes #37)

PR #49 removed four of the five `MORTIE_FORCE_PYTHON`-gated Python parity twins. `mort2norm`'s `_python_mort2norm` fallback, the env-var switch, and the `os` import remain in `tools.py` because `mort2norm` was outside #37's stated scope. CLAUDE.md §7 states the Rust extension is the sole runtime path with no Python fallbacks; this remnant is inconsistent with that guarantee. Scope: one function + ~5 lines of cleanup.

**(4) `from_latlon` point-encode path not surfaced in the `morton_index` skin** — PR #51

The pandas `ExtensionArray` and pyarrow `ExtensionType` skin only produces *area* cells (`Kind::Area`). The Rust kernel also supports `Kind::Point` encoding, but there is no `points=True` constructor parameter or any other public API surface for it. Users who want point-precision encoding (e.g. for observation-location indexing) have no public path. PR #51 flagged this explicitly.

[...]

PR #51, opened 2026-06-18
morton_index datatype skin: pandas + Arrow (phases 4 & 5 of #35)

espg:
`Refs #35`

## What this does

The user-facing `morton_index` datatype skin over the existing Rust kernel (`src_rust/src/decimal_morton.rs`, landed in #43). The kernel was Rust-internal with no PyO3 binding; this PR adds the vectorized bindings, a pandas ExtensionArray, and a pyarrow ExtensionType on top, delivering the headline ask of the issue — a registered datatype you can hold in a `Series` and round-trip through parquet.

Honors the settled decisions on the issue thread: dtype registered as **`morton_index`** (the more-accurate name chosen on the thread); **operator scope = comparisons + sort + `coarsen`/`order`/`base_cell` domain ops only**, no arithmetic; **pandas ExtensionArray** + **pyarrow ExtensionType** now, NEP-42 numpy dtype deferred; pandas and pyarrow are **optional extras**, not runtime deps.

### Approach

1. **PyO3 batch bindings** (`src_rust/src/lib.rs`): seven vectorized `rust_mi_*` pyfunctions over numpy `int64`/`uint64` arrays wrapping the kernel — `rust_mi_from_nested`, `rust_mi_to_nested`, `rust_mi_coarsen`, `rust_mi_order_of`, `rust_mi_base_cell_of`, `rust_mi_encode`, `rust_mi_decode`. Each compute region is wrapped in `py.allow_threads`, mirroring the ~16 existing `rust_*` sites; they work with **numpy only**. Storage is `i64` and the kernel takes `u64`, so each reinterprets the bit pattern with an `as` cast (bits preserved exactly).

2. **pandas ExtensionArray** (`mortie/morton_index.py`): an `ExtensionDtype` + `ExtensionArray` registered as `&#34;morton_index&#34;`, backed by `int64` numpy storage (zero-copy over the raw words). Surface:
   - construction from raw words (`from_words` / constructor), from `(nested, depth)` (`from_nested`), and from lat/lon (`from_latlon`, via the Rust `ang2pix` → `from_nested` bridge).
   - comparisons and sort **on the unsigned (`uint64`) view** of the words (see Z-order note below).
   - domain methods `coarsen(k)`, `orders()`/`order()`, `base_cells()`/`base_cell()`, and the mixed-vs-fixed check `is_fixed_order()`.
   - truncated `__repr__` (`base= order=` per element, head/tail with `...` past 6, array-level `order=` / `order=mixed`).
[...]

PR #86, opened 2026-06-27
morton index surface followup: points= encode + __from_arrow__

### LINKED ISSUE #79: morton index surface followup
from #73 

(4) from_latlon point-encode path not surfaced in the morton_index skin — PR https://github.com/espg/mortie/pull/51

The pandas ExtensionArray and pyarrow ExtensionType skin only produces area cells (Kind::Area). The Rust kernel also supports Kind::Point encoding, but there is no points=True constructor parameter or any other public API surface for it. Users who want point-precision encoding (e.g. for observation-location indexing) have no public path. PR https://github.com/espg/mortie/pull/51 flagged this explicitly.

(5) Arrow → pandas __from_arrow__ hook not wired — PR https://github.com/espg/mortie/pull/51

table.to_pandas() on a table with a MortonIndexDtype Arrow column yields a plain int64 Series, not a MortonIndexArray. The __from_arrow__ hook on the dtype was deliberately skipped to stay in scope. This breaks round-tripping through Arrow-typed interchange (e.g. Parquet → pandas). PR https://github.com/espg/mortie/pull/51 flagged it as a known gap.

espg:
Closes #79

Surfaces the two `morton_index` skin gaps flagged in issue #79 (followups from #73 / #51): the `Kind::Point` encode path and the Arrow -> pandas `__from_arrow__` hook, plus proper Arrow null handling (per @espg's "Option B" directive). One PR, three phases.

## Phase (a) - thread `points=True` -> `Kind::Point`

The pandas/pyarrow skins only ever produced area cells; the kernel's point encode (`encode_point`, `decimal_morton.rs:206`) had no public surface. This adds the full bridge:

- **Rust kernel** (`src_rust/src/decimal_morton.rs`): new `from_nested_point(nested)` - the point twin of `from_nested`. It decomposes an order-29 nested id into its 29 stored `0..=3` tuples and calls `encode_point`, so the result decodes with `kind == Kind::Point`, `order == 29`. It guards an oversized nested index (base > 11) the same way `from_nested` does. Unit tests: `from_nested_point_matches_encode_point` (bit-identical to `encode_point`, `Kind::Point`, order 29, shares the nested cell with the area word) and `from_nested_point_rejects_oversized_nested`.
- **PyO3 binding** (`src_rust/src/lib.rs`): a dedicated single-purpose `rust_mi_from_nested_point(nested_array)`, mirroring `rust_mi_from_nested` (no `point: bool` flag), registered alongside it.
- **Python** (`mortie/morton_index.py`): `MortonIndexArray.from_latlon` gains `points=False`. When `points=True`, lat/lon is hashed at order 29 and routed through `rust_mi_from_nested_point`; an explicit `order != 29` passed with `points=True` raises `ValueError`. The default `order` is already `MAX_ORDER` (29), so the point path needs no extra argument.
- **Tests** (`mortie/tests/test_morton_index.py`, `TestPointEncode`): `points=True` words decode as `Kind::Point`/order 29; points share a nested cell with the matching area word; coarsening a point to a lower order yields an area cell; `points=False` (and the default) stay `Kind::Area`; `order != 29` + `points=True` raises.

### Answer to the `encode_point` order-arg question

**No, `encode_point` does not take an `order` argument.** Its signature is `pub fn encode_point(base_cell: u8, tuples: &[u8]) -> u64` (`decimal_morton.rs:206`). Point encoding is **order-29-only by construction**: it asserts `tuples.len() >= MAX_ORDER` (29) and hardcodes `MAX_ORDER` into both the body pack and the point suffix (`pack_prefix_body(base_cell, tuples, MAX_ORDER) | build_point_suffix(tuples[27], tuples[28])`). A point is the max-resolution "this exact location, no area claim" encoding, so there is no coarser-order point - that is why no order parameter exists. (By contrast the area `encode`/`from_nested` do take an order/depth, because area cells exist at every order 0..=29.) The Python side mirrors this: `points=True` forces order 29 and rejects any explicit `order != 29`, so a caller cannot ask for a non-existent "order-N point".

## Phase (b) - wire `__from_arrow__`

Before: `table.to_pandas()` on a `morton_index` Arrow column produced a plain int64 Series (the dtype had no `__from_arrow__` hook). Now:

- **Python** (`mortie/morton_index.py`): `MortonIndexDtype.__from_arrow__(self, array)` accepts a `pa.Array` or `pa.ChunkedArray`, delegates to `mortie.arrow.to_morton_index`, iterates chunks and concatenates via `MortonIndexArray._concat_same_type` for the chunked case, and returns a `MortonIndexArray`. The pyarrow import stays lazy (reuses `arrow._require_pyarrow`), so `morton_index.py` remains numpy-only importable.
- **Tests** (`mortie/tests/test_arrow.py`, `TestFromArrowHook`): `table.to_pandas()` yields a `morton_index`-dtyped Series backed by a `MortonIndexArray` with bit-identical words; `__from_arrow__` handles both plain and chunked arrays; full Parquet -> `read_table` -> `to_pandas()` preserves words + dtype.

## Phase (c) - Arrow null <-> sentinel round-trip (@espg "Option B")

Per @espg's directive (https://github.com/espg/mortie/pull/86#issuecomment-4845674053) answering the open "Questions for review" item: implement proper null handling as part of this PR. Before, a `pa.Array` null neither survived `to_morton_index` (a uint64 array with a null buffer cannot go straight to numpy) nor was emitted by `from_morton_index` for sentinel/`isna` elements. Now the null<->sentinel contract holds both directions:

- **`from_morton_index`** (`mortie/arrow.py`): builds the pyarrow storage with a validity `mask` set wherever the word equals `MortonIndexArray._SENTINEL` (the all-zero empty word `isna()` treats as missing). So an `isna()`/sentinel element going OUT becomes an Arrow null. Works from a `MortonIndexArray` (its `isna()` drives the mask) or a raw `uint64` array; the docstring notes the mask is read off the words and that re-wrapping an already-built Arrow array goes through `to_morton_index`, not here.
- **`to_morton_index`** (`mortie/arrow.py`): `fill_null(_SENTINEL)` before materializing, so an Arrow null coming IN lands as the all-zero sentinel word and `MortonIndexArray.isna()` reports it missing. `__from_arrow__` (Array + ChunkedArray + parquet) inherits this since it delegates here.
- The lazy pyarrow import is preserved; numpy stays the only runtime dep.
- **Tests** (`mortie/tests/test_arrow.py`, `TestNullRoundTrip`): sentinel words emit Arrow nulls (from raw words and from a `MortonIndexArray` via `isna()`); Arrow nulls come back as the sentinel with `isna()` True; `__from_arrow__` on null-bearing plain `pa.Array` and `pa.ChunkedArray`; full Parquet -> `read_table` -> `to_pandas()` preserves the null buffer and `isna()`; plus the all-null, no-null (mask path leaves null-free arrays byte-identical), and empty plain-array edges.

## Phases checklist

- [x] Phase (a): `from_nested_point` kernel + `rust_mi_from_nested_point` binding + `from_latlon(points=...)` + tests
- [x] Phase (b): `MortonIndexDtype.__from_arrow__` (Array + ChunkedArray) + tests
- [x] Phase (c): Arrow null <-> empty sentinel round-trip in `from_morton_index` / `to_morton_index` + tests

## How it was tested

All run in an isolated venv on this branch:

- `cargo fmt` (applied) + `cargo test` -> **175 passed** (includes the 2 new kernel tests)
- `cargo clippy` -> only pre-existing warnings (the `useless_vec` warnings in `coverage/tests.rs` and the `useless_conversion` on `PyResult` shared by all 31 existing `#[pyfunction]` bindings); the new binding matches that established pattern, so no new warning is introduced.
- `maturin develop --release` -> builds clean
- `flake8 mortie --select=E9,F63,F7,F82` -> clean; non-blocking `--max-line-length=88` pass clean on touched files
- `pytest -q` (phase c, after the self-review fold) -> **462 passed, 8 skipped** (`test_arrow.py` now 25 tests incl. the 10 `TestNullRoundTrip` cases; `test_morton_index.py` 31)

## Questions for review

- **Pre-existing clippy warnings, untouched.** `cargo clippy` reports 31 `useless_conversion` warnings on `PyResult` (every existing binding) and `useless_vec` warnings in `coverage/tests.rs`. None are introduced by this change; the new `rust_mi_from_nested_point` deliberately mirrors the existing `rust_mi_from_nested` style rather than diverging. Left as-is per the "don't fix unrelated CI noise / match the surrounding code" conventions - flag if you'd rather I clean up the binding pattern repo-wide in a separate PR.
- **`__from_arrow__` reuses `arrow._require_pyarrow`** (a leading-underscore helper) to keep the lazy-import contract in one place. If you'd prefer a public accessor instead of importing the private name, say so.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QP5rg6JZ69FxVtjH5mrixz)_

### DIFF
--- a/mortie/arrow.py
+++ b/mortie/arrow.py
@@ -124,7 +124,10 @@ def from_morton_index(array):
     same ``uint64`` words. ``array`` may also be a raw ``uint64`` array-like of
     words. Missing elements -- a ``MortonIndexArray`` for which :meth:`isna` is
     True, i.e. the all-zero empty sentinel word -- emit Arrow nulls, so a null
-    survives the round-trip back through :func:`to_morton_index`.
+    survives the round-trip back through :func:`to_morton_index`. (The missing
+    mask is read off the ``uint64`` words, so a sentinel word in a raw array is
+    treated as a null too; an already-built Arrow array goes back through
+    :func:`to_morton_index`, not here.)
     """
     pa = _require_pyarrow()
     ext_type = _build_type()
--- a/mortie/arrow.py
+++ b/mortie/arrow.py
@@ -122,12 +122,19 @@ def from_morton_index(array):
 
     Builds a pyarrow ``ExtensionArray`` of the ``morton_index`` type over the
     same ``uint64`` words. ``array`` may also be a raw ``uint64`` array-like of
-    words.
+    words. Missing elements -- a ``MortonIndexArray`` for which :meth:`isna` is
+    True, i.e. the all-zero empty sentinel word -- emit Arrow nulls, so a null
+    survives the round-trip back through :func:`to_morton_index`.
     """
     pa = _require_pyarrow()
     ext_type = _build_type()
-    data = getattr(array, "_data", array)
-    storage = pa.array(np.asarray(data, dtype=np.uint64), type=pa.uint64())
+    data = np.asarray(getattr(array, "_data", array), dtype=np.uint64)
+    # The empty sentinel (all-zero word, prefix 0) is the missing value on the
+    # pandas side; mirror it as an Arrow null so isna() round-trips both ways.
+    from .morton_index import MortonIndexArray
+
+    mask = data == MortonIndexArray._SENTINEL
+    storage = pa.array(data, type=pa.uint64(), mask=mask)
     return pa.ExtensionArray.from_storage(ext_type, storage)
 
 
@@ -136,12 +143,17 @@ def to_morton_index(array):
 
     Accepts the extension array (or its plain ``uint64`` storage) and returns the
     pandas-side :class:`~mortie.morton_index.MortonIndexArray` over the same
-    words.
+    words. Arrow nulls come back as the all-zero empty sentinel word, so the
+    pandas :meth:`isna` reports them as missing.
     """
     _require_pyarrow()
     from .morton_index import MortonIndexArray
 
     storage = getattr(array, "storage", array)
+    # Fill nulls with the empty sentinel before materializing: a uint64 array
+    # with a null buffer cannot go straight to numpy.
+    if storage.null_count:
+        storage = storage.fill_null(int(MortonIndexArray._SENTINEL))
     words = storage.to_numpy(zero_copy_only=False).astype(np.uint64, copy=False)
     return MortonIndexArray(words)
 
--- a/mortie/morton_index.py
+++ b/mortie/morton_index.py
@@ -90,6 +90,25 @@ def _build_classes():
         def construct_array_type(cls):
             return MortonIndexArray
 
+        def __from_arrow__(self, array):
+            """Build a :class:`MortonIndexArray` from a pyarrow array.
+
+            This is the hook pandas calls on ``table.to_pandas()`` for a column
+            tagged with the ``morton_index`` Arrow extension type, so the words
+            land back as a ``MortonIndexArray`` (not a plain int64 Series).
+            Accepts a ``pa.Array`` or a ``pa.ChunkedArray``; the pyarrow import
+            stays lazy so this module remains numpy-only importable.
+            """
+            from .arrow import _require_pyarrow, to_morton_index
+
+            pa = _require_pyarrow()
+            if isinstance(array, pa.ChunkedArray):
+                parts = [to_morton_index(chunk) for chunk in array.chunks]
+                if not parts:
+                    return MortonIndexArray(np.empty(0, dtype=np.uint64))
+                return MortonIndexArray._concat_same_type(parts)
+            return to_morton_index(array)
+
     class MortonIndexArray(ExtensionArray):
         """An array of packed 64-bit ``morton_index`` MOC words.
 
--- a/mortie/morton_index.py
+++ b/mortie/morton_index.py
@@ -129,19 +129,36 @@ def _build_classes():
             return cls(words, copy=copy)
 
         @classmethod
-        def from_latlon(cls, lat, lon, order=MAX_ORDER):
+        def from_latlon(cls, lat, lon, order=MAX_ORDER, points=False):
             """Hash lat/lon (degrees) to ``morton_index`` words at ``order``.
 
             Routes through the Rust ``healpix`` bridge: lat/lon -> NESTED ids ->
             packed words, so it matches the cross-library nested representation.
+
+            With ``points=False`` (the default) the result is an order-``order``
+            **area** cell (``Kind::Area``). With ``points=True`` the location is
+            encoded as a max-resolution **point** (``Kind::Point``); point
+            encoding is order-29-only, so an explicit ``order != 29`` raised
+            together with ``points=True`` is a ``ValueError`` (the default
+            ``order`` is ``MAX_ORDER`` so the point path needs no extra argument).
             """
+            if points and int(order) != MAX_ORDER:
+                raise ValueError(
+                    "points=True encodes an order-29 point; pass order=29 "
+                    "(the default) or omit it"
+                )
             lat = np.ascontiguousarray(np.asarray(lat), dtype=np.float64)
             lon = np.ascontiguousarray(np.asarray(lon), dtype=np.float64)
             if lat.shape != lon.shape:
                 raise ValueError("lat and lon must have the same shape")
-            nested = _rustie.rust_ang2pix(int(order), lon, lat)
-            nested = np.ascontiguousarray(nested, dtype=np.uint64)
-            words = _rustie.rust_mi_from_nested(nested, int(order))
+            if points:
+                nested = _rustie.rust_ang2pix(MAX_ORDER, lon, lat)
+                nested = np.ascontiguousarray(nested, dtype=np.uint64)
+                words = _rustie.rust_mi_from_nested_point(nested)
+            else:
+                nested = _rustie.rust_ang2pix(int(order), lon, lat)
+                nested = np.ascontiguousarray(nested, dtype=np.uint64)
+                words = _rustie.rust_mi_from_nested(nested, int(order))
             return cls(words)
 
         @classmethod
--- a/mortie/tests/test_arrow.py
+++ b/mortie/tests/test_arrow.py
@@ -169,7 +169,14 @@ class TestFromArrowHook:
 # Null / NA round-trip: sentinel <-> Arrow null (issue #79)
 # ---------------------------------------------------------------------------
 
-SENTINEL = 0  # the all-zero empty word MortonIndexArray.isna() treats as missing
+# The missing value is the all-zero empty word MortonIndexArray.isna() treats
+# as missing; derive it from the array constant (pandas-gated) rather than
+# hardcoding 0, falling back to the documented all-zero value when pandas is
+# absent so this module still imports numpy/pyarrow-only.
+try:
+    SENTINEL = int(mortie.MortonIndexArray._SENTINEL)
+except ImportError:  # pragma: no cover - pandas absent
+    SENTINEL = 0
 
 
 def _words_with_gap():
@@ -187,6 +194,28 @@ class TestNullRoundTrip:
         arr = marrow.from_morton_index(words)
         assert arr.null_count == 2
         assert arr.is_null().to_pylist() == [w == SENTINEL for w in words]
+
+    def test_no_null_words_round_trip_unchanged(self):
+        # The mask path leaves a null-free array byte-identical both ways.
+        words = _sample_words()
+        arr = marrow.from_morton_index(words)
+        assert arr.null_count == 0
+        np.testing.assert_array_equal(marrow.to_morton_index(arr)._data, words)
+
+    def test_all_null_array_round_trips_to_sentinel(self):
+        # Every element missing -> every word the sentinel, isna() all True.
+        pytest.importorskip("pandas")
+        arr = pa.array([None, None, None], type=pa.uint64())
+        mia = marrow.to_morton_index(arr)
+        np.testing.assert_array_equal(mia._data, np.zeros(3, dtype=np.uint64))
+        assert mia.isna().all()
+
+    def test_empty_plain_array_round_trips(self):
+        # The empty plain-Array path (distinct from the empty-chunked case).
+        empty = pa.array([], type=pa.uint64())
+        mia = marrow.to_morton_index(empty)
+        assert len(mia) == 0
+        assert marrow.from_morton_index(mia).null_count == 0
 
     def test_mortonindexarray_isna_emits_arrow_nulls(self):
         # Same, but starting from a MortonIndexArray (isna() drives the mask).
--- a/mortie/tests/test_arrow.py
+++ b/mortie/tests/test_arrow.py
@@ -116,6 +116,48 @@ class TestParquet:
 
 
 # ---------------------------------------------------------------------------
+# __from_arrow__ hook: table.to_pandas() yields a MortonIndexArray (issue #79)
+# ---------------------------------------------------------------------------
+
+
+class TestFromArrowHook:
+    def test_to_pandas_yields_morton_index_series(self):
+        pytest.importorskip("pandas")
+        words = _sample_words()
+        table = pa.table({"mi": marrow.from_morton_index(words)})
+        series = table.to_pandas()["mi"]
+        # Not a plain int64 column: the extension dtype survives.
+        assert series.dtype.name == "morton_index"
+        assert isinstance(series.array, mortie.MortonIndexArray)
+        np.testing.assert_array_equal(series.array._data, words)
+
+    def test_from_arrow_handles_chunked_array(self):
+        pytest.importorskip("pandas")
+        words = _sample_words(n=8)
+        ext = marrow.from_morton_index(words)
+        chunked = pa.chunked_array([ext[:3], ext[3:]])
+        mia = mortie.MortonIndexDtype().__from_arrow__(chunked)
+        assert isinstance(mia, mortie.MortonIndexArray)
+        np.testing.assert_array_equal(mia._data, words)
+
+    def test_from_arrow_handles_plain_array(self):
+        pytest.importorskip("pandas")
+        words = _sample_words()
+        mia = mortie.MortonIndexDtype().__from_arrow__(marrow.from_morton_index(words))
+        assert isinstance(mia, mortie.MortonIndexArray)
+        np.testing.assert_array_equal(mia._data, words)
+
+    def test_parquet_to_pandas_full_round_trip(self, tmp_path):
+        pytest.importorskip("pandas")
+        words = _sample_words()
+        path = tmp_path / "mi.parquet"
+        pq.write_table(pa.table({"mi": marrow.from_morton_index(words)}), path)
+        series = pq.read_table(path).to_pandas()["mi"]
+        assert series.dtype.name == "morton_index"
+        np.testing.assert_array_equal(series.array._data, words)
+
+
+# ---------------------------------------------------------------------------
 # Optional-extra contract
 # ---------------------------------------------------------------------------
 
--- a/mortie/tests/test_arrow.py
+++ b/mortie/tests/test_arrow.py
@@ -147,6 +147,13 @@ class TestFromArrowHook:
         assert isinstance(mia, mortie.MortonIndexArray)
         np.testing.assert_array_equal(mia._data, words)
 
+    def test_from_arrow_handles_empty_chunked_array(self):
+        pytest.importorskip("pandas")
+        empty = pa.chunked_array([], type=marrow.morton_index_type())
+        mia = mortie.MortonIndexDtype().__from_arrow__(empty)
+        assert isinstance(mia, mortie.MortonIndexArray)
+        assert len(mia) == 0
+
     def test_parquet_to_pandas_full_round_trip(self, tmp_path):
         pytest.importorskip("pandas")
         words = _sample_words()
@@ -154,6 +161,7 @@ class TestFromArrowHook:
         pq.write_table(pa.table({"mi": marrow.from_morton_index(words)}), path)
         series = pq.read_table(path).to_pandas()["mi"]
         assert series.dtype.name == "morton_index"
+        assert isinstance(series.array, mortie.MortonIndexArray)
         np.testing.assert_array_equal(series.array._data, words)
 
 
--- a/mortie/tests/test_arrow.py
+++ b/mortie/tests/test_arrow.py
@@ -166,6 +166,76 @@ class TestFromArrowHook:
 
 
 # ---------------------------------------------------------------------------
+# Null / NA round-trip: sentinel <-> Arrow null (issue #79)
+# ---------------------------------------------------------------------------
+
+SENTINEL = 0  # the all-zero empty word MortonIndexArray.isna() treats as missing
+
+
+def _words_with_gap():
+    """Sample words with the empty sentinel planted at two positions."""
+    words = _sample_words().copy()
+    words[2] = SENTINEL
+    words[5] = SENTINEL
+    return words
+
+
+class TestNullRoundTrip:
+    def test_isna_words_emit_arrow_nulls(self):
+        # An isna() element (sentinel word) goes OUT as an Arrow null.
+        words = _words_with_gap()
+        arr = marrow.from_morton_index(words)
+        assert arr.null_count == 2
+        assert arr.is_null().to_pylist() == [w == SENTINEL for w in words]
+
+    def test_mortonindexarray_isna_emits_arrow_nulls(self):
+        # Same, but starting from a MortonIndexArray (isna() drives the mask).
+        pytest.importorskip("pandas")
+        mia = mortie.MortonIndexArray(_words_with_gap())
+        arr = marrow.from_morton_index(mia)
+        assert arr.null_count == 2
+        assert arr.is_null().to_pylist() == list(mia.isna())
+
+    def test_arrow_nulls_come_back_as_sentinel(self):
+        # A null IN comes back as the sentinel so isna() is True.
+        pytest.importorskip("pandas")
+        words = _words_with_gap()
+        arr = marrow.from_morton_index(words)
+        mia = marrow.to_morton_index(arr)
+        np.testing.assert_array_equal(mia._data, words)
+        np.testing.assert_array_equal(mia.isna(), words == SENTINEL)
+
+    def test_from_arrow_handles_nulls_plain_array(self):
+        pytest.importorskip("pandas")
+        words = _words_with_gap()
+        arr = marrow.from_morton_index(words)
+        mia = mortie.MortonIndexDtype().__from_arrow__(arr)
+        np.testing.assert_array_equal(mia._data, words)
+        np.testing.assert_array_equal(mia.isna(), words == SENTINEL)
+
+    def test_from_arrow_handles_nulls_chunked_array(self):
+        pytest.importorskip("pandas")
+        words = _words_with_gap()
+        ext = marrow.from_morton_index(words)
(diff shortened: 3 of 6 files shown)


mortie/morton_index.py
```diff
@@ -129,19 +148,36 @@ def from_words(cls, words, copy=False):
...
             packed words, so it matches the cross-library nested representation.
+
+            With ``points=False`` (the default) the result is an order-``order``
+            **area** cell (``Kind::Area``). With ``points=True`` the location is
+            encoded as a max-resolution **point** (``Kind::Point``); point
+            encoding is order-29-only, so an explicit ``order != 29`` raised
+            together with ``points=True`` is a ``ValueError`` (the default
+            ``order`` is ``MAX_ORDER`` so the point path needs no extra argument).
             """
+            if points and int(order) != MAX_ORDER:
```
espg, line 164: 🤖 *from Claude (review)*

The guard is `points and int(order) != MAX_ORDER`. Worth being explicit that an *explicit* `order=29, points=True` is accepted silently (it can't be distinguished from the default, since both resolve to `int(order)==29`). The docstring already says "pass order=29 (the default) or omit it", so this is intended — just confirming the edge case is by design and the test `test_points_true_with_non29_order_raises` covers only the reject path (order=12), not the accept-explicit-29 path. A one-line test that `from_latlon(..., order=29, points=True)` succeeds and yields points would lock that contract in.

Also note `int(order)` is evaluated before the lat/lon coercion, so a non-integer `order` with `points=True` raises a `TypeError` (from `int()`) rather than the `ValueError` — fine, but a different exception type than the explicit guard suggests.

---

mortie/tests/test_morton_index.py
```diff
@@ -87,6 +87,53 @@ def test_latlon_matches_cdshealpix_nested(self):
...
+    LATS = np.array([0.0, 41.8, -41.8, 80.0, -80.0, 12.3, -67.9])
+    LONS = np.array([0.0, 45.0, 135.0, 200.0, 305.0, 91.5, 270.2])
+
+    def test_points_true_yields_order29_points(self):
+        a = MIA.from_latlon(self.LATS, self.LONS, points=True)
+        # rust_mi_decode returns (base_cells, orders, kinds, tuples); kind 1=point.
+        _, orders, kinds, _ = _rustie.rust_mi_decode(a._data)
+        np.testing.assert_array_equal(orders, np.full(len(self.LATS), 29))
+        np.testing.assert_array_equal(kinds, np.ones(len(self.LATS), dtype=np.uint8))
+
```
espg, line 102: 🤖 *from Claude (review)*

Coverage is solid — `test_points_true_yields_order29_points` asserts both `order==29` and `kind==1` (Point), `test_points_false_default_is_area` pins the area default, and `test_points_coarsen_and_to_nested` confirms coarsen→area. The `kinds==1` assertion is what guards against a regression that silently routed `points=True` back through the area path.

Two small additions worth considering (non-blocking): (1) a direct assertion that `points=True` words *differ* from the `order=29` area words for the same location (the kind bit makes them distinct; asserting `pts._data != area._data` documents that intent next to `test_points_share_nested_cell_with_area`, which only checks the shared nested cell). (2) The `LONS` fixture uses values >180 (200.0, 305.0, 270.2) — confirm that's the intended 0–360 convention `rust_ang2pix` expects, since the area tests elsewhere may use a different range.

---

mortie/morton_index.py
```diff
@@ -90,6 +90,25 @@ class MortonIndexDtype(ExtensionDtype):
...
+            """
+            from .arrow import _require_pyarrow, to_morton_index
+
+            pa = _require_pyarrow()
+            if isinstance(array, pa.ChunkedArray):
+                parts = [to_morton_index(chunk) for chunk in array.chunks]
+                if not parts:
+                    return MortonIndexArray(np.empty(0, dtype=np.uint64))
+                return MortonIndexArray._concat_same_type(parts)
+            return to_morton_index(array)
```
espg, line 110: 🤖 *from Claude (review)*

Null/NA round-trip is silently unhandled here, and this is the one path where it could bite. `to_morton_index` does `storage.to_numpy(zero_copy_only=False)` on the chunk/array storage; if an Arrow `morton_index` column carries nulls (a real scenario once a `MortonIndexArray` with `isna()` sentinels is written to Parquet and read back), the uint64 `to_numpy` either raises or fills with an arbitrary value rather than mapping back to the all-zero `_SENTINEL` that `isna()` expects. Phase (b)'s stated acceptance is words+dtype preservation for non-null data, so this isn't a blocker — but the whole Arrow suite (`test_arrow.py`) never exercises a null, so the gap is invisible. Standing for @espg: is null/NA Arrow interop in scope for issue #79, or a follow-up? If in scope, a test writing a column with a `pd.NA` element through Parquet and asserting `isna()` survives would pin it.

---

mortie/morton_index.py
```diff
@@ -90,6 +90,25 @@ class MortonIndexDtype(ExtensionDtype):
...
+
+            This is the hook pandas calls on ``table.to_pandas()`` for a column
+            tagged with the ``morton_index`` Arrow extension type, so the words
+            land back as a ``MortonIndexArray`` (not a plain int64 Series).
+            Accepts a ``pa.Array`` or a ``pa.ChunkedArray``; the pyarrow import
+            stays lazy so this module remains numpy-only importable.
+            """
+            from .arrow import _require_pyarrow, to_morton_index
+
+            pa = _require_pyarrow()
```
espg, line 104: 🤖 *from Claude (review)*

Lazy-import discipline is correctly preserved: the `from .arrow import _require_pyarrow, to_morton_index` is method-local, the module top still imports only `numpy` and `_rustie`, so `import mortie.morton_index` stays numpy-only. Reusing the sibling module's `_require_pyarrow` helper matches the existing intra-package pattern (arrow.py and morton_index.py both gate optional deps through their own `_require_*`), so the private-name import is consistent with the codebase rather than a smell. One small note: `to_morton_index` already calls `_require_pyarrow()` internally, so the `pa` bound here is needed only for the `isinstance(array, pa.ChunkedArray)` discriminator — which is legitimate, just worth knowing it isn't doing double duty as the import guard for the conversion itself.

---

espg (commented):
🤖 *from Claude (review)*

Phase 1 review (commit 3ab4bea, issue #79 phase (a)). The change is correct and meets the acceptance criteria: `from_latlon(..., points=True)` produces `Kind::Point` words at order 29, and `points=False` (default) is unchanged `Kind::Area`.

Verified:
- `from_nested_point` tuple decomposition is bit-faithful to `encode_point` (orders 28/29 land in `tuples[27]`/`tuples[28]`); pair extraction mirrors `from_nested`'s body loop; the base>11 guard mirrors `from_nested`.
- The PyO3 binding `rust_mi_from_nested_point` faithfully mirrors `rust_mi_from_nested` — `allow_threads`, `catch_unwind`, `par_iter`, and the `panic_msg`/`PyValueError` error path all match.
- Kernel tests assert `encode_point` equality, `Kind::Point`, `order==29`, nested round-trip, and the oversized-nested panic. Python tests assert order 29 + kind Point, area-unchanged default, coarsen→area, and the non-29 reject.
- No flake8 style-pass (88-col) violations in the Python diff; no over-long Rust lines.

Findings (all minor / non-blocking, details inline):
(1) `from_nested_point` silently masks order-≥30 bits via `& 3` rather than rejecting them — consistent with `from_nested`, flagged only to make it a conscious choice.
(2) The ValueError guard accepts an explicit `order=29, points=True` (indistinguishable from default); a non-integer `order` raises `TypeError` from `int()` rather than the guard's `ValueError`. Both look intended.
(3) Test nice-to-haves: assert point words *differ* from the order-29 area words for the same location, add the accept-explicit-29 path, and confirm the >180 longitude fixture matches `rust_ang2pix`'s expected range.

Nothing here blocks advancing to the next phase. Note `decimal_morton.rs` is already ~1568 lines (pre-existing, not from this diff) — standing observation for @espg per §4, not a finding against this PR.

---

mortie/tests/test_arrow.py
espg, line 1: 🤖 *from Claude (review)*

Test coverage for phase (b) is solid and meets the acceptance criteria: `test_to_pandas_yields_morton_index_series` asserts all three of `dtype.name == "morton_index"`, `isinstance(series.array, MortonIndexArray)`, and bit-identical words; `test_from_arrow_handles_chunked_array` exercises a 2-chunk `ChunkedArray` and confirms both order preservation across chunks and bit-identical words; `test_from_arrow_handles_plain_array` covers the `pa.Array` branch; and `test_parquet_to_pandas_full_round_trip` pins the Parquet -> `to_pandas()` path end to end.

Two non-blocking nits:
(1) `test_parquet_to_pandas_full_round_trip` asserts `dtype.name` and compares `series.array._data`, but unlike the in-memory test it doesn't assert `isinstance(series.array, mortie.MortonIndexArray)` directly (it's only implied by the `._data` access). One explicit `isinstance` line would make the full-roundtrip path assert the backing-array type the same way the in-memory test does.
(2) The empty-`ChunkedArray` branch (`pa.chunked_array([], type=morton_index)` -> empty `MortonIndexArray`) is handled in the code but never tested. A one-line test would lock that edge in.

---

espg (commented):
🤖 *from Claude (review)*

Phase 2 review (commit ddfda07, issue #79 phase (b): wire `MortonIndexDtype.__from_arrow__` so `table.to_pandas()` on a `morton_index` column yields a `MortonIndexArray`). The change is correct and meets the stated acceptance criteria.

Verified:
- pyarrow stays **lazy** — the `from .arrow import _require_pyarrow, to_morton_index` is method-local; module top still imports only `numpy` and `_rustie`, so `import mortie.morton_index` remains numpy-only importable. No module-level pyarrow import added.
- `__from_arrow__` is correctly a method of `MortonIndexDtype` (8-space indent, same level as `construct_array_type`).
- **ChunkedArray** path iterates `array.chunks` in order, converts each via `to_morton_index`, and concatenates with `MortonIndexArray._concat_same_type` (which does `np.concatenate` in list order) — order and words preserved across chunks.
- **Empty ChunkedArray** (zero chunks) returns an empty `MortonIndexArray(np.empty(0, dtype=np.uint64))`.
- **Plain `pa.Array`** path returns `to_morton_index(array)`, i.e. a `MortonIndexArray`.
- Wiring is complete: `arrow.py`'s `MortonIndexType.to_pandas_dtype()` returns `MortonIndexDtype()`, which is what routes `to_pandas()` into this hook.
- Tests cover dtype-name + backing-array-type + bit-identical words (in-memory), the chunked order-preservation path, the plain-array path, and the full Parquet -> `to_pandas()` round-trip.
- Line lengths within the 88-col style pass; no `E9/F63/F7/F82` concerns; commit message is title-only per §3.

Findings (all minor / non-blocking, details inline):
(1) Null/NA Arrow round-trip is silently unhandled (`to_morton_index` -> `storage.to_numpy(zero_copy_only=False)` won't map Arrow nulls back to the all-zero `_SENTINEL` that `isna()` expects); the whole Arrow suite never exercises a null, so the gap is invisible. Phase (b)'s acceptance is non-null words+dtype, so not a blocker — standing for @espg whether null interop is in scope for issue #79 or a follow-up.
(2) Test nice-to-haves: assert `isinstance(series.array, MortonIndexArray)` directly in the Parquet round-trip test, and add a one-line empty-`ChunkedArray` test to lock that handled-but-untested edge.
(3) Note on the private `_require_pyarrow` import: consistent with the codebase's intra-package optional-dep pattern, and `pa` is genuinely needed for the `isinstance(..., pa.ChunkedArray)` discriminator (`to_morton_index` already guards pyarrow internally).

Nothing here blocks the phase. None of the findings imply a side-effecting action; items for @espg are flagged as standing, not acted on.

---

espg:
🤖 *from Claude*

Folded the diff-scoped findings from the two adversarial self-reviews into commit `91b73a7` ("fold self-review on #79"):

- Phase (a): added `test_point_words_differ_from_area_words` (asserts the packed point word `!=` the order-29 area word for the same location) and `test_points_true_explicit_order29_is_accepted` (locks the accept-explicit-`order=29` path next to the existing reject-`order=12` test).
- Phase (b): `test_parquet_to_pandas_full_round_trip` now asserts `isinstance(series.array, MortonIndexArray)` directly (matching the in-memory test), and added `test_from_arrow_handles_empty_chunked_array` to cover the empty-`ChunkedArray` branch.

Addressed-by-design (no code change), grounding the self-review notes:

- The `points and int(order) != MAX_ORDER` guard accepting an explicit `order=29` is intentional (the docstring says "pass order=29 (the default) or omit it"); the new accept-explicit-29 test pins it.
- `from_nested_point` masking order-≥30 bits via `& 3` mirrors `from_nested` exactly — kept consistent on purpose.
- The `LONS` fixture values `>180` (200.0, 305.0, 270.2) use the 0–360 convention `rust_ang2pix` expects; this is the same spread the existing `TestVsCdshealpix.test_latlon_matches_cdshealpix_nested` already uses, so it's consistent with the area path.

Standing for review (1): **Null/NA Arrow interop.** The phase-(b) self-review flagged that `__from_arrow__` / `to_morton_index` don't round-trip a null/NA element back to the all-zero `_SENTINEL` that `isna()` expects, and the whole `test_arrow.py` suite never exercises a null. Issue #79 phase (b)'s stated acceptance is words+dtype preservation for non-null data, so I scoped this out. Options: (a) leave it for a dedicated follow-up issue (Arrow null <-> sentinel is its own contract spanning `from_morton_index`/`to_morton_index` too), or (b) fold a null round-trip (write a `pd.NA`-bearing column through Parquet, assert `isna()` survives) into this PR. I left it out pending your call — happy to do (b) here if you'd prefer it in scope.

All phases complete; local gates green (`cargo test` 175 passed, `cargo clippy` only pre-existing warnings, `flake8 --select=E9,F63,F7,F82` clean, `pytest` 452 passed / 8 skipped). Draft kept in draft.

---

espg:
Option B — let’s implement proper null handling as part of this PR

mortie/arrow.py
```diff
@@ -122,12 +122,19 @@ def from_morton_index(array):
...
-    data = getattr(array, "_data", array)
-    storage = pa.array(np.asarray(data, dtype=np.uint64), type=pa.uint64())
+    data = np.asarray(getattr(array, "_data", array), dtype=np.uint64)
+    # The empty sentinel (all-zero word, prefix 0) is the missing value on the
+    # pandas side; mirror it as an Arrow null so isna() round-trips both ways.
+    from .morton_index import MortonIndexArray
+
+    mask = data == MortonIndexArray._SENTINEL
+    storage = pa.array(data, type=pa.uint64(), mask=mask)
     return pa.ExtensionArray.from_storage(ext_type, storage)
```
espg, line 141: 🤖 *from Claude (review)*

The OUT direction is correct: `mask = data == _SENTINEL` yields a bool numpy array, pyarrow accepts it as the null mask, and empty/all-null/no-null inputs all behave (verified: empty -> `null_count == 0`, all-null -> all nulls, no-null -> unchanged).

One scoping note for the docstring (lines 125-127): it now reads as though `from_morton_index` handles "missing elements" generally, but the function only handles the `MortonIndexArray` / raw-words input path. The `getattr(array, "_data", array)` fallthrough means that if an Arrow `MortonIndexExtArray` (which can itself carry nulls) is passed in, `np.asarray(ext, dtype=np.uint64)` would materialize it without honoring its null buffer. That's pre-existing behavior and likely out of scope for Option B, but since the docstring now advertises null handling, consider a one-line clarification that the null<->sentinel mapping applies to pandas-side / raw-word input (not re-wrapping an Arrow array). Not blocking.

---

mortie/arrow.py
```diff
@@ -136,12 +143,17 @@ def to_morton_index(array):
...
     """
     _require_pyarrow()
     from .morton_index import MortonIndexArray
 
     storage = getattr(array, "storage", array)
+    # Fill nulls with the empty sentinel before materializing: a uint64 array
+    # with a null buffer cannot go straight to numpy.
+    if storage.null_count:
+        storage = storage.fill_null(int(MortonIndexArray._SENTINEL))
     words = storage.to_numpy(zero_copy_only=False).astype(np.uint64, copy=False)
```
espg, line 160: 🤖 *from Claude (review)*

The IN direction holds, and it's pleasingly robust: `getattr(array, "storage", array)` followed by `.null_count` / `.fill_null` / `.to_numpy` works on **both** a plain `pa.Array` (extension storage) and a `ChunkedArray` reaching `to_morton_index` directly — I verified a `ChunkedArray` exposes all three and `fill_null` returns a `ChunkedArray` that `to_numpy()` flattens correctly. So the guard is correct even though `__from_arrow__` already splits chunks before calling here.

Minor: `fill_null(int(MortonIndexArray._SENTINEL))` passes a plain Python `int` into a `uint64` array — fine here since the sentinel is 0, and pyarrow casts it. No change needed; flagging only because a non-zero sentinel would make the implicit int->uint64 cast worth a typed `pa.scalar`.

---

mortie/tests/test_arrow.py
```diff
@@ -115,6 +115,126 @@ def test_parquet_words_round_trip_to_pandas(self, tmp_path):
...
+    words[5] = SENTINEL
+    return words
+
+
+class TestNullRoundTrip:
+    def test_isna_words_emit_arrow_nulls(self):
+        # An isna() element (sentinel word) goes OUT as an Arrow null.
+        words = _words_with_gap()
+        arr = marrow.from_morton_index(words)
+        assert arr.null_count == 2
```
espg, line 195: 🤖 *from Claude (review)*

`SENTINEL = 0` here re-declares a value that already lives as `MortonIndexArray._SENTINEL` (an `np.uint64(0)`). The tests would read more robustly — and stay correct if the sentinel ever changes — by deriving it from the source of truth, e.g. `SENTINEL = int(mortie.MortonIndexArray._SENTINEL)`. As written, a future change to `_SENTINEL` would leave these tests silently asserting against a stale literal. Minor, but it's the kind of duplication §4 ("match the surrounding code, no speculative literals") warns against.

---

mortie/tests/test_arrow.py
```diff
@@ -115,6 +115,126 @@ def test_parquet_words_round_trip_to_pandas(self, tmp_path):
...
+        pq.write_table(pa.table({"mi": marrow.from_morton_index(words)}), path)
+        # The parquet column carries the null buffer through.
+        assert pq.read_table(path).column("mi").null_count == 2
+        series = pq.read_table(path).to_pandas()["mi"]
+        assert series.dtype.name == "morton_index"
+        np.testing.assert_array_equal(series.array._data, words)
+        np.testing.assert_array_equal(series.array.isna(), words == SENTINEL)
+
+
 # ---------------------------------------------------------------------------
```
espg, line 267: 🤖 *from Claude (review)*

Coverage is good on the mixed-null case (plain, chunked, parquet, both mask directions). Two edge cases the new `TestNullRoundTrip` class does **not** pin, both of which the new code now has live branches for:

1. **All-null array** — every word is the sentinel. Exercises the path where `mask` is all-True and `to_morton_index`'s `if storage.null_count:` fires for the whole array. (Verified working, but untested here.)
2. **No-null array** — confirms the new `mask`/`null_count == 0` branches leave a null-free round-trip byte-identical (i.e. the sentinel handling doesn't disturb arrays without gaps). `TestConversion` predates this commit and doesn't assert `null_count == 0`.

Also worth one assertion: an **empty** array straight through `from_morton_index`/`to_morton_index` (not just the empty *chunked* case in `test_from_arrow_handles_empty_chunked_array`) — to lock in that `from_morton_index(np.empty(0, uint64))` produces `null_count == 0` and `to_morton_index` returns an empty `MortonIndexArray` rather than tripping the `fill_null` branch. Each is a 3-4 line test mirroring `_words_with_gap`; recommend adding all three to fully close the acceptance criteria.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of commit 7d29097 (Arrow null<->sentinel round-trip, Option B on issue #79).

**The contract holds, both directions.** An `isna()`/sentinel element going OUT becomes an Arrow null (`mask = data == _SENTINEL`), and an Arrow null coming IN is filled with the sentinel before materializing (`fill_null` + `to_numpy`), so `MortonIndexArray.isna()` is True for it. I verified the pyarrow mechanics directly: bool-mask is accepted by `pa.array(..., mask=...)`; empty / all-null / no-null inputs all behave; and `to_morton_index` is robust on both a plain `pa.Array` and a `ChunkedArray` reaching it directly (both expose `null_count` / `fill_null` / `to_numpy`). The lazy-pyarrow / numpy-only-importability contract is preserved — no new top-level imports; the added `from .morton_index import MortonIndexArray` is inside the function bodies, matching the existing pattern.

**Findings (all non-blocking, see inline):**
- (1) `from_morton_index` docstring now advertises null handling generally, but only the pandas/raw-words input path is covered — passing an Arrow ext-array back in would not honor its null buffer (pre-existing; consider a one-line docstring scope note).
- (2) Test-local `SENTINEL = 0` duplicates `MortonIndexArray._SENTINEL`; derive it from the source of truth so the tests don't go stale if it ever changes.
- (3) Missing test coverage for three edge cases the new branches now have: **all-null** array, **no-null** array (confirms the mask path leaves null-free round-trips byte-identical), and an **empty plain** array through `from`/`to` (distinct from the empty-chunked case already tested). Each is ~3-4 lines.

No correctness defects found and no scope creep introduced. Recommend folding in (2) and (3) before clearing review; (1) is optional polish.

---

espg:
🤖 *from Claude*

Implemented phase (c): proper Arrow null handling, per the "Option B" directive (https://github.com/espg/mortie/pull/86#issuecomment-4845674053). This closes the open "Questions for review" item — the null<->sentinel contract now holds in both directions.

**The contract.** The all-zero empty word (`MortonIndexArray._SENTINEL`, what `isna()` treats as missing) maps to/from an Arrow null:
- **OUT** (`from_morton_index`, `mortie/arrow.py`): storage is built with a validity `mask = data == _SENTINEL`, so an `isna()`/sentinel element becomes an Arrow null. Driven by `MortonIndexArray.isna()` when given an array, or read off the raw `uint64` words directly.
- **IN** (`to_morton_index`, `mortie/arrow.py`): `if storage.null_count: storage = storage.fill_null(_SENTINEL)` before `to_numpy`, so an Arrow null lands back as the all-zero sentinel word and `isna()` reports it missing. `__from_arrow__` (plain `pa.Array`, `pa.ChunkedArray`, and the Parquet path) inherits this since it delegates to `to_morton_index`. The lazy pyarrow import is untouched; numpy stays the only runtime dep.

**Paths changed.** Only `from_morton_index` / `to_morton_index` in `mortie/arrow.py` (plus docstrings). `__from_arrow__` needed no change.

**Tests added** (`mortie/tests/test_arrow.py`, `TestNullRoundTrip`): sentinel words emit Arrow nulls (from raw words and from a `MortonIndexArray` via `isna()`); nulls come back as the sentinel with `isna()` True; `__from_arrow__` on null-bearing plain and chunked arrays; Parquet -> `read_table` -> `to_pandas()` preserves the null buffer + `isna()`; plus all-null, no-null (mask path leaves null-free arrays byte-identical), and empty-plain-array edges.

**Self-review fold.** The fresh-context review flagged: derive the test `SENTINEL` from `MortonIndexArray._SENTINEL` rather than a literal `0`; add the all-null / no-null / empty-plain edge tests; and a one-line docstring scope note on `from_morton_index`. All three folded in (commit `bf07a09`). The non-zero-sentinel `pa.scalar` nit was endorsed as "no change needed" (sentinel is 0), left as-is.

**How it was tested** (isolated venv on this branch):
- `flake8 mortie --select=E9,F63,F7,F82` -> clean; `--max-line-length=88` pass clean on touched files
- `pytest -q` -> **462 passed, 8 skipped** (`test_arrow.py` now 25 tests)

All three phases are complete; the ball is back with you for review.

---

Issue #78, opened 2026-06-27
on_minor_arc FP/SoS hardening


espg:
from #73 

(1) on_minor_arc FP sensitivity in robust_crossing — PR https://github.com/espg/mortie/pull/44 (hemisphere+ PIP)

The Phase-7 SoS hardening covers robust_crossing's four straddle gates but not its antipodal-disambiguation on_minor_arc, which still uses raw f64 signs. A probe point lying exactly on a polygon edge's great circle can flip on floating-point noise, silently producing incorrect PIP results with no error raised. PR https://github.com/espg/mortie/pull/44's author explicitly recommended opening a follow-up issue for this. Not a known active bug, but a latent correctness risk on edge-great-circle inputs.

espg:
🤖 *from Claude*

## Plan: SoS-harden `on_minor_arc` in `robust_crossing`

**Recommendation:** mirror the existing `orient_sos` SoS treatment already applied to `robust_crossing`'s four straddle gates onto the antipodal-disambiguation step, replacing the two raw `f64`-sign half-plane tests in `on_minor_arc` with an exact/SoS-consistent decision keyed to the same point identities the gates already carry. This is the residual sensitivity the code itself already flags as un-hardened.

### Where the gap is (grounded)

`robust_crossing` (`src_rust/src/sphere.rs:284-312`) hardens its four straddle determinants through `orient_sos` (`sphere.rs:295-301`), which decides on a canonical, identity-sorted evaluation and breaks exact-zero ties with Simulation of Simplicity (`sphere.rs:181-249`). But the antipodal disambiguation that follows is **not** hardened — it calls `on_minor_arc` (`sphere.rs:265-268`):

```rust
fn on_minor_arc(x: &Vec3, a: &Vec3, b: &Vec3) -> bool {
    let ab = cross(a, b);
    dot(&cross(a, x), &ab) >= 0.0 && dot(&cross(x, b), &ab) > 0.0
}
```

[...]

PR #44, opened 2026-06-15
Robust hemisphere+ point-in-polygon (S2-style orientation + SoS)

espg:
Refs #22. Closes #11 (the `Closes` keyword closes #11 only when this PR merges, per the [agreement to absorb #11 into #22](https://github.com/espg/mortie/issues/11#issuecomment-4700895124)).

## What this does

Replaces the spherical point-in-polygon core with a **single, any-size robust backend** and enforces the ring-orientation contract at ingest, fixing both the hemisphere+ case (#22) and the #11 cell-centre degeneracy.

1. **`point_in_ring_robust` / `parity_filled_robust`** — the one PIP path, computed from the **signed spherical winding number** (`ring_winding_at`, Bevis &amp; Chatelain 1989): `winding &gt; π` is the inside test. No projection centre to go singular, no sub-hemisphere reference — **correct at any polygon size, including hemisphere+**. It runs at the 12 base-cell seeds, off the descent hot path.

2. **`orient_sos` + `robust_crossing`** — the **f64 + Simulation-of-Simplicity** orientation primitives (Edelsbrunner &amp; Mücke 1990) that remove the #11 cell-centre degeneracy. **Now wired into the descent&#39;s per-cell fill walk** (`arc_crossing_parity`), which is what actually lands the #11 over-coverage flood at the area-fill level (not just point probes).

3. **Orientation normalization at ingest (now an opt-in flag).** `build_ring` reverses a CW **sub-hemisphere-vertex** ring to CCW; rings whose vertices span ≥ a hemisphere are never reordered. This is now controlled by a `normalize: bool` keyword on the public `morton_coverage` entry point (Phase 9) — default `True` (current behaviour).

Interior follows the **RFC 7946 §3.1.6 / S2 right-hand rule** (region to the left of the directed edges; CCW exterior, CW holes).

## Phases

- [x] **Phase 1** — robust orientation predicate (`orient_sos`) + robust crossing (`robust_crossing`) + any-size seed PIP; Rust tests incl. #11 + hemisphere+.
- [x] **Phase 2** — cap-cull complement guard (`covers_complement`).
- [x] **Housekeeping** — merged `main`; moved coverage tests; documented CCW contract.
- [x] **Phase 3** — single robust PIP + ingest orientation normalization.
- [x] **Phase 4** — Python tests vs `cdshealpix.polygon_search`.
[...]

Issue #11, opened 2025-11-20
Polygon's malformed when crossing the prime meridian


espg:
Potential issue-- we fixed this in #10 for cases where we touch the 180 / -180 anti-meridian. That fixes things for all of the polar cells, since the base cells lie along that meridian but don't cross it. However, there is one mid-latitude base cell that does straddle either side, so we may run into issues there...

espg:
## Over-coverage bug localized — related to this base-cell-meridian family, but broader than the prime-meridian case

While benchmarking **0.7.2** (freshly built and version-verified — note an earlier editable build was silently stale at 0.4.0) `morton_coverage` against `cdshealpix.polygon_search`, I found a gross **over-coverage** (commission) bug and traced it. It's in the same base-cell-meridian family this issue flags, but the prime meridian (lon 0) is *not* the trigger.

### Minimal reproducer
```python
import numpy as np, mortie   # 0.7.2
# 2°×2° box, left edge exactly ON lon = 45° (a polar base-cell-center meridian)
bad  = mortie.morton_coverage([40.,42.,42.,40.], [45.,45.,47.,47.], order=8)  # n = 34122
# same box shifted 1° east — edge off the meridian
good = mortie.morton_coverage([40.,42.,42.,40.], [46.,46.,48.,48.], order=8)  # n = 82
```
`bad` returns **34,122 cells ≈ 52% of base cell 0**; `good` returns the correct ~82. Verified two independent ways (cdshealpix cell centers *and* mortie's own `mort2geo`): of the 34k cells, **0 have centers inside the polygon** — they flood the half of base cell 0 on one side of the lon-45 meridian.

### What triggers it (empirical)
- **A polygon edge lying (nearly) on a base-cell boundary/center meridian** floods ~half that base cell. Sweep of a box's left-edge longitude (lat[40,42], order 8):

  | edge lon | 0 | 22 | 44 | **45** | 46 | 89 | **90** | **135** | **180** | **225** |
  |---|---|---|---|---|---|---|---|---|---|---|
  | cells | 99 | 77 | 78 | **34122** | 82 | 77 | **12321** | **34122** | **12321** | **32985** |

  Triggers on lon ∈ {45, 90, 135, 180, 225, 270, 315}; **lon 0 does *not* trigger** (n=99). Off by ±1° → always correct.
[...]

PR #87, opened 2026-06-27
SoS-harden on_minor_arc in robust_crossing

espg:
Refs #78

## What & why

`robust_crossing` (`src_rust/src/sphere.rs`) hardens its four straddle determinants through `orient_sos`, but its antipodal-disambiguation step `on_minor_arc` still used raw f64 signs with no identity tie-break:

```rust
fn on_minor_arc(x: &Vec3, a: &Vec3, b: &Vec3) -> bool {
    let ab = cross(a, b);
    dot(&cross(a, x), &ab) >= 0.0 && dot(&cross(x, b), &ab) > 0.0
}
```

In `robust_crossing` the probe `x` is always the AB×CD intersection, so it lies on AB's great circle by construction. The two wedge determinants are O(1) in the arc interior but vanish *exactly* when `x` coincides with an endpoint (`±a` / `±b`); there the raw f64 sign is noise and can flip on argument/edge reordering, silently producing a wrong PIP with no error raised (latent risk flagged in #78, follow-up from #73 / PR #44).

## Approach

Simulation of Simplicity, mirroring the existing `orient_sos` pattern (no exact/extended arithmetic):

- Each wedge sign `(u × v) · n` (with `n = a × b`) is decided on a **canonical, identity-sorted** evaluation with the sort parity reapplied — the same canonical-order trick the straddle gates use — so the same geometry rounds the same way in every argument order. An exact-`0.0` determinant (canonical-order tie) is broken by an identity-keyed symbolic perturbation (`sos_half_plane_sign`) that is **total** (never undecided) and **antisymmetric**, matching the `orient_sos` contract.
- The half-open `[a, b)` convention is preserved **without a tolerance**: endpoint coincidence is settled first via a tolerance-free `coincident` test (`a × b` exactly the zero vector and not antipodal) — start vertex inclusive, end vertex exclusive — exactly as the original `>= 0.0` / `> 0.0` bounds did.
- The derived intersection point `x` gets a reserved `INTERSECTION_ID` (distinct from every real vertex), so it perturbs consistently whether checked against arc AB or CD.
- On degeneracy the predicate **silently resolves** to a definite side (no error). A `debug_assert!` records the totality invariant (debug-level, off the hot path).

Per @espg's plan decisions: the `node_straddles` over-refinement at `coverage.rs:442` is left **untouched** (Phase 4 flag below); benchmarks are run-and-reported only (see PR comment), not committed.

## Phases

- [x] **Phase 1** — pinned tests driving `robust_crossing` with a probe exactly on a polygon edge's great circle (mid-arc and endpoint-coincident), asserting argument/edge-reorder invariance (`crossing_under_reorder`, now incl. the AB↔CD role swap; endpoint case pins a golden value).
- [x] **Phase 2** — gave `on_minor_arc` the point identities; routed each wedge through a canonical-order + SoS tie-break (`half_plane_sign` / `sos_half_plane_sign`); reserved `INTERSECTION_ID`; added the `debug_assert!` invariant; updated the now-accurate caveat at `coverage.rs` (it had said `on_minor_arc` was *not* SoS-hardened).
- [x] **Phase 3** — extended the suite with deterministic edge-on-great-circle fixtures: lon-45/lon-90 meridian-edge crossings, a 4000-case splitmix64 brute-force on a tilted great circle (every probe exactly on the edge's circle), a crossing-number PIP built **directly on `robust_crossing`** checked against the independent `winding_inside` trig oracle on the meridian + tilted boxes, and vertex-rotation invariance of that crossing-PIP.
- [x] **Phase 4** — flag-only (Questions for review below); no descent-behavior change.
- [x] **Phase 5** — split the `tests` module out of `sphere.rs` into the sibling file `src_rust/src/sphere/tests.rs` (declared via `#[cfg(test)] mod tests;`), mirroring the existing `coverage.rs` / `coverage/tests.rs` pattern, per @espg's review below. Pure code-move (no test-logic or production change): `sphere.rs` drops from ~1452 to **572** lines (under the §4 soft limit); the moved test body is byte-identical to the original block. Full `cargo test` stays green at **179** tests; `cargo fmt --check` and `cargo clippy --all-targets` clean on the moved code (pre-existing `coverage/tests.rs` warnings unchanged).

## How it was tested

All commands run locally green (own scratch venv, `maturin develop --release`):

- `cargo test` — full Rust suite green (**179** tests; includes the #11 meridian-box regression, the endpoint-pin, the 4000-case brute force, and the crossing-PIP-vs-oracle agreement).
- `cargo clippy --all-targets` — no new warnings on the changed code (`sphere.rs` / `sphere/tests.rs` / `coverage.rs` clean; remaining warnings are pre-existing and unrelated).
- `cargo fmt` — clean.
- `flake8 mortie --select=E9,F63,F7,F82` — clean; `pytest -q` — 441 passed, 8 skipped (no Python surface change).
- Numerical pre-validation (throwaway Python prototype): `half_plane_sign` antisymmetric + total over 25k coplanar/endpoint cases; the robust `on_minor_arc` matches the original on 30k clean random quadruples while making `robust_crossing` fully reorder-invariant; the half-open `[a, b)` boundary is preserved (x≡a IN, x≡b OUT) with zero violations.
- Benchmark before/after posted as a follow-up PR comment (run-and-report only; no bench code committed).

## Review resolutions

All three review items from the maintainer (https://github.com/espg/mortie/pull/87#issuecomment-4845761644) are now resolved:

1. **Module size (§4).** Resolved — the `tests` module is split out to `src_rust/src/sphere/tests.rs` (Phase 5 above), bringing `sphere.rs` to 572 lines, under the ~1000-line soft limit.
2. **SoS perturbation convention.** Confirmed fine as-is (perturb the canonical-lower-id point first along e₀/e₁/e₂, mirroring `sos_sorted_sign`); no change needed.
3. **`coverage.rs:442` (`node_straddles`) relaxation.** Tracked as a separate follow-up — opened #90 to evaluate loosening the containment now that disambiguation is exact, with the recommendation to **measure** the current over-refinement before changing any descent/containment behavior. Left untouched in this PR.

---

https://claude.ai/code/session_01QP5rg6JZ69FxVtjH5mrixz


espg (commented):
🤖 *from Claude (review)*

Adversarial review of **Phase 1 only** (commit d27cdb7, test-only additions to `src_rust/src/sphere.rs`). Both new tests compile and pass on the current (pre-fix) tree. The harness (`great_circle`, `crossing_under_reorder`) is clean and mirrors the existing `test_orient_sos_breaks_coplanar_consistently` style well. A few substantive findings on whether these tests actually pin the #78 bug:

**1. `test_robust_crossing_probe_on_edge_great_circle` does not exercise the `on_minor_arc` degeneracy — its docstring claim is geometrically wrong.**
The comment says the probe makes "both half-plane dot products round near zero." I instrumented the exact geometry: for the mid-arc probe, `on_minor_arc`'s two dots are `-7.11e-1` and `-7.77e-1` — large and well-conditioned, nowhere near zero. The reason: `x` lying on AB's great circle makes `cross(a,x)` *parallel* to `ab`, so `dot(cross(a,x), ab) = ±|cross(a,x)||ab|` is maximal, not minimal. `on_minor_arc`'s actual degeneracy is x at an **endpoint** (`cross(a,x)→0`) or x **antipodal**, not "x on AB's GC." So this test traverses the robust straight-line path of `on_minor_arc`; with raw f64 it would still be order-independent. It's a fine sanity/PIP check but is **not** a regression guard for the flip #78 describes — it would pass identically before and after Phase 2. Recommend fixing the comment to not overclaim, and/or repurposing it so the intersection genuinely lands in the near-zero regime.

**2. `test_robust_crossing_probe_through_edge_endpoint` *does* hit the degeneracy — but asserts the weaker property.** Instrumenting it: the computed intersection is `x ≈ -a` (antipodal to endpoint a), and `dot(cross(a,x), ab) = 3.9e-17` — genuinely on the `>= 0.0` boundary, exactly the flippable case. Good. But this test only asserts `res.len()==1` (order-independence) and does **not** pin the boolean. Test 1 (the non-degenerate one) pins the value; test 2 (the degenerate one) does not. That's backwards — the degenerate case is where a pinned golden value matters most. Suggest pinning the expected boolean here too (capture it from the post-fix Phase-2 path).

**3. Reorder coverage is incomplete for "edge-reorder invariance."** `crossing_under_reorder` swaps within each edge (a↔b, c↔d) with ids in lockstep — that part is correct. But it never swaps the **edge roles** (`robust_crossing(c,d,a,b, ic,id,ia,ib)`), which the issue's "argument/edge-reorder invariance" wording implies and which is the other symmetry `robust_crossing` should satisfy (AB-straddle vs CD-straddle gates evaluated in opposite roles). Worth adding the 2 role-swapped orderings (8 total) so the guard is complete.

**4. Caveat on the `x = -x` self-correction (for Phase 2).** `robust_crossing` flips `x→-x` when `!on_minor_arc(x,a,b)` and then requires `on_minor_arc(&x,a,b) && on_minor_arc(&x,c,d)`. For a *single* call this flip can mask single-evaluation instability, so a single-call assertion would be a weak guard — using `crossing_under_reorder` (multi-order set) is the right call to surface the inconsistency. Flagging so Phase 2 keeps the multi-order assertion rather than collapsing to a single-call golden.

**5. Module size (§4).** `sphere.rs` goes 1021 → 1136 lines at this commit; it was already past the ~1000-line soft limit before Phase 1. Test-only growth is defensible, but per §4 the Phase-2 fix should note this and confirm with @espg whether `sphere.rs` needs a split rather than continuing to grow silently.

Nits: `std::collections::BTreeSet` is std-only (fine, no dependency concern per §4). Helper docstrings are appropriately terse. No dead code.

Phase 1 is a reasonable scaffold and none of the above blocks landing it, but findings (1)–(3) mean the suite does not yet robustly pin the previously-flippable case — please address before/with Phase 2.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of **Phase 2 only** (commit `2bfe997`, touching `src_rust/src/sphere.rs` + `src_rust/src/coverage.rs`). Verdict: the SoS hardening looks correct and matches the approved plan. No blocking findings. Detailed pass against the acceptance criteria below.

**Verified correct**

1. **Half-open `[a, b)` survives, including the antipode flip** (`sphere.rs:350-366`). Traced against the original `>= 0.0` / `> 0.0` bounds:
   - `x ≡ a`: `coincident(x,b)` false; `coincident(x,a)` true → start satisfied; `s_end = half_plane_sign(a,b,n) = (a×b)·n = |n|² > 0` → **IN**. Original: `dot(cross(a,a),ab)=0 >= 0` and `dot(cross(a,b),ab)=|ab|²>0` → IN. Match.
   - `x ≡ b`: `coincident(x,b)` true → early `false` → **OUT**. Original: second term `dot(cross(b,b),ab)=0 > 0.0` is false → OUT. Match.
   - `x ≡ -b`: `coincident(-b,b)` false (`dot=-1`), so the `robust_crossing` flip (`x=-x`) and re-test behave as before; no special-casing was added or lost.

2. **`coincident` (`sphere.rs:373-375`) is sound and tolerance-free** — `cross(p,q)==[0,0,0] && dot(p,q)>0` captures exactly the same-direction case and (correctly) rejects the antipodal pair the `x=-x` flip relies on. It cannot miss a coincidence the old `>=`/`>` caught: at exact `x≡a`/`x≡b` the old comparisons hit their boundary value, which is exactly where `coincident` fires.

3. **`half_plane_sign` is total + antisymmetric** (`sphere.rs:278-291`), mirroring `orient_sos`. Canonical identity order + reapplied parity; `det=(p×q)·n` negates under swap while the SoS branch perturbs the lower-id point regardless of arg order, so `half_plane_sign(u,v)=-half_plane_sign(v,u)` exactly. `n=a×b` is intentionally *not* reordered with the points — the per-call wedge question changes when a↔b swaps (the interval reverses), and it is the final `robust_crossing` boolean that must be invariant; `crossing_under_reorder` (now covering the AB↔CD role swap too) tests exactly that.

4. **`sos_half_plane_sign` perturbation order is right** (`sphere.rs:301-321`). First loop `(eₖ×q)·n` perturbs the lower-id slot `p` (dominant), second loop `(p×eₖ)·n` perturbs `q` — lower-id-dominates, consistent with `sos_sorted_sign`. The final `+1` is genuinely unreachable in the descent: all six terms vanishing forces `n ∥ p` and `n ∥ q`, but `n=a×b ⊥ x` on the great circle, so `n=0` (degenerate `a∥b`) is the only way to reach it — a pure total-function guard, accurately documented.

5. **`INTERSECTION_ID = PointId::MAX` is collision-safe.** Vertex ids in `coverage.rs` start at `VERTEX_ID_BASE=2` and increment by ring length (`build_edges`, `coverage.rs:366-405`); reaching `u64::MAX` needs ~2⁶⁴ vertices. Probe ids are `0`/`1`. No real id can collide.

6. **`debug_assert!` is the right level** (`sphere.rs:361-364`) — the plan called for a debug-assert invariant, not a hot-path panic. Correct.

7. **Tests green locally:** `cargo test` → 178 passed / 0 failed, including the #11 regression (`test_robust_pip_issue_11_meridian_box`) and the new endpoint-coincident pin. `cargo clippy` surfaces only pre-existing `useless conversion: PyErr` warnings in unrelated PyO3 binding code — none from this diff. The two `#[allow(clippy::too_many_arguments)]` additions are warranted.

8. **No undiscussed dependency** — pure std/arithmetic, as the plan required. The `node_straddles` over-refinement at `coverage.rs:442` was correctly left untouched, and the caveat comment update accurately reflects the new state.

**Non-blocking findings**

- **(1) §4 soft module limit.** `sphere.rs` is now 1266 lines (was 461 on `main`; the prior count cited in the task predates these phases). This is past the ~1000-line soft limit and per §4 warrants an @espg heads-up before further growth. Mitigating: production (non-test) code ends at ~line 571 — well under 1000 — and the ~694 lines below `mod tests` (line 572) are test code that the issue plan explicitly directed adding here. So the reviewability concern §4 targets is largely not triggered, but the rule is by line count, so flagging it rather than silently passing. No action needed beyond a note for @espg.

- **(2) Minor doc nit (optional).** The `on_minor_arc` docstring says coincidence is "⇔ the cross product is exactly the zero vector and the points are not antipodal" — accurate, but the early-return only short-circuits the `b` case; the `a` case is folded into the `||` at the return. The code is correct; a one-word tweak ("settled at `b` up front, at `a` in the final test") would make the asymmetry self-evident. Purely cosmetic.

Nothing here blocks advancing the phase checklist.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of Phase 3 only (commit `5f619a9`, test-only additions to `src_rust/src/sphere.rs`). Net: the two `crossing_under_reorder` tests are solid #78 guards; the oracle test does not reach the changed code and should be re-labelled or dropped; and the plan's vertex-rotation invariance item is not added here.

**(1) Blocker for the stated goal — the oracle test does not guard `on_minor_arc`/`robust_crossing`.** `test_robust_pip_meridian_box_matches_oracle_on_edge` asserts `point_in_ring_robust == winding_inside`. But `point_in_ring_robust` is a pure **winding-number** test (`ring_winding_at` > π, line 434/493) — it never calls `robust_crossing` or `on_minor_arc`. Worse, the oracle `winding_inside` (line 1123) is a line-for-line copy of `ring_winding_at` (line 434): same projection, same `acos`-clamp, same `dot(cross(pa,pb),x)` sign, same `> π`. So the assertion is `winding(p) > π == winding(p) > π` modulo a duplicated implementation — a near-tautology that exercises neither the SoS code nor the #78 fix. The prompt's framing ("a DIFFERENT/independent computation") is not satisfied: it is the **same** computation. This is at best a regression guard for the winding-PIP on the #11 meridian family, not an `on_minor_arc` test. Recommend: either re-comment it honestly as a winding-PIP/#11 regression guard (and stop implying it covers #78), or replace `winding_inside` with a genuinely independent oracle (e.g. an even-odd crossing count built on `robust_crossing`) so the agreement is meaningful. As written it gives false confidence.

**(2) The two `crossing_under_reorder` tests are the real guards, and they are correct.**
- `test_robust_crossing_meridian_edge_on_circle`: AB is on the lon-45/90 meridian (both endpoints at that lon), and CD has both endpoints at the same latitude symmetric in lon (`lon±8`), so CD is mirror-symmetric across the meridian plane and its intersection with AB lands exactly on the meridian's great circle — the on-arc degeneracy. Good construction; the lon-45/90 coordinates tie it to the #11 family as the plan asked.
- `test_robust_crossing_tilted_circle_brute_force_invariance`: 4000 splitmix64 cases on a tilted GC, each probe forced exactly onto AB's GC and the `orient(a,b,xm).abs() < 1e-12` gate confirms it before the c/d offset is applied along the plane normal. Meaningful gate, deterministic, no `rand` dep (good). `crossing_under_reorder` size 1 is the right invariant.

**(3) Overlap with existing tests.** The tilted-GC + `nrm`-offset probe construction duplicates the pre-existing `test_robust_crossing_probe_on_edge_great_circle` (line 882); the new test generalizes it to 4000 randomized cases, so it is additive rather than redundant — fine, but worth a one-line note in the new test pointing at the single-case sibling.

**(4) Plan item "vertex-rotation invariance" appears unmet.** The plan called for "brute-force argument/**vertex-rotation** invariance." Both new crossing tests vary only edge/probe **argument** order (endpoint reversal + AB↔CD swap via `crossing_under_reorder`). Ring-start (vertex) rotation — the property `test_robust_pip_consistency_under_vertex_rotation` (line 1249) checks — is not exercised against the on-arc `robust_crossing` path. If vertex-rotation coverage on the degenerate on-arc geometry was intended, it is missing; if it was deemed covered by the existing PIP test, say so explicitly.

**(5) `rotate_about` (Rodrigues) is correct, but only used by the tautological test (1).** Formula `v cos θ + (k×v) sin θ + k (k·v)(1−cos θ)` is right for unit `k`. The tilted square (40–50°N, −125..−115°E rotated 0.9 rad about a fixed axis) is a proper non-collinear quad and rotation is an isometry, so the ring cannot collapse — fine. But since its only consumer is the test flagged in (1), if that test is reworked to a real independent oracle, re-confirm `rotate_about` is still needed (it should be) so it does not become dead code (§4).

**(6) Determinism / float fragility.** Seeds are fixed and splitmix64 is integer-deterministic, so the random draws are reproducible. The `1e-12` gate on `orient` and the `< π` winding threshold are coarse relative to the configurations chosen, so cross-platform fragility looks low. No concern, just noting the 4000-case loop is pure CPU (no I/O) so it stays cheap.

**(7) Module size (§4).** `sphere.rs` is now ~1409 lines, past the ~1000-line soft limit. It is all test code below line 571, but the rule says raise it before continuing to grow the file — flagging for @espg to confirm the test module is exempt or wants splitting into a `tests/` sibling. No dependency added (good).

Nothing here is irreversible and I am not requesting changes; (1) and (4) are the items that most affect whether Phase 3 actually demonstrates the #78 fix.

---

espg:
🤖 *from Claude*

**Benchmark: before/after impact of the `on_minor_arc` SoS hardening.**

Measured with a throwaway criterion bench (run-and-report only; not committed per the plan), built `--release`, same machine, back-to-back. "before" = this branch's `src_rust/src/{sphere,coverage}.rs` checked out at `origin/main`; "after" = the fix. Numbers are criterion's median estimate [low mid high].

**Micro: `robust_crossing` over 2000 quadruples per iteration**

| case | before (main) | after (fix) | Δ median |
|---|---|---|---|
| clean (generic-position crossings) | 55.19 µs | 64.07 µs | +16% |
| on_great_circle (probe exactly on edge's GC) | 67.80 µs | 110.69 µs | +63% |

The per-call cost rises because every `on_minor_arc` wedge now runs the canonical-id-sort + `coincident` check and, in the degenerate branch, the SoS perturbation. This is the function whose body changed, so the micro-bench shows the worst case.

**End-to-end: `polygon_to_morton_coverage` on the #11 lon-45 meridian box** (the degenerate edge-on-meridian descent that routes through `robust_crossing`)

| order | before (main) | after (fix) | Δ median |
|---|---|---|---|
| 6 | 127.97 µs | 127.78 µs | ~0% |
| 8 | 220.40 µs | 222.08 µs | +0.8% |
| 10 | 509.10 µs | 519.31 µs | +2% |

**Takeaway.** The end-to-end coverage cost is essentially unchanged (within noise) even on the meridian-box case that maximally exercises the degenerate path. That is expected: `edge_crosses_probe` only falls back to `robust_crossing` when a straddle determinant is within `ORIENT_EPS` of zero (rare); the bulk descent stays on the `arcs_cross_n` fast path, which this PR does not touch. The micro-bench overhead is real but confined to the already-rare robust fallback, so it does not move the descent's wall time.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of fae2844 ("fold phase 3 self-review on #78") only — the `sphere.rs` test-module change that replaces the tautological `test_robust_pip_meridian_box_matches_oracle_on_edge` with a `crossing_pip` helper plus two tests. Verdict: this directly fixes the gap the prior self-review flagged, and the construction is sound. No blocking findings. A few notes.

**The core concern is resolved.** `crossing_pip` calls `robust_crossing` per edge, which calls `on_minor_arc` — so unlike `point_in_ring_robust` (winding number, never touches the #78 code path), this test genuinely exercises the SoS-hardened crossing predicate and compares it against the *independent* `winding_inside` trig oracle. That is now a real guard for #78, not a same-algorithm tautology.

**Id assignment is sound.** ref id 0, probe id 1, vertex `i → i+2` are all distinct, and `INTERSECTION_ID = u64::MAX` (sphere.rs:165) is far above any test vertex id, so `robust_crossing`'s internal intersection id can't collide with the ring/probe/ref ids. Good.

**Reference-outside / parity-vs-winding validity checks out.**
- South pole `(-89,0)` is unambiguously outside the lat-40-42 boxes, so crossing parity (odd=inside) is anchored correctly.
- For the tilted box, `r_t = rot(south_pole)` and `tilted = rot(square)`: rotation is an isometry, so `r_t` is outside `tilted` iff south pole is outside `square` (lat ~45-50) — which it is. So rotation cannot move the reference inside. Correctly reasoned in the new comment.

**One unstated invariant worth a comment (low severity, non-blocking).** The parity-vs-winding agreement only holds because every test ring is wound **CCW**. Crossing-parity is orientation-agnostic (odd crossings = inside regardless of winding direction), but `winding_inside` uses `>π`, which is true *only* for the CCW interior — a CW ring of the same geometry winds `≈ -2π` and the two would disagree. `ring()` does no orientation normalization (sphere.rs:575), so this rests on the literal vertex order of `box45`/`box90`/`square` being CCW. They evidently are (suite is green), but a one-line note on `crossing_pip` ("oracle agreement assumes CCW-wound rings") would make the precondition explicit and protect against a future test author pasting a CW ring and getting a confusing failure. Optional.

**Vertex-rotation test is genuine.** `base` and `rotated` are the same 4 vertices cyclically rotated, same orientation, same reference; probes straddle the lon-45/lon-47 edges and the edge latitudes (incl. 41.81, the base-cell-centre lat). This actually exercises traversal-order independence of the crossing path on the meridian-box degeneracy — the property the prior review noted was missing. Good.

**Conventions.** No new dependency. Terse, matches surrounding test style, dead `(0.0,0.0)` antipodal-ish probe and the redundant `dlon=0.0` column dropped cleanly. Module size: `sphere.rs` is now ~1452 lines, past the ~1000 soft limit (§4) — already flagged in PR Question (3) with an offer to split a `sphere/tests.rs`, and it is test code, so no further action needed from this review beyond confirming the flag stands for @espg.

Net: the commit does what it claims and closes the tautology gap for #78. Only the CCW-precondition comment is suggested, and it's optional.

---

espg:
🤖 *from Claude*

**Status: all four phases complete, CI green, ready for review.** Marked `waiting` pending your answers to the three Questions for review in the PR description.

Run summary:
- Phase 1 (pinned reorder-invariance tests) → Phase 2 (the SoS hardening of `on_minor_arc` + `half_plane_sign`/`sos_half_plane_sign`, `INTERSECTION_ID`, `debug_assert!`, caveat-comment update) → Phase 3 (meridian/tilted edge-on-great-circle fixtures + a crossing-number PIP built on `robust_crossing` checked against the `winding_inside` oracle + vertex-rotation invariance) → Phase 4 (flag-only).
- A fresh-context adversarial self-review ran after each phase push. Folded findings: added the AB↔CD edge-role swap and an endpoint golden-value pin to the reorder test; replaced a self-review-flagged tautological winding-vs-winding test with a genuine `robust_crossing`-based crossing-PIP-vs-oracle test plus vertex-rotation invariance; documented the CCW/reference-outside precondition for that oracle agreement.
- Local gates green: `cargo test` 179 passed, `cargo clippy`/`cargo fmt` clean on the changed code, `flake8 mortie --select=E9,F63,F7,F82` clean, `pytest` 441 passed / 8 skipped. CI on the PR: `ruff`, `test (3.10/3.11/3.12)`, all wheel builds, and codecov all green.
- Benchmark before/after posted above (run-and-report only; no bench code committed).

Open decisions are the three Questions for review — the `node_straddles`/`coverage.rs:442` relaxation (left untouched per plan; recommend a separate follow-up issue), the SoS perturbation convention, and the `sphere.rs` ~1452-line module-size flag (happy to split a `sphere/tests.rs` sibling if you'd prefer). Left as a draft until you weigh in.

---

espg:
> sphere.rs module size 

yes, split the tests out please.

> SoS perturbation convention

Confirmed, fine as is

> node_straddles/coverage.rs:442 relaxation (left untouched per plan; recommend a separate follow-up issue)

sure, open a follow up issue

espg (commented):
🤖 *from Claude (review)*

**Verdict: LGTM.** Clean, faithful test-only code-move refactor. All acceptance criteria verified against the conventions; nothing blocking, nothing to fix.

Evidence (each finding led with the prefix):

🤖 *from Claude (review)* — **Move is faithful, no tests dropped.** The 25 `#[test]` fn names in `HEAD~1:src_rust/src/sphere.rs` are identical to the set now in `src_rust/src/sphere/tests.rs` (diff of sorted `#[test]` fn names is empty). Whitespace-insensitive comparison of the moved body vs the old `mod tests { ... }` body is byte-identical (24765 chars each) — the only delta is the dedent from being un-nested out of the `mod` wrapper.

🤖 *from Claude (review)* — **No production code changed.** `git diff HEAD~1 -- src_rust/src/sphere.rs` reduces to exactly one substantive change: the `mod tests { ... }` block collapsing to `mod tests;` (single added line `+mod tests;`). No non-test lines touched.

🤖 *from Claude (review)* — **`sphere.rs` is now under the soft limit.** 572 lines (down from ~1459). `sphere/tests.rs` is 889 lines. Both under the §4 ~1000-line soft limit.

🤖 *from Claude (review)* — **Wiring mirrors the coverage idiom exactly.** Declaration is `#[cfg(test)]` newline `mod tests;` under the same `// ── tests ──` divider, matching `coverage.rs`. The new sibling file's header doc-comment (`//! Unit tests for the spherical-geometry primitives (issue #78).` … "Split out of `sphere.rs` to keep that module under the project's ~1000-line soft limit; wired back in via `#[cfg(test)] mod tests;`.") parallels `coverage/tests.rs` line-for-line in form, with `use super::*;` following. Identical idiom.

🤖 *from Claude (review)* — **Test count preserved; build green.** `cargo test` → `179 passed; 0 failed`. `cargo fmt --check` clean (exit 0). `cargo clippy --all-targets` → 38 warnings, all pre-existing (lib.rs 30, coverage/tests.rs 5, geo2mort.rs 1); **zero** reference sphere.rs or sphere/tests.rs, so no new warnings introduced by the moved code.

🤖 *from Claude (review)* — **Commit style compliant.** Message is title-only (`split sphere tests on #78`), matching repo style; no body, no Claude credit / Co-Authored-By. Touches only the two expected files (+890/-886).

No issues found. Ready to advance.

---

espg:
🤖 *from Claude*

All three review items from https://github.com/espg/mortie/pull/87#issuecomment-4845761644 are now resolved; PR awaits review.

**1. Module size — tests split out.** Moved the `#[cfg(test)] mod tests { ... }` body out of `sphere.rs` into the sibling file `src_rust/src/sphere/tests.rs`, declared via `#[cfg(test)] mod tests;`, mirroring the existing `coverage.rs` / `coverage/tests.rs` pattern:

```rust
// ── tests ────────────────────────────────────────────────────────────────

#[cfg(test)]
mod tests;
```

Pure code-move — no test-logic or production change. `sphere.rs` drops from ~1452 to **572** lines (under the §4 ~1000 soft limit); the 25 `#[test]` functions are unchanged and the moved body is byte-identical to the original block (verified by reconstructing the original and diffing — no difference).

**Cargo gates (own scratch venv):**
- `cargo test` — **179** passed / 0 failed (same count as before the split).
- `cargo fmt --check` — clean.
- `cargo clippy --all-targets` — **zero** warnings on the moved `sphere` code; the remaining warnings are pre-existing and unrelated (`lib.rs`, `coverage/tests.rs`, `geo2mort.rs`).
- `maturin develop --release` + `pytest -q` — 441 passed, 8 skipped (no Python surface change).

A fresh-context adversarial self-review of the split commit returned LGTM with no diff-scoped findings, so no follow-up commit was needed.

**2. SoS perturbation convention.** Confirmed fine as-is — no change.

**3. `node_straddles` / `coverage.rs:442` relaxation.** Opened follow-up #90 to evaluate loosening the containment now that `on_minor_arc` is exactly SoS-hardened, with the recommendation to **measure** the current over-refinement before changing any descent/containment behavior. It's a tracking follow-up left unlabelled for triage; this PR leaves the descent untouched.

---

PR #91, opened 2026-06-30
Tag-driven Cargo.toml version sync

### LINKED ISSUE #67: v1.0: packaging & extras audit (pyproject extras + wheel matrix)
🤖 *from Claude*

Sub-issue of #48 (v1.0 sweep), Tier-2.

## Scope
Confirm the build/packaging surface is correct and minimal before the 1.x contract:

- Verify the runtime dependency set stays **numpy-only**; that pandas/pyarrow/cdshealpix and the notebook stack live only in the `pandas`/`pyarrow`/`test`/`bench`/`docs` extras (`pyproject.toml`).
- Audit the `build-wheels.yml` matrix (abi3-py310; the 3.10/3.11/3.12 + linux/macos/windows/arm coverage) and the import smoke test on each wheel.
- Confirm the maturin-only build path (no setuptools resurrection; stale `setup.py`/`setup.cfg`/`*~` removed) and that the dynamic, tag-driven version is single-sourced from `Cargo.toml`.

## Acceptance criteria
A short audit note + any minimal `pyproject.toml`/workflow fixes; no change to the tag-driven publish pipeline without explicit sign-off.

Refs #48.

espg:
Closes #67.

## What

Makes the `Cargo.toml` version **tag-driven** instead of hand-bumped, so cutting a release is just pushing a `*.*.*` tag — no more manual `version =` edit, and no more "tag != Cargo.toml" release failures (the failure that blocked the `0.8.3` run).

## Approach

Mirrors the existing changelog-update job rather than asserting the version:

- **`update-changelog` job (runs on the tag, pushes to `main`):**
  - `Get tag name` now also emits `version=${TAG#v}` (Cargo.toml's line is bare semver, no leading `v`).
  - New **Sync Cargo.toml version to tag** step rewrites the `version =` line from the tag.
  - The commit step now `git add CHANGELOG.md Cargo.toml` and pushes both back to `main` — so the committed `Cargo.toml` tracks the released version, exactly as the changelog already does.
- **Three build jobs (wheels ×2 + sdist):** the old *Verify release tag matches Cargo.toml version* fail-steps are **removed** and replaced with an **ephemeral** same-sed sync. These jobs `checkout` the *tag* ref, which predates the changelog job's bump commit, so they re-apply the bump in-place (not committed) to guarantee the wheel/sdist version matches regardless of checkout timing — belt-and-suspenders with the `main` commit.

Plain `sed`, no new tooling (per the issue's greenlit choices: plain sed, commit `Cargo.toml` back to `main`).

**One deviation from the diff in the issue, flagged for review:** the issue's diff used `sed -E -i`. GitHub's macOS runners use **BSD sed**, where `-i` consumes the next arg as a backup suffix and the in-place edit breaks — this would fail the `macos-15-intel` / `macos-latest` wheel jobs at the sync step. I used the file's existing portable idiom (`mktemp` + redirect + `mv`, same as the changelog job at lines 133/149) so it works on both GNU and BSD sed.

## Phases

- [x] Phase 1 — version-sync in `build-wheels.yml` (changelog-job commit + 3 ephemeral build-job syncs; verify-and-fail steps removed)

## Testing

- `python -c "import yaml; yaml.safe_load(...)"` → parses clean.
- Grep audit: 0 `Verify release tag` steps remain, 4 `Sync Cargo.toml version to tag` steps, 0 `sed -E -i`.
- `sed` rewrite exercised against both `v1.2.3` → `1.2.3` and bare `2.0.0` → `2.0.0` (BSD sed, no `-i`), correct in both.

End-to-end can only be verified by an actual tag push (production publish), which I won't do.

## Questions for review

- Confirm you're OK with the **portability deviation** (`mktemp`+`mv` instead of `sed -i`) — it's the one place this differs from the diff you saw in the issue.

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -181,6 +181,13 @@ jobs:
         tmp=$(mktemp)
         sed -E 's/^version = ".*"/version = "'"$VERSION"'"/' Cargo.toml > "$tmp"
         mv "$tmp" Cargo.toml
+        # Fail loudly if the version line never matched (sed no-op), so a stale
+        # version can't silently ship to PyPI — restores the fail-fast that the
+        # removed "Verify release tag" step provided.
+        grep -qF "version = \"$VERSION\"" Cargo.toml || {
+          echo "::error::Cargo.toml version sync did not apply (expected version = \"$VERSION\")."
+          exit 1
+        }
         echo "Cargo.toml version set to $VERSION"
         grep -m1 '^version = ' Cargo.toml
 
@@ -240,6 +247,13 @@ jobs:
         tmp=$(mktemp)
         sed -E 's/^version = ".*"/version = "'"$VERSION"'"/' Cargo.toml > "$tmp"
         mv "$tmp" Cargo.toml
+        # Fail loudly if the version line never matched (sed no-op), so a stale
+        # version can't silently ship to PyPI — restores the fail-fast that the
+        # removed "Verify release tag" step provided.
+        grep -qF "version = \"$VERSION\"" Cargo.toml || {
+          echo "::error::Cargo.toml version sync did not apply (expected version = \"$VERSION\")."
+          exit 1
+        }
         echo "Cargo.toml version set to $VERSION"
 
     - name: Build wheels
@@ -283,6 +297,13 @@ jobs:
         tmp=$(mktemp)
         sed -E 's/^version = ".*"/version = "'"$VERSION"'"/' Cargo.toml > "$tmp"
         mv "$tmp" Cargo.toml
+        # Fail loudly if the version line never matched (sed no-op), so a stale
+        # version can't silently ship to PyPI — restores the fail-fast that the
+        # removed "Verify release tag" step provided.
+        grep -qF "version = \"$VERSION\"" Cargo.toml || {
+          echo "::error::Cargo.toml version sync did not apply (expected version = \"$VERSION\")."
+          exit 1
+        }
         echo "Cargo.toml version set to $VERSION"
 
     - name: Build wheels
@@ -329,6 +350,13 @@ jobs:
         tmp=$(mktemp)
         sed -E 's/^version = ".*"/version = "'"$VERSION"'"/' Cargo.toml > "$tmp"
         mv "$tmp" Cargo.toml
+        # Fail loudly if the version line never matched (sed no-op), so a stale
+        # version can't silently ship to PyPI — restores the fail-fast that the
+        # removed "Verify release tag" step provided.
+        grep -qF "version = \"$VERSION\"" Cargo.toml || {
+          echo "::error::Cargo.toml version sync did not apply (expected version = \"$VERSION\")."
+          exit 1
+        }
         echo "Cargo.toml version set to $VERSION"
 
     - name: Build sdist
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -28,7 +28,10 @@ jobs:
     - name: Get tag name
       id: get_tag
       run: |
-        echo "tag=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
+        TAG=${GITHUB_REF#refs/tags/}
+        echo "tag=$TAG" >> $GITHUB_OUTPUT
+        # Cargo.toml's version line is bare semver (no leading "v").
+        echo "version=${TAG#v}" >> $GITHUB_OUTPUT
 
     - name: Get previous tag
       id: get_previous_tag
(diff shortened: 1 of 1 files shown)


.github/workflows/build-wheels.yml
```diff
@@ -213,17 +228,19 @@ jobs:
...
-          exit 1
-        fi
-        echo "Release tag matches Cargo.toml version: $TAG"
+        # Build jobs check out the tag ref, which predates the changelog job's
+        # version-bump commit; re-apply the same sed so the wheel version matches.
+        VERSION=${GITHUB_REF#refs/tags/}
+        VERSION=${VERSION#v}
+        # mktemp + mv (not sed -i) so this works on both GNU and BSD/macOS sed.
+        tmp=$(mktemp)
+        sed -E 's/^version = ".*"/version = "'"$VERSION"'"/' Cargo.toml > "$tmp"
```
espg, line 248: 🤖 *from Claude (review)*

Removing the *Verify release tag matches Cargo.toml version* step deleted the only fail-fast guard standing between the build and the **irreversible** PyPI publish, and this sed is never checked for success. `sed -E 's/^version = ".*"/.../'` silently **no-ops** (exit 0, file unchanged) if the `version =` line ever stops matching the anchored pattern — e.g. a future `version.workspace = true`, a reformat, an inline comment, or any whitespace drift. In that case Cargo.toml stays at the stale committed value (`0.8.2`), maturin builds that version, and `publish-testpypi`/`publish` ship the wrong number to PyPI, where a version can never be re-uploaded. None of the four sync steps assert the outcome (the changelog-job step at line 185 only `grep`s for display; the three build steps don't even do that).

A cheap fix restores fail-fast without bringing back the manual bump — assert the result right after the `mv`, e.g.:

```bash
grep -qF "version = \"$VERSION\"" Cargo.toml || { echo "::error::Cargo.toml version sync no-op (pattern did not match)"; exit 1; }
```

This applies to all four sync steps (this one at 241, plus 182, 284, 330).

.github/workflows/build-wheels.yml
```diff
@@ -170,25 +173,37 @@ jobs:
...
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
         git config --local user.name "github-actions[bot]"
         git config --local user.email "github-actions[bot]@users.noreply.github.com"
 
-        # Switch to main branch and pull latest changes
+        # Switch to main branch and pull latest changes (uncommitted
+        # CHANGELOG.md/Cargo.toml edits carry over since they share main's base).
         git checkout main
```
espg, line 203: 🤖 *from Claude (review)*

The "Sync Cargo.toml version" step runs on the **detached tag checkout**, then this `git checkout main` relies on git carrying the uncommitted Cargo.toml edit across the branch switch. That carry succeeds only when the tag commit's committed `version =` equals main's committed `version =` (true for a tag cut from main's tip, the normal path). For a tag pushed from a commit whose Cargo.toml version differs from current main — a hotfix tag off a side branch, or re-tagging an older commit — `git checkout main` aborts with *"Your local changes to Cargo.toml would be overwritten by checkout"* and fails this job.

That failure is silently swallowed downstream: the build jobs gate on `needs.update-changelog.result == 'success' || 'skipped'` (lines 212/262/306), so a *failed* update-changelog skips every wheel/sdist/publish job — the release produces nothing and nobody is alerted.

Robust fix: move the Cargo.toml sed to run **after** `git checkout main && git pull origin main` (operate on main's own tree) instead of editing the tag checkout and carrying it over.

.github/workflows/build-wheels.yml
espg: 🤖 *from Claude*

Folded into phase 2 (`676e3de`). After each of the four `mv "$tmp" Cargo.toml`, added an anchored assertion:

```bash
grep -qF "version = \"$VERSION\"" Cargo.toml || {
  echo "::error::Cargo.toml version sync did not apply (expected version = \"$VERSION\")."
  exit 1
}
```

This restores the fail-fast the removed verify step provided: if the `version =` line ever stops matching the `^version` pattern (sed no-ops), the job now errors instead of shipping the stale committed version to PyPI. Verified locally — a no-op (stale `0.8.2` vs tag `0.8.3`) fails the assertion; a real bump passes.

.github/workflows/build-wheels.yml
espg: 🤖 *from Claude*

Real, but leaving this one **standing for review rather than folding it**, because the clean fix expands scope beyond #67.

The dominant blocker on the edge case you describe (tagging a non-tip / old commit) is the **pre-existing** `CHANGELOG.md` carry-over: the "Update CHANGELOG.md" step already edits `CHANGELOG.md` in the tag checkout *before* `git checkout main`, so on a non-tip tag that checkout aborts on `CHANGELOG.md` first — before `Cargo.toml` is even reached. This PR added `Cargo.toml` to that same carry-over set but did not introduce the pattern. So moving only the `Cargo.toml` sed below `checkout main && pull` would not actually rescue the old-tag case (the changelog edit still aborts the checkout), and fully fixing it means restructuring the pre-existing changelog-commit flow — outside this issue.

Normal tip-of-`main` releases (the documented flow) are unaffected: tag and `main` share the same `Cargo.toml`, so there is no carry-over conflict.

Proposed handling: track the carry-over hardening as a separate small-fix issue. Flagging for a decision before touching the changelog step — happy to do it here instead if preferred.

espg: 🤖 *from Claude*

Tracked as #92 (`small-fix`) — keeping #91 scoped to the version-sync. This thread can stay as the reference.

PR #95, opened 2026-07-01
Harden release changelog/version commit against non-tip tags

### LINKED ISSUE #92: Harden release changelog/version commit against non-tip tags
🤖 *from Claude*

Follow-up from the adversarial review of #91 (version-sync). Tracking a low-severity edge case so #91 stays scoped to the version-sync itself.

## Problem

In `build-wheels.yml`, the `update-changelog` job edits tracked files **in the tag-ref checkout** and then switches branches to commit them:

- `Update CHANGELOG.md` rewrites `CHANGELOG.md` in place (the detached tag checkout).
- `Sync Cargo.toml version to tag` rewrites `Cargo.toml` in place.
- `Commit changelog and version updates` then runs `git checkout main` / `git pull origin main` and relies on the **uncommitted edits carrying over** to `main` before `git add` + commit + push.

That carry-over only works when the tag's committed copy of those files equals `main`'s copy. For the normal flow — tagging the **tip of `main`** — that holds, so releases are fine. But when a `*.*.*` tag is pushed on a **non-tip / older commit** (a hotfix, or re-tagging an old commit), `CHANGELOG.md` (and now `Cargo.toml`) differ between the tag and `main`, so `git checkout main` aborts with *"Your local changes ... would be overwritten by checkout."* Because the build jobs gate on `needs: [update-changelog]` being `success`/`skipped`, that failure can **silently skip the whole release** rather than failing loudly.

Note this is a **pre-existing** pattern for `CHANGELOG.md`; #91 only added `Cargo.toml` to the same carry-over set. The `CHANGELOG.md` edit aborts the checkout first, so it must be fixed together.

## Fix

Restructure the commit step so the file edits happen **after** `git checkout main && git pull origin main` (apply the changelog + version edits on `main` directly), rather than editing in the tag checkout and carrying across. The build jobs already re-derive the version from the tag via their own ephemeral sed, so the committed-to-`main` edits only need to be correct on `main`.

## Acceptance

- Tagging a non-tip commit completes the `update-changelog` job (no aborted `git checkout`).
- Normal tip-of-`main` releases unchanged.
- No manual version bump reintroduced.

Refs #67. Surfaced in https://github.com/espg/mortie/pull/91#discussion_r3501074100

espg:
Closes #92.

## What this does

Two CI-config fixes, both explicitly requested by @espg (issue #92 by name, and the one-line `codecov.yml` change referenced in the PR #94 comment permalink).

### 1. Harden the release changelog/version commit against non-tip tags (#92)

`build-wheels.yml`'s `update-changelog` job edited tracked files **in the detached tag checkout** and then switched to `main` to commit them, relying on the uncommitted edits carrying over:

- `Update CHANGELOG.md` rewrote `CHANGELOG.md` in place.
- `Sync Cargo.toml version to tag` rewrote `Cargo.toml` in place.
- `Commit changelog and version updates` then ran `git checkout main` / `git pull origin main` and committed the carried-over edits.

That carry-over only works when the tag's committed copies of those files equal `main`'s. For the normal flow (tagging the **tip of `main`**) that holds. But tagging a **non-tip / older commit** (a hotfix, or re-tagging) makes `CHANGELOG.md`/`Cargo.toml` differ, so `git checkout main` aborts with *"Your local changes … would be overwritten by checkout."* Because the build jobs gate on `needs: [update-changelog]` being `success`/`skipped`, that failure can **silently skip the whole release**.

**Fix (per the issue's suggested approach):** switch to `main` and `git pull` *before* the file edits, so the changelog + version edits are applied on `main` directly rather than edited in the tag checkout and carried across. Concretely:

- Added a `Switch to main before editing tracked files` step immediately after `Generate changelog entry` (which writes to a `mktemp` temp file, independent of the working tree) and before `Update CHANGELOG.md`.
- Removed the now-redundant `git checkout main` / `git pull origin main` from `Commit changelog and version updates`.

Changelog generation is unaffected by the current branch: `Get tag name`, `Get previous tag`, and `Generate changelog entry` reference tags **by name** (`git log … $CURRENT_TAG`, `git describe --tags … ${tag}^`), which resolve from any branch (the job checks out with `fetch-depth: 0`). The per-job ephemeral `sed` in each build job still re-derives the wheel/sdist version from the tag independently, so the committed-to-`main` edits only need to be correct on `main`.

### 2. Make `codecov.yml`'s test-file ignore actually exclude patch coverage (folded in)

The referenced comment on PR #94 (`pull/94#issuecomment-4859353255`) noted that `ignore: ["mortie/tests/**/*", …]` fails to exclude files sitting *directly* under `mortie/tests/` from patch coverage, which drags the codecov/patch check down for intentionally-CI-skipped test files. Changed `"mortie/tests/**/*"` → `"mortie/tests"` so the directory (and everything under it) is excluded, making the existing intent work.

## How it was tested

- Both YAML files validated with `yaml.safe_load` (parse clean).
- This is a CI-config-only change with **no Rust/Python source touched**, so `maturin develop` / `pytest` / `flake8` have nothing to exercise here; the release-workflow path can only be exercised by pushing a `*.*.*` tag, which is forbidden (§1, §7). The change is a mechanical reordering of existing, already-working steps plus one glob edit — reviewed by tracing the step order and git semantics rather than by execution.

## Questions for review

- **`codecov.yml` glob form:** the comment offered `"mortie/tests"` **or** adding `"mortie/tests/*"`. I used `"mortie/tests"` (directory form) since it's the single-line change the comment led with and matches the sibling `setup.py` / `setup.cfg` path-style entries. Happy to switch to `"mortie/tests/*"` if you prefer the glob form.
- The `Switch to main` step assumes `main` is fetchable at that point (it is — the job checks out `fetch-depth: 0` with the default token, same as the original commit step's `git pull origin main`).

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -127,6 +127,18 @@ jobs:
         # Store the new entry file path
         echo "new_entry_file=$NEW_ENTRY_FILE" >> $GITHUB_OUTPUT
 
+    - name: Switch to main before editing tracked files
+      run: |
+        # Edit CHANGELOG.md/Cargo.toml on main directly instead of in the tag
+        # checkout and carrying uncommitted edits across the branch switch.
+        # The carry-over only works when the tag's copies of those files equal
+        # main's; tagging a non-tip/older commit makes them differ, so
+        # `git checkout main` would abort ("local changes would be overwritten").
+        # Changelog generation above reads tags by name, so it is unaffected by
+        # the current branch.
+        git checkout main
+        git pull origin main
+
     - name: Update CHANGELOG.md
       run: |
         NEW_ENTRY_FILE="${{ steps.generate_changelog.outputs.new_entry_file }}"
@@ -198,11 +210,8 @@ jobs:
         git config --local user.name "github-actions[bot]"
         git config --local user.email "github-actions[bot]@users.noreply.github.com"
 
-        # Switch to main branch and pull latest changes (uncommitted
-        # CHANGELOG.md/Cargo.toml edits carry over since they share main's base).
-        git checkout main
-        git pull origin main
-
+        # Already on main (switched before the file edits above), so the
+        # CHANGELOG.md/Cargo.toml changes are staged and committed on main.
         # Check if there are any changes to commit
         if git diff --quiet CHANGELOG.md Cargo.toml; then
           echo "No changes to CHANGELOG.md or Cargo.toml"
--- a/codecov.yml
+++ b/codecov.yml
@@ -10,7 +10,7 @@ coverage:
         threshold: 1%
 
 ignore:
-  - "mortie/tests/**/*"
+  - "mortie/tests"
   - "setup.py"
   - "setup.cfg"
 


.github/workflows/build-wheels.yml
```diff
@@ -127,6 +127,18 @@ jobs:
...
+      run: |
+        # Edit CHANGELOG.md/Cargo.toml on main directly instead of in the tag
+        # checkout and carrying uncommitted edits across the branch switch.
+        # The carry-over only works when the tag's copies of those files equal
+        # main's; tagging a non-tip/older commit makes them differ, so
+        # `git checkout main` would abort ("local changes would be overwritten").
+        # Changelog generation above reads tags by name, so it is unaffected by
+        # the current branch.
+        git checkout main
+        git pull origin main
```
espg, line 140: 🤖 *from Claude (review)*

Non-blocking: moving `git pull origin main` here (instead of immediately before the commit/push, as previously) widens the window between the pull and the final `git push origin main` by three steps (Update CHANGELOG.md, Sync Cargo.toml, Commit). If a concurrent commit lands on `main` in that window, the push would be rejected non-fast-forward and the job would fail. In practice this is a tag-triggered release job and concurrent pushes to `main` mid-release are unlikely, so this is low severity and the same failure mode existed before (just a marginally smaller window). Flagging only for awareness — no change required.

espg (commented):
🤖 *from Claude (review)*

Adversarial fresh-eyes review of the `update-changelog` restructure and the codecov glob change. **No blocking issues found.** The core fix is sound: editing CHANGELOG.md/Cargo.toml after switching to `main` avoids the abort that occurred when `git checkout main` tried to carry uncommitted edits over a diverging working tree while tagging a non-tip commit.

What I verified:
- **Changelog/tag steps are branch-independent.** `Get tag name` uses `GITHUB_REF`; `Get previous tag` uses `git describe --tags ...`; `Generate changelog entry` resolves tags by name (`git log ... $CURRENT_TAG`, `$PREVIOUS_TAG`) and writes only to a `mktemp` file in `/tmp`. None read the working-tree copies of CHANGELOG.md/Cargo.toml, so moving the `main` checkout ahead of them is safe.
- **`Update CHANGELOG.md` / `Sync Cargo.toml` now read main's copies** (correct target for a non-tip tag). Previously they read the tag's copies. The `## [Unreleased]` / `version = ` no-match edge cases pre-date this PR (awk is a no-op passthrough; the sed step already fails loudly on no-match) and are unchanged.
- **The commit step is still correct on main.** No step between `Switch to main` and the commit changes branches, so `git diff --quiet` / add / commit / `git push origin main` all run on main. The stale carry-over comment was removed and replaced with an accurate one; no leftover cross-branch language remains.
- **`git checkout main` won't abort** at the new location — the changelog steps leave no untracked/modified files in the repo tree (temp files live in `/tmp`).
- **codecov `mortie/tests`** is an improvement: codecov treats a bare folder path as a recursive prefix, so it excludes both direct children (`mortie/tests/test_tools.py`) and nested files (`mortie/tests/foo/bar.py`). The prior `mortie/tests/**/*` glob risked missing direct children. Style is consistent with the neighboring `setup.py`/`setup.cfg` entries.
- **YAML is valid** — the inserted `Switch to main before editing tracked files` step matches sibling indentation (4-space `- name:`, `run: |` block).

See one minor non-blocking inline note below.

espg:
🤖 *from Claude*

**Heads-up on the `codecov/project` red check — it's an expected, one-time rebaseline, not a regression, but it's a policy call so I'm flagging it rather than marking this ready.**

The folded-in `codecov.yml` one-liner (`"mortie/tests/**/*"` → `"mortie/tests"`) does exactly what the [#94 comment](https://github.com/espg/mortie/pull/94#issuecomment-4859353255) intended: it now excludes **all 19 files under `mortie/tests/`** from coverage (the old glob missed the files sitting *directly* under `mortie/tests/`). codecov's report confirms it — head measures exactly the **9 source files** (`__init__`, `_healpix`, `arrow`, `coverage`, `geometry`, `linestring`, `morton_index`, `prefix_trie`, `tools`) instead of 27.

Consequences:

- ✅ **`codecov/patch` is now green** — that was the actual target of the #94 comment (previously red at ~85% because CI-skipped test files counted against the patch).
- ⚠️ **`codecov/project` fails** because total coverage moves **96.90% → 93.56% (−3.34%)**, tripping the 1% threshold. This is *not* lost coverage — the previously-counted test files are ~100% covered (they execute), so they were **inflating** the project number. 93.56% is the honest source-only figure. On merge, `main`'s baseline becomes 93.56% and subsequent PRs compare against that, so this red is a **one-time artifact** of the base still counting tests.
- The exclusion inherently applies to *both* patch and project — there's no glob that drops test files from patch coverage while keeping them in the project total. So the rebaseline is unavoidable with this fix.

The required gates (`test 3.10/3.11/3.12`, `ruff`, wheel builds, sdist) are all green.

**Question for review:** OK to accept the project-coverage rebaseline to 93.56% (recommended — it's the accurate source-only number and greens patch)? If you'd rather keep #92 scoped purely to the release-workflow fix, I can drop the `codecov.yml` line into its own separate PR instead. Your call.

PR #100, opened 2026-07-04
numpy-level point-kind geo2mort: geo2mort(..., points=True)

### LINKED ISSUE #96: numpy-level point-kind geo2mort: geo2mort(lat, lon, order=29, points=True)
## What's needed

A vectorized, numpy-in/numpy-out point-kind encoder on the `geo2mort` surface:

```python
geo2mort(lats, lons, order=29, points=True)  # -> np.ndarray[uint64], Kind::Point words
```

- `points=False` (default): exactly today's behavior — area-kind words, byte-identical, fully backward compatible.
- `points=True`: route through the kernel's existing point encoding (`rust_ang2pix` at order 29 → `rust_mi_from_nested_point`, i.e. `Kind::Point` via `from_nested_point` in `src_rust/src/decimal_morton.rs`) and return a plain contiguous `uint64` ndarray — no pandas ExtensionArray wrapper.
- Point encoding is order-29-only, so `points=True` with an explicit `order != 29` should raise `ValueError`, matching `MortonIndexArray.from_latlon`'s existing contract (`mortie/morton_index.py`).
- Null/sentinel semantics: same as the area path — inputs are dense float64 lat/lon with no null mask; non-finite lat/lon should fail loudly (or produce the documented sentinel `0`, whichever the area path does today — the two paths must agree). `0` remains the only reserved word (empty/invalid), never a valid encode result.
- Tests to include: (a) `points=True` words differ from area words at order 29 but `clip2order` coarsening is bit-identical at every order (this is the property zagg's grouping relies on); (b) a lone `points=True` word survives `common_ancestor` unchanged (kind preserved); (c) `points=True, order != 29` raises; (d) dtype is `uint64`, C-contiguous, same shape as input.

## Why (the hot-path numbers)

zagg's `HealpixGrid.assign` runs on every observation of every granule read (per-photon for ATL03 — millions of points per shard). It now needs point-kind words end-to-end (englacial/zagg#87: located t-digest channel via `common_ancestor`). The only public point encode today is the pandas ExtensionArray classmethod `MortonIndexArray.from_latlon(..., points=True)`, and the public ways out of the wrapper are element-wise. Measured on 1M random points (mortie 0.8.4):

| path | ms / 1M pts |
|---|---|
| `geo2mort` (area) | 42.6 |
| raw kernel `rust_ang2pix` + `rust_mi_from_nested_point` | 46.0 |
| `from_latlon(points=True)` construction alone | 45.2 |
| + public unwrap (`np.asarray` / `.to_numpy` / `.astype`) | ~490–560 |

The kernel is at parity with the area encode (~8%); the ~12x penalty is entirely the ExtensionArray→ndarray unwrap. zagg currently works around it by reading the array's internal `uint64` storage through its own boundary adapter (`zagg.grids.morton.morton_words`), which is fast (~70 ms total) but leans on a private attribute. A numpy-level `geo2mort(..., points=True)` gives the hot path a public vectorized surface and drops the wrapper entirely.

## Consumer

englacial/zagg#87 / englacial/zagg#165 — `HealpixGrid.assign` swaps one call (`from_latlon(points=True)` + unwrap → `geo2mort(..., points=True)`) the moment this lands in a release; that swap is the PR's tracked phase 6.

espg:
Closes #96.

## What this does

Adds a vectorized, numpy-in/numpy-out **point-kind** encoder on the `geo2mort` surface, so zagg's hot path (`HealpixGrid.assign`, per-photon for ATL03) gets a public vectorized point encode and drops the `MortonIndexArray` → ndarray unwrap entirely (the ~12x penalty measured in the issue).

lat/lon inputs are treated as **points by default** (indeterminate resolution, encoded at max precision), so:

```python
geo2mort(lats, lons)                    # -> order-29 Kind::Point words (bare call)
geo2mort(lats, lons, order=12)          # -> order-12 Kind::Area cells (explicit order => area)
geo2mort(lats, lons, points=True)       # -> order-29 Kind::Point words (explicit)
geo2mort(lats, lons, order=12, points=False)  # -> order-12 area (same as `order=12`)
```

Result is always a plain contiguous `uint64` ndarray — no ExtensionArray wrapper. Point encoding is order-29-only, so `points=True` with an explicit `order != 29` raises `ValueError` (matching `MortonIndexArray.from_latlon`). Non-finite `lat`/`lon` encode to the reserved empty word `0` (base cell 0 is the null sentinel) on both routes.

## Recorded design decisions (from review thread)

- **`points` defaults to point.** Per [this comment](https://github.com/espg/mortie/pull/100#issuecomment-4883387347) ("lat/lon inputs are generally expected to be of type point ... flip it"), a bare `geo2mort(lat, lon)` returns order-29 point words. An explicit `order` implies an **area** cell (points inferred `False`), so `geo2mort(lat, lon, order=18)` does not raise. Resolved via a tri-state `order=None`/`points=None`: `points = order is None`, then `order = order or 29` — using an identity check so `order=0` stays 0.
- **Non-finite → reserved `0`.** Per [this comment](https://github.com/espg/mortie/pull/100#issuecomment-4883376017) ("why don't we hash to the reserved 0? we already have base cell '0' reserved for this"), non-finite lat/lon map to `0` on both paths. The guard lives in a new `geo2mort_word` helper scoped to the public `geo2mort` surface, so `linestring`'s internal `geo2mort_scalar` hashing is unchanged.
- **Default order 18→29 is breaking** (bare call now returns order-29 words, not order-18); flagged in `CHANGELOG.md`. All in-tree callers pass an explicit `order`.

## Phases

- [x] **Phase 1 — Rust fused point path + binding.** `geo2mort_point_scalar`; `rust_geo2mort` gains `points` flag; Rust unit tests (point ≠ area at 29 but same nested cell; fused == compose).
- [x] **Phase 2 — Python surface + tests.** `geo2mort` points/order plumbing + validation + uint64 coercion; the four issue-specified tests + `from_latlon` equivalence.
- [x] **Phase 3 — fold self-review 1.** Real non-finite assertion, dead-`noqa` removal, broadcast test, `MAX_ORDER` constant.
- [x] **Phase 4 — maintainer directives.** Non-finite → reserved `0` (`geo2mort_word`, scoped to the public surface, + Rust test); flip the `points` default to point-by-default via the tri-state resolution; CHANGELOG line; updated/added tests (reserved-0, bare-call-defaults-to-point).
- [x] **Phase 5 — fold self-review 2.** Pin `order=0` resolves to an order-0 word (guards against a future truthiness regression); clarify the binding's low-level-default doc comment.

## How it was tested

- `cargo test` (13 geo2mort tests), `cargo fmt --check` clean, `cargo clippy` (no new warnings — only the pre-existing `useless_conversion` false-positives, none in this diff).
- `pytest -v`: **548 passed, 11 skipped**. `test_geo2mort_points.py` covers the four issue tests, `from_latlon`/kernel equivalence, broadcast, reserved-0 non-finite handling, bare-call-defaults-to-point, and pins the explicit-order (incl. `order=0`) area path.
- CI on the PR: codecov 100% on modified lines, CodSpeed reports no perf change.

Side note: the scalar path previously returned `int64` for small (northern) words and `uint64` for large (bit-63-set southern) words — a latent hemisphere-dependent dtype inconsistency. `geo2mort` now always returns a contiguous `uint64` ndarray.

## Questions for review

None outstanding — both earlier questions were answered on the thread and implemented in phase 4. Ready for your review.

### DIFF
--- a/mortie/tests/test_geo2mort_points.py
+++ b/mortie/tests/test_geo2mort_points.py
@@ -9,7 +9,7 @@ import numpy as np
 import pytest
 
 from mortie import _rustie, common_ancestor, geo2mort
-from mortie.tools import clip2order
+from mortie.tools import clip2order, infer_order_from_morton
 
 # A spread of locations across hemispheres and near the poles/prime meridian.
 _LATS = np.array([45.0, -80.0, 0.0, 12.3, 89.9, -89.9])
@@ -103,10 +103,13 @@ def test_point_matches_morton_index_from_latlon():
 def test_area_default_backward_compatible_at_explicit_order():
     """points=False is byte-identical to the pre-#96 area encode at a given order."""
     for order in (0, 6, 12, 18, 29):
+        area = geo2mort(_LATS, _LONS, order=order)
         np.testing.assert_array_equal(
-            geo2mort(_LATS, _LONS, order=order),
-            geo2mort(_LATS, _LONS, order=order, points=False),
+            area, geo2mort(_LATS, _LONS, order=order, points=False)
         )
+        # Pin the resolved order — order=0 is falsy, so this catches a regression
+        # to a truthiness `order or MAX_ORDER` resolution that would coerce 0->29.
+        assert infer_order_from_morton(int(area[0])) == order
 
 
 def test_nonfinite_encodes_reserved_zero():
--- a/mortie/tests/test_geo2mort_points.py
+++ b/mortie/tests/test_geo2mort_points.py
@@ -1,0 +1,128 @@
+"""Tests for the numpy-level point-kind encoder ``geo2mort(..., points=True)``.
+
+Covers the contract from issue #96: a vectorized, numpy-in/numpy-out point
+encoder on the ``geo2mort`` surface that routes through the kernel's order-29
+point path (``Kind::Point``) and returns a plain contiguous ``uint64`` ndarray.
+"""
+
+import numpy as np
+import pytest
+
+from mortie import _rustie, common_ancestor, geo2mort
+from mortie.tools import clip2order
+
+# A spread of locations across hemispheres and near the poles/prime meridian.
+_LATS = np.array([45.0, -80.0, 0.0, 12.3, 89.9, -89.9])
+_LONS = np.array([-122.0, 120.0, 0.0, 45.6, 179.9, -179.9])
+
+
+def _area(order=29):
+    return geo2mort(_LATS, _LONS, order=order)
+
+
+def _point():
+    return geo2mort(_LATS, _LONS, order=29, points=True)
+
+
+def test_point_differs_from_area_but_coarsens_identically():
+    """(a) point != area at order 29, but coarsening is bit-identical below 29.
+
+    This is the property zagg's grouping relies on: point and area words for the
+    same location share the entire order-<29 prefix, so grouping by a coarsened
+    word is independent of whether the words started as points or areas.
+    """
+    area = _area(29)
+    point = _point()
+    assert not np.array_equal(area, point), "point and area words must differ at 29"
+    # At every coarser order the two collapse to the same area cell.
+    for order in range(0, 29):
+        np.testing.assert_array_equal(
+            clip2order(order, point),
+            clip2order(order, area),
+            err_msg=f"point/area coarsening diverged at order {order}",
+        )
+
+
+def test_lone_point_survives_common_ancestor_unchanged():
+    """(b) a single point word passes through common_ancestor unchanged (kind kept)."""
+    for word in _point():
+        got = common_ancestor(np.array([word], dtype=np.uint64))
+        assert int(got) == int(word), "lone point word must be returned unchanged"
+
+
+def test_points_true_wrong_order_raises():
+    """(c) points=True with an explicit order != 29 raises ValueError."""
+    for bad_order in (0, 18, 28):
+        with pytest.raises(ValueError):
+            geo2mort(1.0, 2.0, order=bad_order, points=True)
+    # order=29 (or the default / None) is fine.
+    geo2mort(1.0, 2.0, order=29, points=True)
+    geo2mort(1.0, 2.0, points=True)
+    geo2mort(1.0, 2.0, order=None, points=True)
+
+
+def test_point_output_dtype_shape_contiguity():
+    """(d) result is uint64, C-contiguous, same shape as the input."""
+    out = _point()
+    assert out.dtype == np.uint64
+    assert out.flags["C_CONTIGUOUS"]
+    assert out.shape == _LATS.shape
+    # scalar in -> length-1 ndarray, matching the area path's always-ndarray return.
+    scalar = geo2mort(45.0, -122.0, points=True)
+    assert isinstance(scalar, np.ndarray)
+    assert scalar.shape == (1,)
+    assert scalar.dtype == np.uint64
+
+
+def test_point_matches_from_latlon_kernel_path():
+    """geo2mort(points=True) is bit-identical to the from_latlon internal path.
+
+    This is the equivalence zagg needs to swap ``from_latlon(points=True)`` +
+    unwrap for ``geo2mort(..., points=True)``. Cross-checked against the raw
+    kernel bindings (ang2pix(29) -> from_nested_point) that from_latlon composes,
+    so it holds without the optional pandas dependency.
+    """
+    via_geo = _point()
+    nested = _rustie.rust_ang2pix(29, _LONS, _LATS)
+    nested = np.ascontiguousarray(nested, dtype=np.uint64)
+    via_kernel = _rustie.rust_mi_from_nested_point(nested)
+    np.testing.assert_array_equal(via_geo, via_kernel)
+
+
+def test_point_matches_morton_index_from_latlon():
+    """Same equivalence, but against the public MortonIndexArray.from_latlon."""
+    pytest.importorskip("pandas")
+    from mortie.morton_index import MortonIndexArray
+
+    via_geo = _point()
+    arr = MortonIndexArray.from_latlon(_LATS, _LONS, points=True)
+    via_arr = np.asarray(arr).astype(np.uint64)
+    np.testing.assert_array_equal(via_geo, via_arr)
+
+
+def test_area_default_backward_compatible_at_explicit_order():
+    """points=False is byte-identical to the pre-#96 area encode at a given order."""
+    for order in (0, 6, 12, 18, 29):
+        np.testing.assert_array_equal(
+            geo2mort(_LATS, _LONS, order=order),
+            geo2mort(_LATS, _LONS, order=order, points=False),
+        )
+
+
+def test_nonfinite_handling_agrees_between_paths():
+    """Non-finite lat/lon resolve identically on the area and point paths.
+
+    The two share the order-29 hash step, so whatever the area path does with a
+    non-finite input (raise vs. produce a word), the point path must do too.
+    """
+    for bad in (np.nan, np.inf, -np.inf):
+        area_err = point_err = None
+        try:
+            geo2mort(bad, 0.0, order=29, points=False)
+        except Exception as exc:  # noqa: BLE001 - we compare the failure mode
+            area_err = type(exc)
+        try:
+            geo2mort(bad, 0.0, order=29, points=True)
+        except Exception as exc:  # noqa: BLE001
+            point_err = type(exc)
+        assert area_err == point_err, f"non-finite path mismatch for {bad}"
--- a/mortie/tests/test_geo2mort_points.py
+++ b/mortie/tests/test_geo2mort_points.py
@@ -109,36 +109,40 @@ def test_area_default_backward_compatible_at_explicit_order():
         )
 
 
-def test_nonfinite_handling_agrees_between_paths():
-    """Non-finite lat/lon resolve identically on the area and point paths.
+def test_nonfinite_encodes_reserved_zero():
+    """Non-finite lat/lon encode to the reserved empty word 0 on both paths.
 
-    The two share the order-29 hash step, so a non-finite input must land on the
-    same HEALPix nested cell (they differ only in Kind), and must fail — or not —
-    the same way. Today neither path raises (the hash maps non-finite inputs to a
-    definite cell rather than the reserved ``0``); see the PR's "Questions for
-    review" on whether the area path should instead reject non-finite input.
-    ``BaseException`` is used so a would-be PyO3 panic is caught, not just
-    ``Exception``.
+    Base cell 0 is the null sentinel, so 0 never collides with a real encode;
+    a NaN/inf lat or lon yields it on the area and point routes alike.
     """
-    from mortie.tools import mort2healpix
-
     for bad in (np.nan, np.inf, -np.inf):
-        area_err = point_err = None
-        area_word = point_word = None
-        try:
-            area_word = int(geo2mort(bad, 0.0, order=29, points=False)[0])
-        except BaseException as exc:
-            area_err = type(exc)
-        try:
-            point_word = int(geo2mort(bad, 0.0, order=29, points=True)[0])
-        except BaseException as exc:
-            point_err = type(exc)
-        assert area_err == point_err, f"non-finite failure mode mismatch for {bad}"
-        if area_err is None:
-            # Both produced a word: they must share the underlying nested cell.
-            assert mort2healpix(area_word)[0] == mort2healpix(point_word)[0], (
-                f"non-finite nested cell mismatch for {bad}"
-            )
+        assert int(geo2mort(bad, 0.0, order=29, points=False)[0]) == 0
+        assert int(geo2mort(bad, 0.0, points=True)[0]) == 0
+        assert int(geo2mort(0.0, bad, points=True)[0]) == 0
+    # In an array, only the non-finite rows are zeroed; finite rows encode normally.
+    lats = np.array([45.0, np.nan, -80.0])
+    lons = np.array([-122.0, 0.0, np.inf])
+    out = geo2mort(lats, lons, points=True)
+    assert int(out[0]) != 0
+    assert int(out[1]) == 0
+    assert int(out[2]) == 0
+
+
+def test_bare_call_defaults_to_point():
+    """A bare geo2mort(lat, lon) encodes max-precision point words (issue #96).
+
+    An explicit order instead asks for an area cell at that resolution, with
+    ``points`` inferred ``False`` (so no ValueError for a non-29 order).
+    """
+    bare = geo2mort(_LATS, _LONS)
+    np.testing.assert_array_equal(bare, _point())  # bare == points=True
+    # ... and differs from the order-29 area cell.
+    assert not np.array_equal(bare, geo2mort(_LATS, _LONS, order=29, points=False))
+    # An explicit order implies area (points inferred False), no raise.
+    np.testing.assert_array_equal(
+        geo2mort(_LATS, _LONS, order=12),
+        geo2mort(_LATS, _LONS, order=12, points=False),
+    )
 
 
 def test_point_broadcast_matches_elementwise():
--- a/mortie/tests/test_geo2mort_points.py
+++ b/mortie/tests/test_geo2mort_points.py
@@ -112,17 +112,39 @@ def test_area_default_backward_compatible_at_explicit_order():
 def test_nonfinite_handling_agrees_between_paths():
     """Non-finite lat/lon resolve identically on the area and point paths.
 
-    The two share the order-29 hash step, so whatever the area path does with a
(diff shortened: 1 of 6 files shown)


mortie/tests/test_geo2mort_points.py
```diff
@@ -0,0 +1,128 @@
...
+        area_err = point_err = None
+        try:
+            geo2mort(bad, 0.0, order=29, points=False)
+        except Exception as exc:  # noqa: BLE001 - we compare the failure mode
+            area_err = type(exc)
+        try:
+            geo2mort(bad, 0.0, order=29, points=True)
+        except Exception as exc:  # noqa: BLE001
+            point_err = type(exc)
+        assert area_err == point_err, f"non-finite path mismatch for {bad}"
```
espg: 🤖 *from Claude (review)*

This is the acceptance criterion the issue cares most about ("non-finite lat/lon should fail loudly *or* produce the sentinel `0` — the two paths must agree"), but as written the test is close to vacuous. I built the extension and ran the actual behavior:

```
nonfinite nan  points=False -> value 5764607523034234909
nonfinite nan  points=True  -> value 5764607523034234928
nonfinite inf  points=False -> value 1152921504606847005
nonfinite -inf points=False -> value 10376293541461622813
```

So on **both** paths a non-finite input neither raises nor returns `0` — the `healpix` hash silently maps NaN/±inf to a real cell (NaN → nested `1152921504606846976`) and packs a valid-looking non-zero word. Because nothing raises, `area_err == point_err == None` and the assert passes without checking anything meaningful.

Two concrete problems:
1. **It never verifies value agreement.** The docstring promises the paths "resolve identically", but the test only compares whether-it-raised. I confirmed the area and point words for NaN decode to the *same* nested cell (`1152921504606846976` both) — that's the property worth pinning. Assert decoded-nested equality (via `rust_mort2nested`), not just exception type.
2. **`except Exception` is the wrong net for the Rust boundary.** If the hash ever panicked instead of returning a word, PyO3 surfaces it as `pyo3_runtime.PanicException`, which derives from `BaseException`, not `Exception` — it would escape this handler and error the test rather than being compared. `except BaseException` (or asserting the value) is the robust form.

Separately, worth surfacing for a maintainer decision (not something to silently absorb): the issue said "`0` remains the only reserved word, never a valid encode result," yet a NaN photon — a realistic ATL03 fill value on the per-photon hot path this PR targets — silently encodes to a plausible non-zero word on **both** the area and point paths. This is pre-existing area-path behavior and matches the "mirror the area path" directive, so it isn't a regression, but the current test actively obscures it.

---

mortie/tests/test_geo2mort_points.py
```diff
@@ -0,0 +1,128 @@
...
+    """Non-finite lat/lon resolve identically on the area and point paths.
+
+    The two share the order-29 hash step, so whatever the area path does with a
+    non-finite input (raise vs. produce a word), the point path must do too.
+    """
+    for bad in (np.nan, np.inf, -np.inf):
+        area_err = point_err = None
+        try:
+            geo2mort(bad, 0.0, order=29, points=False)
+        except Exception as exc:  # noqa: BLE001 - we compare the failure mode
```
espg: 🤖 *from Claude (review)*

`# noqa: BLE001` is dead here. Per CLAUDE.md §7 the only linters configured are `flake8 mortie --select=E9,F63,F7,F82` (blocking) and a non-blocking `--max-line-length=88` style pass — there is no `ruff`/flake8-bugbear, so `BLE001` (a ruff code) is never emitted. The suppression silences nothing and reads as if a bugbear check exists. CLAUDE.md §4 also discourages adding `# noqa` to quiet warnings rather than fixing the cause. Drop the `# noqa` (and if you take the `except BaseException` suggestion on the assert below, the broad-except concern the code was pre-empting goes away too).

---

mortie/tests/test_geo2mort_points.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def test_point_output_dtype_shape_contiguity():
+    """(d) result is uint64, C-contiguous, same shape as the input."""
+    out = _point()
+    assert out.dtype == np.uint64
+    assert out.flags["C_CONTIGUOUS"]
+    assert out.shape == _LATS.shape
+    # scalar in -> length-1 ndarray, matching the area path's always-ndarray return.
+    scalar = geo2mort(45.0, -122.0, points=True)
```
espg, line 71: 🤖 *from Claude (review)*

Coverage gap on the point path: the `points=True` branch in `rust_geo2mort` has three sub-branches — both-scalar, equal-length array, and **broadcast** (scalar-lat + array-lon, or vice-versa, via the `lat_bcast`/`lon_bcast` flags in `lib.rs`). This suite exercises the equal-length array path (`_point()`) and the both-scalar path (this line), but never the broadcast branch with `points=True`. A one-liner like `geo2mort(45.0, _LONS, points=True)` asserting shape `== _LONS.shape` and dtype `uint64` would close it and guard the branch that `geo2mort_point_scalar` is threaded through. (Empty-input `geo2mort([], [], points=True)` also works — returns `(0,)` uint64 — and is worth a line if you want the boundary pinned.)

---

mortie/tools.py
```diff
@@ -103,23 +103,55 @@ def geo2uniq(lats, lons, order=18):
     return uniq
 
 
-def geo2mort(lats, lons, order=18):
+def geo2mort(lats, lons, order=29, points=False):
```

The `order=18 → order=29` default flip is an authorized change (the maintainer asked for it directly on the issue), so this is not an objection — two follow-through notes:

1. **It's a silent public behavior change for the bare `geo2mort(lat, lon)` call**: external callers relying on the old order-18 default now get a different (order-29) word. In-tree this is safe — I grepped every caller and all pass an explicit `order=` except the docstring examples — but it should be called out explicitly in the PR description / changelog as a breaking behavior change, not just as "issue #96".
2. **Minor consistency:** this function hardcodes the literal `29` twice (the `order=None` resolution on line 140 and the guard on line 141), whereas the sibling `MortonIndexArray.from_latlon` and the Rust guard both compare against the `MAX_ORDER` constant. Not load-bearing (29 is the u64-packing ceiling and won't move), but a named constant would keep the two surfaces from drifting.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial pass on the point-encoder. I built the extension (`maturin develop --release`) and exercised the behavior directly. The core encode is correct against every acceptance criterion in issue #96: `points=True` words differ from area at order 29 but decode to the same nested cell; `clip2order` coarsening is bit-identical at orders 0–28 (clip at 29 preserves kind, so the test's `range(0, 29)` cap is intentional and correct — including 29 would fail, which is right); a lone point survives `common_ancestor` unchanged; output is `uint64`/C-contiguous/same-shape; the uint64 coercion is correct including base cells 7–11 with bit 63 set; `from_latlon(points=True)` equivalence holds; and the `ValueError` message matches `from_latlon` verbatim. The fused Rust path and default-order flip are both what the maintainer signed off on. No correctness bug found.

The findings are all test-quality / robustness, plus one thing to surface:

1. **`test_nonfinite_handling_agrees_between_paths` is close to vacuous** (biggest one). It compares only *exception type*, but empirically neither path raises on NaN/±inf — both silently emit a valid-looking non-zero word — so the assert passes without checking anything. It should assert decoded-nested equality (which does hold), and use `except BaseException` (a PyO3 panic surfaces as `PanicException`, a `BaseException`). Related: this quietly means a NaN fill value on the per-photon hot path encodes to a plausible word rather than the reserved `0` or a loud failure the issue mentioned — pre-existing area-path behavior, but worth a maintainer eye.
2. **Dead `# noqa: BLE001`** — no ruff/bugbear is configured (§7), so it silences nothing and misleads.
3. **Untested broadcast branch** for `points=True` (scalar + array), one of the three sub-paths in `rust_geo2mort`.
4. **Default `order` 18→29 is a silent public behavior change** — authorized, but flag it as breaking in the PR body/changelog; and the Python side hardcodes `29` where the siblings use `MAX_ORDER`.

None of these block the encode; (1) is the one I'd fix before merge.

---

espg:
> Non-finite input semantics. The issue asked the point path to match the area path on non-finite lat/lon. Today neither raises: a NaN/±inf lat/lon hashes to a definite order-29 nested cell (e.g. NaN → 1152921504606846976) and encodes to a plausible-looking word rather than the reserved 0 or a loud error. The point path matches the area path exactly (the shared hash step), so this PR is faithful to the directive — but on zagg's per-photon hot path a NaN fill silently becoming a valid word may be undesirable. Do you want the area path (and therefore both) changed to reject non-finite input? That's a behavior change to the existing area encode, so I've left it out of this PR's scope and raise it here rather than acting on it.

why don't we has to the reserved 0 ? we already have base cell '0' reserved for this...

espg:
> Default-order change (breaking). I read "update the default to order 29 / point" as: default order → 29, points stays an opt-in flag defaulting False (so a bare geo2mort(lat, lon) returns order-29 area words, and points=True opts into point words). This honors the issue body's "points=False (default) → area-kind" while making 29 the new default order. The backward-compat consequence, flagged explicitly: a bare no-order geo2mort(lat, lon) now returns order-29 words instead of order-18. All in-tree callers pass an explicit order, so nothing internal breaks, but it's a public change worth a changelog line. 

this seems fine.

> All in-tree callers pass an explicit order, so nothing internal breaks, but it's a public change worth a changelog line.

agreed.

> If you instead meant the default should emit point words, say so and I'll flip the points default.

i did mean that. lan / lon inputs are generally expected to be of type point (i.e., indeterminate resolution, encoded at max precision). so... flip it

mortie/tools.py
```diff
@@ -103,23 +108,67 @@ def geo2uniq(lats, lons, order=18):
...
+    Returns
+    -------
+    ndarray
+        Packed ``uint64`` morton word(s), same shape family as the input
+        (scalar in -> length-1 ndarray)."""
+
+    # Resolve the point/area mode: a bare call encodes points; an explicit order
+    # implies an area cell at that resolution unless the caller forces points.
+    if points is None:
+        points = order is None
```
espg, line 155: 🤖 *from Claude (review)*

Tri-state resolution is correct, and I want to explicitly flag the `order=0` falsy trap since it's the easiest place to have introduced a bug — you avoided it. Because this uses the identity test `order is None` (not a truthiness test like `order or MAX_ORDER`), a falsy-but-valid `order=0` is handled right:

- `geo2mort(lat, lon, order=0)` → `points = (0 is None) = False`, `order` stays `0` → order-0 area cell. ✔
- `geo2mort(lat, lon, order=0, points=True)` → raises (0 != 29). ✔ (tested)

Enumerated the full truth table (bare / 18 / 29 / points=True / points=True,order=18 / points=False / points=False,order=18 / order=0) — all resolve as intended.

One residual gap: nothing in the suite actually **pins** that `order=0` produces an order-0 word. `test_area_default_backward_compatible_at_explicit_order` iterates `order=0` but only compares `order=0` against `order=0, points=False`, so a hypothetical future `order or MAX_ORDER` regression (0 → 29) would keep both sides equal and slip past. A one-line `assert infer_order_from_morton(int(geo2mort(_LATS, _LONS, order=0)[0])) == 0` would lock the falsy-order behavior down. Non-blocking.

---

mortie/tests/test_geo2mort_points.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    bare = geo2mort(_LATS, _LONS)
+    np.testing.assert_array_equal(bare, _point())  # bare == points=True
+    # ... and differs from the order-29 area cell.
+    assert not np.array_equal(bare, geo2mort(_LATS, _LONS, order=29, points=False))
+    # An explicit order implies area (points inferred False), no raise.
+    np.testing.assert_array_equal(
+        geo2mort(_LATS, _LONS, order=12),
+        geo2mort(_LATS, _LONS, order=12, points=False),
+    )
```
espg, line 148: 🤖 *from Claude (review)*

Good coverage of the flipped default here (bare == points=True, explicit order == area). The one thing not asserted anywhere is that an explicit **falsy** `order=0` stays order 0 rather than being coerced to 29 — see my note on `tools.py`. Since `order=0` is now the only falsy-but-valid order flowing through the `order is None` gate, it's worth an explicit pin (`infer_order_from_morton(...) == 0`) so the tri-state logic can't silently regress. Non-blocking.

---

src_rust/src/lib.rs
```diff
@@ -202,27 +202,35 @@ fn split_children_rust(
...
+        ));
+    }
 
     let (lat_arr, lats_is_scalar) = extract_f64_input(lats)?;
     let (lon_arr, lons_is_scalar) = extract_f64_input(lons)?;
 
     // Both scalars → return scalar
     if lats_is_scalar && lons_is_scalar {
-        let result = geo2mort::geo2mort_scalar(lat_arr[0], lon_arr[0], order);
+        let result = geo2mort::geo2mort_word(lat_arr[0], lon_arr[0], order, points);
```
espg, line 238: 🤖 *from Claude (review)*

Routing both the scalar and array paths through `geo2mort_word` is a clean dedup, and scoping the non-finite→0 guard inside `geo2mort_word` (rather than `geo2mort_scalar`) correctly leaves `linestring`'s internal hashing untouched — verified `linestring.rs:126` still calls `geo2mort_scalar` directly. ✔

Minor drift note (non-blocking): this pyfunction still carries `#[pyo3(signature = (lats, lons, order=29, points=false))]` and the docstring says "order (default 29)". All of the new point/area/tri-state policy now lives in Python (`tools.py` always passes both args explicitly), so this Rust-level default is dead for the public surface — but a direct `_rustie.rust_geo2mort(lat, lon)` still yields an order-29 **area** word, i.e. the pre-flip behavior. Harmless today, but it's a documentation/behavior-drift footgun if anyone later calls the binding directly. Consider a one-line comment noting the binding defaults are internal-only and the public default lives in `geo2mort`.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-eyes review of phase 4 (`295113f`) — the non-finite→reserved-`0` and default-flip directives. Both directives are faithfully implemented; the change is correct. Findings are all non-blocking nits.

**Verified correct**
- **Tri-state truth table** — enumerated all 8 cases (bare / order=18 / order=29 / points=True / points=True+order=18 / points=False / points=False+order=18 / order=0). All resolve as intended. Critically, the `order=0` falsy trap is **avoided**: the code uses `order is None` (identity), not `order or MAX_ORDER` (truthiness), so `order=0` stays `0` and infers `points=False` — an order-0 area cell, not a coerced order-29 word.
- **Non-finite guard** — correct for NaN/±inf in either lat or lon, on both routes, and correctly scoped: living in `geo2mort_word` (not `geo2mort_scalar`) leaves `linestring`'s internal hashing unchanged (`linestring.rs:126` still calls the scalar directly). Array path zeroes only the bad rows (tested), empty array still fine. `0` is genuinely unreachable by a real encode (prefix is base+1, so the empty sentinel never collides).
- **CHANGELOG "Breaking" claim** — "was order-18 area cells" is accurate against the last release (0.8.4 defaulted `order=18`); the intermediate order-29-area default was itself unreleased. The `[Unreleased]` block has a single, non-contradictory line.
- Point/area/broadcast/dtype/from_latlon-equivalence and the `points=True, order=18` raise all remain covered.

**Non-blocking nits** (inline)
1. No test **pins** that a falsy `order=0` yields an order-0 word — the backward-compat test compares `order=0` to `order=0`, so a future `order or MAX_ORDER` regression would slip past. A one-line `infer_order_from_morton(...) == 0` assert would lock it down. (Highest-value follow-up.)
2. `rust_geo2mort`'s pyo3 signature default (`order=29, points=false`) and docstring now describe an internal-only default that diverges from the public `geo2mort` point default — harmless (Python always passes explicit args) but a drift footgun for direct-binding callers.

Nothing here blocks the phase.

---

espg:
🤖 *from Claude*

Both landed in phase 4 (self-reviewed + folded in phase 5):

- **Non-finite → reserved `0`.** Non-finite `lat`/`lon` now encode to `0` on both the area and point routes, via a new `geo2mort_word` helper. I scoped the guard to the public `geo2mort` surface rather than the shared `geo2mort_scalar` kernel, so `linestring`'s internal hashing is unchanged (it can keep its own handling of bad coordinates). In an array, only the non-finite rows are zeroed.
- **Point-by-default.** A bare `geo2mort(lat, lon)` now returns order-29 `Kind::Point` words. To avoid making `geo2mort(lat, lon, order=18)` raise, an explicit `order` implies an **area** cell (points inferred `False`); `points=True` still forces the order-29 point path. Resolved with a tri-state (`points = order is None`, then `order = order or 29`), using an identity check so `order=0` stays 0 — pinned by a test.

Truth table now: bare → point@29; `order=N` → area@N; `points=True` → point@29 (non-29 order raises); `points=False` → area (order defaults 29). CHANGELOG carries the breaking-default note. Green locally (548 pytest passed, cargo test/fmt/clippy clean) and CI is green (codecov 100% on the diff, CodSpeed no perf change). Ready for another look.


---

Issue #93, opened 2026-06-30
make the morton-index Arrow surface library-agnostic (arro3-core / PyCapsule + extension type)


espg:
Follow-on to #86 (the pyarrow-based from/to-Arrow surface for morton indices). Context thread: englacial/zagg#135.

**Goal.** Make the `morton_index` Arrow surface consumable *and* producible by **arro3-core** (the carrier zagg runs on its AWS Lambda worker, where **pyarrow is intentionally not on the layer**), carrying a real **Arrow extension type** rather than shimming through `uint64`.

**Why.** #86 added `MortonIndexDtype.__from_arrow__` (Arrow→pandas import) and the `mortie.arrow` null↔sentinel mapping, but (a) it's gated on pyarrow (`arrow._require_pyarrow`, built from `pa.Array`/`fill_null`/`to_pandas`), and (b) it has no **export** hook (pandas/numpy → Arrow). zagg's hot path needs the export direction, on arro3-core, with no pyarrow.

**Proposed scope.**
1. **Export via the Arrow PyCapsule C Data Interface** — `MortonIndexArray.__arrow_c_array__` / `__arrow_c_schema__` so any Arrow lib (arro3-core, pyarrow, polars) can pull a typed morton column zero-copy; add the import side over the same interface.
2. **Carry a `morton_index` Arrow extension type** (logical type + metadata) so the dtype survives the boundary — the deliberate decision over plain-`uint64` storage. Confirm arro3-core round-trips the extension metadata on the Lambda layer; if not, document the fallback to `uint64` storage with the morton-ness re-attached at the edges.
3. **Drop the hard pyarrow assumption** from the critical path; keep the pyarrow `__from_arrow__`/`to_pandas` convenience as an optional nicety for off-worker analysis flows.
4. **Preserve null/sentinel semantics** (#86's `data == _SENTINEL` → Arrow null; `fill_null(_SENTINEL)` back) byte-for-byte, with a round-trip test on an **arro3-core-only env (no pyarrow)** — the environment #86 never exercises.
5. **Point encoding** (`from_latlon(points=True)`, order-29) is orthogonal and out of scope here.

[...]

PR #101, opened 2026-07-04
Wire up the arro3-no-pyarrow CI leg

### LINKED ISSUE #92: Harden release changelog/version commit against non-tip tags

### LINKED ISSUE #98: arro3-core CI leg for the pyarrow-free consumer was designed but never wired up
PR https://github.com/espg/mortie/pull/94 built the Arrow C Data Interface specifically so zagg's pyarrow-free Lambda worker can consume morton_index (TestArro3Interop, currently skipif-gated and local-only — arro3-core is deliberately kept out of the test extra). A concrete arro3-no-pyarrow CI job was proposed with exact YAML in a comment on issue https://github.com/espg/mortie/issues/92 (https://github.com/espg/mortie/issues/92#issuecomment-4859163096), explicitly because the routine lacks the workflow OAuth scope needed to push .github/workflows/ changes itself. Issue https://github.com/espg/mortie/issues/92 was subsequently closed by PR https://github.com/espg/mortie/pull/95 — but https://github.com/espg/mortie/pull/95 only fixed the changelog/version-carryover and codecov.yml; the CI leg was never added. The one real pyarrow-absent consumer this feature exists for currently has zero CI coverage.

espg:
Closes #98. Refs #92, #93.

## What this does

Adds the `arro3-no-pyarrow` job to `test.yml` — the CI leg designed in the [#92 proposal comment](https://github.com/espg/mortie/issues/92#issuecomment-4859163096) but never landed (the routine lacked the `workflow` OAuth scope; #95 closed #92 with only the changelog/codecov fixes). This is the leg that actually exercises `TestArro3Interop` from PR #94: mortie built and installed with **arro3-core 0.8.1 present and pyarrow absent**, proving the pyarrow-free carrier contract zagg's Lambda worker depends on.

The job:

1. installs `maturin numpy pandas pytest pytest-cov "arro3-core==0.8.1"` — **no pyarrow**;
2. builds/installs mortie (`pip install -e .` — numpy is the only runtime dep, so nothing drags pyarrow in transitively);
3. **asserts pyarrow is absent** (`python -c "import pyarrow"` must fail, else the job errors) — so a future transitive pyarrow can't silently un-prove the contract;
4. runs `pytest mortie/tests/test_arrow_cdata.py -v -rs --junitxml=arro3-junit.xml`;
5. **hard-fails on silent skips**: parses the junit XML and errors if `TestArro3Interop` collected zero tests or any of its tests skipped. A leg that quietly skips the arro3 tests would recreate #98, so skipping is a failure, not a pass.

The 9 pyarrow-gated skips in the rest of `test_arrow_cdata.py` are expected in this leg (that's the point); the guard is scoped to `TestArro3Interop` only. The full pyarrow-dependent suite stays in the existing `test` matrix job, untouched.

## Deltas from the #92-proposed YAML

- `pip install` list gains **`pytest-cov`** — `pyproject.toml`'s pytest `addopts` pass `--cov=mortie`, so bare `pytest` errors with "unrecognized arguments" without it.
- `maturin develop --release` → **`pip install -e .`** — `maturin develop` requires an activated virtualenv/conda env, which a plain `setup-python` runner doesn't provide; `pip install -e .` builds through the maturin backend (release profile) exactly like the existing `test` job.
- Added the pyarrow-absence assertion and the junit skip guard (steps 3 and 5 above) — the issue #98 requirement beyond the original sketch.
- Steps carry names matching the existing job's style; everything else (ubuntu-latest, Python 3.11, `dtolnay/rust-toolchain@stable`, `arro3-core==0.8.1` pin, scope limited to `test_arrow_cdata.py`) is as proposed.

## How it was tested

Local dry-run of the exact leg semantics (macOS arm64, Python 3.11.9, cargo 1.96.0, maturin 1.14.1): built the wheel with `maturin build --release`, installed it into a fresh venv with `numpy pandas pytest pytest-cov "arro3-core==0.8.1"` and **no pyarrow** (`import pyarrow` fails; `pip list` shows no pyarrow), ran `test_arrow_cdata.py`:

- **14 passed, 9 skipped** — all 9 skips are `importorskip("pyarrow")` in `TestPyarrowInterop`, as designed;
- **all 4 `TestArro3Interop` tests passed** (extension-metadata consume, null round-trip, chunked stream, no-pyarrow-symbol check);
- the junit guard script run against that report prints `TestArro3Interop: 4 ran, 0 skipped` and was also verified to fail on a skipped/missing class.

Workflow YAML parses clean (`yaml.safe_load`). The real proof is this PR's own CI run of the new job — see the `arro3-no-pyarrow` check on this PR.

## Questions for review

- The guard requires ≥1 `TestArro3Interop` test collected and 0 skipped, rather than pinning the exact count (currently 4), so adding an arro3 test doesn't need a workflow edit. Pinning `len(cases) == 4` is the stricter alternative if you'd rather be alerted on count drift.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,3 +57,56 @@ jobs:
         name: codecov-mortie
         fail_ci_if_error: false
         token: ${{ secrets.CODECOV_TOKEN }}
+
+  # Proves the pyarrow-free Arrow C Data carrier (issue #93 / PR #94): arro3-core
+  # installed, pyarrow deliberately absent, so TestArro3Interop actually runs
+  # instead of skipping. Designed on issue #92; wired up for issue #98.
+  arro3-no-pyarrow:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Set up Python 3.11
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.11'
+
+    - name: Install Rust toolchain
+      uses: dtolnay/rust-toolchain@stable
+
+    - name: Install dependencies (NOTE - no pyarrow)
+      run: |
+        python -m pip install --upgrade pip
+        # pytest-cov: pyproject addopts pass --cov, so bare pytest would error.
+        # arro3-core==0.8.1 is the pinned Lambda version (issue #93 item 2).
+        pip install maturin numpy pandas pytest pytest-cov "arro3-core==0.8.1"
+
+    - name: Build and install mortie
+      run: pip install -e .
+
+    - name: Assert pyarrow is absent
+      run: |
+        if python -c "import pyarrow" 2>/dev/null; then
+          echo "::error::pyarrow is importable in the no-pyarrow leg; it must stay absent"
+          exit 1
+        fi
+        echo "pyarrow absent, as required"
+
+    - name: Run Arrow C Data interop tests (arro3-core, no pyarrow)
+      run: |
+        pytest mortie/tests/test_arrow_cdata.py -v -rs --junitxml=arro3-junit.xml
+
+    - name: Fail if the arro3 interop tests skipped
+      run: |
+        python - <<'EOF'
+        import xml.etree.ElementTree as ET
+
+        cases = [c for c in ET.parse("arro3-junit.xml").getroot().iter("testcase")
+                 if c.get("classname", "").endswith("TestArro3Interop")]
+        assert cases, "no TestArro3Interop tests collected -- the arro3 leg did not run"
+        skipped = [c.get("name") for c in cases if c.find("skipped") is not None]
+        assert not skipped, f"arro3 interop tests skipped instead of running: {skipped}"
+        print(f"TestArro3Interop: {len(cases)} ran, 0 skipped")
+        EOF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -88,11 +88,13 @@ jobs:
 
     - name: Assert pyarrow is absent
       run: |
-        if python -c "import pyarrow" 2>/dev/null; then
-          echo "::error::pyarrow is importable in the no-pyarrow leg; it must stay absent"
-          exit 1
-        fi
-        echo "pyarrow absent, as required"
+        python - <<'EOF'
+        import importlib.util
+
+        spec = importlib.util.find_spec("pyarrow")
+        assert spec is None, f"pyarrow is installed in the no-pyarrow leg: {spec.origin}"
+        print("pyarrow absent, as required")
+        EOF
 
     - name: Run Arrow C Data interop tests (arro3-core, no pyarrow)
       run: |


.github/workflows/test.yml
```diff
@@ -57,3 +57,56 @@ jobs:
...
+        # pytest-cov: pyproject addopts pass --cov, so bare pytest would error.
+        # arro3-core==0.8.1 is the pinned Lambda version (issue #93 item 2).
+        pip install maturin numpy pandas pytest pytest-cov "arro3-core==0.8.1"
+
+    - name: Build and install mortie
+      run: pip install -e .
+
+    - name: Assert pyarrow is absent
+      run: |
+        if python -c "import pyarrow" 2>/dev/null; then
```
espg: 🤖 *from Claude (review)*

**Minor (non-blocking): the absence gate asserts "not importable", not "not installed".**

`if python -c "import pyarrow"` treats a pyarrow that is *installed but fails to import* (e.g. a broken transitive wheel / ABI mismatch) the same as "absent" — the leg would stay green with a pyarrow dist sitting in the env, and `2>/dev/null` also discards the traceback that would tell you which case you're in. Since the step's stated purpose is "a future transitive pyarrow can't silently un-prove the contract", a distribution-level check is the tighter match:

```yaml
run: |
  python -c "import importlib.util, sys; sys.exit(1 if importlib.util.find_spec('pyarrow') else 0)" \
    || { echo "::error::pyarrow is installed in the no-pyarrow leg; it must stay absent"; exit 1; }
  echo "pyarrow absent, as required"
```

For what it's worth, the current `if`-form is otherwise correct under the default `bash -e {0}` step shell — a non-zero exit inside an `if` condition doesn't abort the step, so there's no false-fail lurking here; this is purely about tightening what "absent" means.

.github/workflows/test.yml
```diff
@@ -57,3 +57,56 @@ jobs:
...
+
+    - name: Fail if the arro3 interop tests skipped
+      run: |
+        python - <<'EOF'
+        import xml.etree.ElementTree as ET
+
+        cases = [c for c in ET.parse("arro3-junit.xml").getroot().iter("testcase")
+                 if c.get("classname", "").endswith("TestArro3Interop")]
+        assert cases, "no TestArro3Interop tests collected -- the arro3 leg did not run"
+        skipped = [c.get("name") for c in cases if c.find("skipped") is not None]
```
espg, line 111: 🤖 *from Claude (review)*

**Informational (non-blocking): xfail counts as "skipped" here.**

Verified empirically (pytest 8.3, default `xunit2` junit family): an `@pytest.mark.xfail` test is recorded as `<testcase ...><skipped type="pytest.xfail" .../></testcase>`, so this guard would hard-fail the leg if a `TestArro3Interop` test is ever marked xfail — the same way it fails on a real skip. That's arguably the *right* strictness for a contract-proving leg (an xfailed interop test isn't proving anything), so no change requested — just flagging it so a future xfail-triggered red on this job reads as a design decision, not a guard bug. If the looser behavior were ever wanted, filter on `c.find("skipped").get("type") != "pytest.xfail"`.

Separately: I checked the classname-matching assumption the guard rests on — with rootdir at the repo root, junit `classname` comes out as `mortie.tests.test_arrow_cdata.TestArro3Interop`, and a class-level `skipif` still emits one `<testcase>` per gated test with a `<skipped>` child (verified locally). So `endswith("TestArro3Interop")` + the non-empty-cases assert catch both the silent-skip and the renamed/vanished-class failure modes correctly.

.github/workflows/test.yml
espg: 🤖 *from Claude*

Folded in `5e63783`: the gate now uses `importlib.util.find_spec("pyarrow")` and asserts it is `None`, so an installed-but-broken pyarrow distribution fails the leg too, and the assertion message surfaces `spec.origin` instead of discarding diagnostics.

.github/workflows/test.yml
espg: 🤖 *from Claude*

Agreed, and leaving as-is: an xfail-marked `TestArro3Interop` test hard-failing this leg is intentional strictness — any state where the arro3 tests are not plainly passing should be a loud red here, since a quietly weakened leg is exactly what issue #98 exists to prevent.

espg:
🤖 *from Claude*

Status: both CI runs green, including the new `arro3-no-pyarrow` leg (proof it exercised the tests, from the [job log](https://github.com/espg/mortie/actions/runs/28715188004/job/85155195163): `pyarrow absent, as required` → `14 passed, 9 skipped` → `TestArro3Interop: 4 ran, 0 skipped`). Self-review findings folded (`5e63783`) or answered inline.

On the codspeed comment: this PR touches only `.github/workflows/test.yml` — no Rust or Python code path — and codspeed itself flags the comparison as cross-environment, so the single 16.65% "regression" is measurement noise, not a real change. The acknowledge action on CodSpeed is left for a human.

PR #102, opened 2026-07-05
Fix res2display() order cap and add km/m/cm unit ladder

### LINKED ISSUE #99: res2display() order cap still wrong above order 20 — carry-forward from #88 — PR #76
mortie/tools.py::res2display() still loops for res in range(20), so it silently returns nothing for any order above 19 even though the packed-u64 kernel supports order 29. 

...it would be ideal if this had some checking added to format the floating point sensibly, especially when it drops below 1.0 ; i.e., 

```
1.5889124267578125 km at tessellation order 12  # this should truncate down to 1.589 km
0.7944562133789063 km at tessellation order 13 # this should return meters, i.e., 794.456 m
```

espg:
Closes #99.

## What this does

`mortie/tools.py::res2display()` looped `for res in range(20)`, so it silently printed nothing for orders 20–29 even though the packed-u64 kernel reaches order 29 (`MAX_ORDER`). It also printed every order in km, so fine orders read as tiny km fractions (`0.0248… km`) instead of naturally.

Two changes, matching the design signed off on the issue thread ([#99#issuecomment-4883350921](https://github.com/espg/mortie/issues/99#issuecomment-4883350921)):

1. **Extend the range** to `range(max_order + 1)`, defaulting to `MAX_ORDER` (29), sourced from the existing module constant rather than a bare literal. Added an optional `max_order=29` parameter so callers can bound it; the no-arg call stays backward compatible.
2. **km → m → cm unit ladder**, choosing the largest unit that keeps the value ≥ 1, and rounding to three decimals *within* that bracket. Order 12 now reads `1.589 km` and order 13 reads `794.456 m` — exactly the issue's requested output.

### Approach

The bracket cutoffs key off `order2res` (which returns km and was already order-unbounded):

```python
km = order2res(res)
if km >= 1.0:
    value, unit = km, 'km'
elif km >= 1e-3:                 # >= 1 m
    value, unit = km * 1e3, 'm'
else:                            # < 1 m
    value, unit = km * 1e5, 'cm'
print(str(round(value, 3)) + ' ' + unit + ' at tessellation order ' + str(res))
```

`round(value, 3)` reproduces the issue's examples: `1.588912… → 1.589`, `794.4562… → 794.456`, `19.396`, `1.212`. The three design questions were answered on the thread: km/m/cm ladder (yes), optional `max_order=29` (yes), round to three decimals within a bracket (yes).

Full rendered output for reference:

```
6508.185 km at tessellation order 0
...
3.178 km at tessellation order 11
1.589 km at tessellation order 12
794.456 m at tessellation order 13
...
1.552 m at tessellation order 22
77.584 cm at tessellation order 23
...
19.396 cm at tessellation order 25
...
1.212 cm at tessellation order 29
```

## Phases

- [x] **Phase 1** — extend range + km/m/cm unit ladder in `res2display`, plus `capsys` tests in `test_tools.py` (order count 0–29, km/m boundary at order 12/13, cm at the fine end, ≤3-decimal rounding, and the `max_order` bound).
- [x] **Phase 2** — folded self-review finding: reject `max_order` outside `0..MAX_ORDER` with a `ValueError` (negative was silently a no-op; `> 29` printed sub-1 cm past the kernel floor), with a test covering both ends; hardened the rounding-test assertion against a `'.'`-less value.
- [x] **Phase 3** — folded self-review convention nits: swapped the `%`-format raise to an f-string to match the module's uniform style, and tightened the docstring to cite the kernel's order-29 encoding limit.

The per-phase adversarial self-review ran on phases 1 and 2 and converged — the remaining findings were all low-severity convention nits, now addressed in phases 2–3; phase 3 changed no logic (f-string + docstring only).

## How it was tested

- `maturin develop --release` (Rust ext rebuilt), `flake8 mortie --select=E9,F63,F7,F82` clean.
- `pytest mortie/tests/test_tools.py` — 46 passed (6 new in `TestRes2Display`). No Rust change, so `cargo` gates are untouched.
- Manually ran `res2display()`; output matches the issue's requested examples byte-for-byte.

## Questions for review

None outstanding — the three design questions were resolved on the issue thread before implementation.

### DIFF
--- a/mortie/tools.py
+++ b/mortie/tools.py
@@ -34,7 +34,15 @@ def res2display(max_order=MAX_ORDER):
     rounded to three decimals within that bracket, so fine orders read
     naturally (e.g. order 12 -> ``1.589 km``, order 13 -> ``794.456 m``)
     rather than as tiny km fractions.
+
+    ``max_order`` must lie in 0..MAX_ORDER; the km/m/cm ladder bottoms out
+    at cm, which only stays legible while resolutions stay at or above the
+    order-29 floor.
     '''
+    if not 0 <= max_order <= MAX_ORDER:
+        raise ValueError(
+            'max_order must be between 0 and %d, got %r'
+            % (MAX_ORDER, max_order))
     for res in range(max_order + 1):
         km = order2res(res)
         if km >= 1.0:
--- a/mortie/tools.py
+++ b/mortie/tools.py
@@ -35,14 +35,12 @@ def res2display(max_order=MAX_ORDER):
     naturally (e.g. order 12 -> ``1.589 km``, order 13 -> ``794.456 m``)
     rather than as tiny km fractions.
 
-    ``max_order`` must lie in 0..MAX_ORDER; the km/m/cm ladder bottoms out
-    at cm, which only stays legible while resolutions stay at or above the
-    order-29 floor.
+    ``max_order`` must lie in 0..MAX_ORDER, the order range the packed-u64
+    kernel encodes.
     '''
     if not 0 <= max_order <= MAX_ORDER:
         raise ValueError(
-            'max_order must be between 0 and %d, got %r'
-            % (MAX_ORDER, max_order))
+            f"max_order must be between 0 and {MAX_ORDER}, got {max_order!r}")
     for res in range(max_order + 1):
         km = order2res(res)
         if km >= 1.0:
--- a/mortie/tools.py
+++ b/mortie/tools.py
@@ -24,10 +24,27 @@ def order2res(order):
     return res
 
 
-def res2display():
-    '''prints resolution levels'''
-    for res in range(20):
-        print(str(order2res(res)) + ' km at tessellation order ' + str(res))
+def res2display(max_order=MAX_ORDER):
+    '''prints resolution levels
+
+    Prints one line per tessellation order from 0 through ``max_order``
+    (default MAX_ORDER = 29, the finest order the packed-u64 kernel reaches).
+    Each resolution is rendered in the largest sensible unit -- km at coarse
+    orders, m once it drops below 1 km, cm once it drops below 1 m -- and
+    rounded to three decimals within that bracket, so fine orders read
+    naturally (e.g. order 12 -> ``1.589 km``, order 13 -> ``794.456 m``)
+    rather than as tiny km fractions.
+    '''
+    for res in range(max_order + 1):
+        km = order2res(res)
+        if km >= 1.0:
+            value, unit = km, 'km'
+        elif km >= 1e-3:
+            value, unit = km * 1e3, 'm'
+        else:
+            value, unit = km * 1e5, 'cm'
+        print(str(round(value, 3)) + ' ' + unit +
+              ' at tessellation order ' + str(res))
 
 
 def unique2parent(unique):
--- a/mortie/tests/test_tools.py
+++ b/mortie/tests/test_tools.py
@@ -86,7 +86,13 @@ class TestRes2Display:
         for line in lines:
             number = line.split(' ', 1)[0]
             # at most three digits after the decimal point
-            assert len(number.split('.')[1]) <= 3
+            assert '.' in number and len(number.split('.')[1]) <= 3
+
+    def test_out_of_range_raises(self):
+        """max_order outside 0..MAX_ORDER is rejected"""
+        for bad in (-1, tools.MAX_ORDER + 1):
+            with pytest.raises(ValueError, match="max_order must be"):
+                tools.res2display(max_order=bad)
 
 
 class TestUnique2Parent:
--- a/mortie/tests/test_tools.py
+++ b/mortie/tests/test_tools.py
@@ -45,6 +45,50 @@ class TestOrder2Res:
         assert all(resolutions[i] > resolutions[i+1] for i in range(len(resolutions)-1))
 
 
+class TestRes2Display:
+    """Test the resolution-ladder printer"""
+
+    def test_prints_all_orders_through_max(self, capsys):
+        """One line per order 0..MAX_ORDER -- no silent drop above 19"""
+        tools.res2display()
+        lines = capsys.readouterr().out.strip().split('\n')
+        assert len(lines) == tools.MAX_ORDER + 1
+        # every order 0..29 appears with its trailing phrasing
+        for order in range(tools.MAX_ORDER + 1):
+            assert any(
+                line.endswith('at tessellation order ' + str(order))
+                for line in lines
+            )
+
+    def test_max_order_argument(self, capsys):
+        """max_order bounds the printed range inclusively"""
+        tools.res2display(max_order=5)
+        lines = capsys.readouterr().out.strip().split('\n')
+        assert len(lines) == 6
+        assert lines[-1].endswith('at tessellation order 5')
+
+    def test_unit_ladder_km_m_cm(self, capsys):
+        """Coarse orders read in km, sub-km in m, sub-m in cm"""
+        tools.res2display()
+        lines = capsys.readouterr().out.strip().split('\n')
+        by_order = {int(line.rsplit(' ', 1)[1]): line for line in lines}
+        # order 12 = 1.589 km, order 13 = 794.456 m (the issue's examples)
+        assert by_order[12] == '1.589 km at tessellation order 12'
+        assert by_order[13] == '794.456 m at tessellation order 13'
+        # finest orders drop to cm rather than tiny km/m fractions
+        assert by_order[25] == '19.396 cm at tessellation order 25'
+        assert by_order[29] == '1.212 cm at tessellation order 29'
+
+    def test_rounds_within_bracket(self, capsys):
+        """Values are rounded to three decimals inside the chosen unit"""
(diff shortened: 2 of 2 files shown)


mortie/tools.py
```diff
@@ -24,10 +24,27 @@ def order2res(order):
...
+
+    Prints one line per tessellation order from 0 through ``max_order``
+    (default MAX_ORDER = 29, the finest order the packed-u64 kernel reaches).
+    Each resolution is rendered in the largest sensible unit -- km at coarse
+    orders, m once it drops below 1 km, cm once it drops below 1 m -- and
+    rounded to three decimals within that bracket, so fine orders read
+    naturally (e.g. order 12 -> ``1.589 km``, order 13 -> ``794.456 m``)
+    rather than as tiny km fractions.
+    '''
+    for res in range(max_order + 1):
```
espg, line 44: 🤖 *from Claude (review)*

`max_order` is unvalidated at both ends. Two edge cases neither guarded nor tested:

- **Negative** `max_order` → `range(n<1)` is empty, so the function silently prints nothing (e.g. `res2display(-1)` returns with no output, no error). Silent no-op on bad input is easy to misread as "nothing to display".
- **`max_order > 29`** → happily prints beyond the kernel's `MAX_ORDER`, contradicting the docstring's "29 = the finest order the kernel reaches". Worse, past order ~32 the cm bracket itself drops below 1 (`0.152 cm`, `0.019 cm`, …) — there's no mm rung, so it degrades to sub-1 cm fractions, the exact readability problem this PR set out to fix.

Suggest clamping/validating to `0..MAX_ORDER` (e.g. raise `ValueError` on out-of-range, or `min(max_order, MAX_ORDER)`), and add a test pinning the chosen behavior. Not a blocker for the sanctioned scope, but the docstring currently over-promises vs. the code.

---

mortie/tools.py
```diff
@@ -24,10 +24,27 @@ def order2res(order):
...
+        km = order2res(res)
+        if km >= 1.0:
+            value, unit = km, 'km'
+        elif km >= 1e-3:
+            value, unit = km * 1e3, 'm'
+        else:
+            value, unit = km * 1e5, 'cm'
+        print(str(round(value, 3)) + ' ' + unit +
```
espg, line 52: 🤖 *from Claude (review)*

Nit / latent: bracket selection uses the raw `km`, but the printed number is `round(value, 3)` *after* conversion. A value just under a boundary can round up across it and print in the smaller unit — e.g. `km = 0.9999996` selects the `m` bracket and prints `1000.0 m` instead of `1.0 km` (same for `1000.0 cm`). No order in `0..29` triggers this, so it's harmless today, but it's a real inconsistency if the range ever widens or `order2res` constants change. Worth a one-line comment or a post-round bracket re-check.

---

espg (commented):
🤖 *from Claude (review)*

Reviewed the diff for #99 against origin/main. Core fix is correct and well-grounded: the km/m/cm ladder cutoffs (`>=1.0` km, `>=1e-3` m, else cm), `range(max_order+1)` inclusivity, and round-to-3 are all right, and the golden values match the issue examples exactly (order 12 → `1.589 km`, 13 → `794.456 m`, 25 → `19.396 cm`, 29 → `1.212 cm`). Within the sanctioned range `0..29` there are no bracket-boundary or sub-1-cm surprises. Tests are a solid addition — order count, `max_order` bound, unit ladder, and rounding are all covered. Terse, matches surrounding style; module is 860 lines (under the ~1000 limit); no new deps. No blocking issues.

Findings (all Low / non-blocking, for @espg to fold in or wave off):

1. **Low — unvalidated `max_order` at both ends.** Negative → silent empty output; `>29` → prints beyond-kernel orders and, past ~32, sub-1 cm fractions with no mm rung, contradicting the docstring. Inline on the loop. Consider clamp-or-raise + a test.
2. **Nit — round-after-convert can cross a bracket boundary** (`0.9999996 km` → `1000.0 m`). Not triggered by any real order; latent if the range widens. Inline.
3. **Nit — `test_rounds_within_bracket` does `number.split('.')[1]`** which would `IndexError` if a value ever printed without a decimal point. Safe for all current values (Python `round(float,3)` always yields a dotted str), but brittle; a `'.' in number` guard or asserting the format directly would be sturdier.

Docstring is otherwise accurate for the default. Nothing here needs to block landing the phase.

---

mortie/tools.py
```diff
@@ -24,10 +24,35 @@ def order2res(order):
...
+    rather than as tiny km fractions.
+
+    ``max_order`` must lie in 0..MAX_ORDER; the km/m/cm ladder bottoms out
+    at cm, which only stays legible while resolutions stay at or above the
+    order-29 floor.
+    '''
+    if not 0 <= max_order <= MAX_ORDER:
+        raise ValueError(
+            'max_order must be between 0 and %d, got %r'
+            % (MAX_ORDER, max_order))
```

Validation logic is correct: `0 <= max_order <= MAX_ORDER` accepts both boundaries (0 and 29) and rejects -1 / 30, and `%d`/`%r` are safe (`%r` reprs any type without a formatting error). No functional objection.

Minor, non-blocking — style: this is the module's only `%`-formatted raise. Every other interpolated message uses an f-string (e.g. `f"Mixed orders in morton array: {set(...)}"` at line 289, and `__init__.py`/`arrow.py`/`coverage.py` follow the same convention). An f-string here matches the surrounding code and collapses three lines to one:

```python
if not 0 <= max_order <= MAX_ORDER:
    raise ValueError(f"max_order must be between 0 and {MAX_ORDER}, got {max_order!r}")
```

(coverage.py's `raise ValueError("Order must be between 1 and 29")` is the plain-literal precedent; nothing in the module uses `%`.)

---

mortie/tools.py
```diff
@@ -24,10 +24,35 @@ def order2res(order):
...
+    (default MAX_ORDER = 29, the finest order the packed-u64 kernel reaches).
+    Each resolution is rendered in the largest sensible unit -- km at coarse
+    orders, m once it drops below 1 km, cm once it drops below 1 m -- and
+    rounded to three decimals within that bracket, so fine orders read
+    naturally (e.g. order 12 -> ``1.589 km``, order 13 -> ``794.456 m``)
+    rather than as tiny km fractions.
+
+    ``max_order`` must lie in 0..MAX_ORDER; the km/m/cm ladder bottoms out
+    at cm, which only stays legible while resolutions stay at or above the
+    order-29 floor.
```

Nit, non-blocking — docstring rationale: "the km/m/cm ladder bottoms out at cm, which only stays legible while resolutions stay at or above the order-29 floor" justifies the upper bound by display *legibility*, but the actual reason `max_order` caps at `MAX_ORDER` is that 29 is the packed-u64 kernel's max order (the same limit already documented on the `MAX_ORDER` constant at line 16-19). The legibility framing reads as the cause when it's really a side effect. Consider tying the cap to the kernel limit instead, and trimming — the paragraph is a touch verbose for a print helper.

---

espg (commented):
🤖 *from Claude (review)*

Phase-2 review (commit 794b89d — validation + docstring + `test_out_of_range_raises`, and the `'.' in number` test guard). Reviewed with fresh context against origin/main.

**Verdict: correct and safe. No blocking issues.** Two minor, non-blocking notes inline.

What I verified:
- **Boundary inclusivity — correct.** `not 0 <= max_order <= MAX_ORDER` accepts 0 and 29 and rejects -1 and 30. Confirmed by executing the predicate directly.
- **Error message format — safe.** `%d` on `MAX_ORDER` (int) and `%r` on `max_order` won't raise a formatting error for any input type.
- **New test exercises both ends.** `test_out_of_range_raises` loops `(-1, MAX_ORDER + 1)`, hitting the just-below and just-above rejections, and pins the message via `match="max_order must be"`. Acceptance of the 0/29 boundaries isn't asserted in this test, but it's covered indirectly (default run reaches 29; `range()` covers 0) — fine as is.
- **`'.' in number` guard** converts a potential `IndexError` (on a value with no decimal point) into a clean assertion. Harmless robustness improvement; `str(round(float, 3))` always contains a `.` at the orders in play, so it never fires today.
- **No new dependencies**, no scope creep, module-size fine.

Minor notes (inline): (1) the raise is the module's only `%`-formatted message — an f-string matches the surrounding convention and is terser; (2) the docstring justifies the 0..MAX_ORDER cap by display legibility rather than the kernel's order-29 limit, which is the real reason.

One observation, no action needed: validation guards *range* but not *type* — a float like `2.5` slips past `0 <= 2.5 <= 29` and then fails later in `range(max_order + 1)` with a bare `TypeError`, and a str fails in the comparison itself. Out of scope for this fix; flagging only for completeness.

---

mortie/tools.py
espg: 🤖 *from Claude*

Left standing intentionally. As the review notes, no order in the sanctioned `0..MAX_ORDER` range lands within rounding distance of a bracket boundary (the closest is order 12 at `1.589 km`, nowhere near `1.000`), so this can't fire today. A post-round bracket re-check would add branching to a print helper for a case that only becomes reachable if `order2res`'s constants change or the range widens past 29 — both of which are now guarded against by the phase-2 `ValueError`. Flagging it here rather than folding it in; happy to add the re-check if you'd prefer the belt-and-suspenders version.

---

Issue #62, opened 2026-06-25
v1.0 docs: specification & conventions page (bit-layout + resolution/Δ table + frozen-convention statement)


espg:
🤖 *from Claude*

Sub-issue of #48 (v1.0 sweep), Tier-2 docs.

## Scope
The user-facing spec that the 1.x stability contract points at:

- **Bit-layout** of the packed-u64 morton index (`[4-bit prefix | 54-bit body | 6-bit suffix]`, base-cell prefix, 27×2-bit body, 6-bit suffix carrying orders 0/28/29) — sourced from `src_rust/src/decimal_morton.rs` module docs.
- **Resolution / Δ table** mapping order 0–29 → HEALPix `nside`, approximate cell edge length / area (the README "resolution table" espg called for).
- **Conventions**: unsigned packed storage, the decimal string as a render-only repr (signed repr kept per the #58 design record), Z-order = raw-word sort, base-cell prefix `base_id + 1`.
- An explicit **"frozen for 1.x"** statement of what the contract guarantees (order 0–29, indexing convention immutable within a major version).

## Acceptance criteria
A markdown spec page (under `docs/`), linked from the README; resolution table generated from `order2res` so it can't drift.

Refs #48.

espg:
🤖 *from Claude*

**Scope addition from the hive-layout design discussion** (the #48 repr thread, continued interactively; full design record in zagg's [`docs/design/sparse_coverage.md`](https://github.com/englacial/zagg/blob/main/docs/design/sparse_coverage.md), tracked at englacial/zagg#198). The spec page this issue tracks gains a **hive layout & interchange section** — mortie owns the layout *convention* (it's part of the frozen 1.x contract); zagg owns the reader/xarray side. Items to fold in:

1. **Hive path schedule**: `{store_root}/{sign+base}/{d1}/…/{full_id}.zarr` — one decimal digit per path level down to shard order; full morton id at the leaf. Rationale: shards live at mixed orders, so every order must be a legal node; sign+base is constant-width so string-prefix = spatial ancestor at every level.
2. **Node invariant**: below root, a tree node contains only digit children (`[1-4]/`) and `*.zarr` objects — nothing else, ever (the walker's child classification depends on it).
3. **Termination condition**: object stores have no empty prefixes and S3 LIST is strongly consistent, so a delimiter-LIST with no digit children is a definitive "nothing finer exists."
[...]

PR #105, opened 2026-07-10
Decimal-string display & casting layer for morton_index

### LINKED ISSUE #104: Decimal-string display & casting layer: element repr, vectorized to_decimal, scalar wrapper, legacy i64 emit, hive_path
🤖 *from Claude*

Implementation issue for the repr/casting decisions settled on #48 (the [zagg worked example](https://github.com/espg/mortie/issues/48#issuecomment-4757339511) thread and follow-on discussion). The Rust layer is done and pinned (`to_decimal_repr` in `src_rust/src/decimal_morton.rs`, golden fixtures, `from_legacy_decimal`); what's missing is the user-visible surface, so printed shard ids downstream (zagg) still resolve to raw packed-word integers.

## Scope

1. **Element repr = the decimal string.** `MortonIndexArray._word_repr` / `_formatter` / `__repr__` (`mortie/morton_index.py:437-458`) currently print `base=X order=Y` labels. Switch element rendering to the decode-through-kernel decimal string (`-31123` style); keep the `len=N, order=K` summary in the array-level header.
2. **Vectorized string emit.** `decimal_repr()` returns `list[str]`; add a vectorized emit returning a numpy fixed-width string array (`<U31` covers sign + base + 29 digits). Naming (`to_decimal()` vs extending `decimal_repr`) per implementation.
3. **Scalar wrapper.** Element access/iteration currently yields bare `numpy.uint64`, so any downstream `f"{shard_key}"` prints the packed word. Provide a lightweight scalar whose `__str__`/`__repr__` is the decimal string and `__int__` is the word.
4. **`to_legacy_i64()`.** Explicit, named, capped emit of the signed decimal `int64` for orders ≤ 18, **hard error above** — for testing against old pinned outputs (pairs with the existing `from_legacy`). Never a default, never data-dependent: the emit-conventions decision from #48 is *always strings* for display/interchange, packed `uint64` for storage, legacy i64 only via this named escape hatch.
5. **`hive_path()` helper + inverse parser.** Path arithmetic for the hive layout convention (one digit per level, full-id leaf — spec lands on #62; design record in zagg's `docs/design/sparse_coverage.md`, englacial/zagg#198). Small: string slicing over the decimal repr, but it belongs next to the repr it depends on.
6. **Golden fixtures** pinning the formatter/emit output (the string layer joins the frozen 1.x contract), plus a documented note that the repr is not injective across kinds (an order-29 *point* renders identically to the order-29 *area* on the same path).

## Sequencing

zagg pins mortie from PyPI (`mortie>=0.8.5`), so englacial/zagg's shard-id/layout migration is blocked until this lands **and a release is tagged** (tag = production action, not the routine's).

espg:
Closes #104. Refs #48 (emit-conventions decisions), #62 (spec page the hive convention lands on).

## What this does

The user-visible half of the repr/casting layer settled on #48: printed shard ids downstream (zagg) currently resolve to raw packed-word integers because the `morton_index` skin never renders the decode-through-kernel decimal string. The Rust layer (`to_decimal_repr`, `rust_mi_decimal_repr`, `from_legacy_decimal`) is already merged and pinned; this PR wires the Python surface to it.

## Phases

- [x] **Phase 1 — element display = the decimal string.** `MortonIndexArray._word_repr` / `_formatter` / `__repr__` now render each element as its decimal Morton id (`-31123` style) instead of the old `base=X order=Y` label; the `len=N, order=K` summary stays in the array-level header. Adds the lightweight scalar wrapper `MortonIndexScalar(np.uint64)`, returned from element access/iteration: `__str__`/`__repr__`/`__format__` are the decimal string, `int()` is the packed word, and comparisons/hashing stay word-valued. The empty sentinel renders `<NA>`; an invalid prefix renders `<invalid 0x...>` rather than raising (a repr must never raise). `__format__` is overridden because numpy's numeric formatting would otherwise print the packed word through `f"{shard_key}"` — exactly the downstream case the issue calls out.
- [x] **Phase 2 — vectorized emit + legacy escape hatch.** `to_decimal()` returns the decimal strings as a fixed-width numpy `<U31` array (sign + base digit + 29 order digits is the widest form, so the width is order-independent and stable across arrays). `to_legacy_i64()` emits the signed decimal `int64` for orders ≤ 18 with a **hard `ValueError` above** (pairs with the existing `from_legacy`; a mixed array with any element above 18 errors — never truncated, never data-dependent, per the #48 emit conventions).
- [x] **Phase 3 — `hive_path()` + inverse parser.** `hive_path(root="", suffix=".zarr")` emits `{root}/{sign+base}/{d1}/…/{d_order}/{full_id}.zarr` (one digit per level, full-id leaf inside its own node; order-0 `-3` → `-3/-3.zarr`), so mixed shard orders nest naturally — every order is a node. `from_hive_path` parses the leaf basename back to words via the new `_decimal_to_word` (full validation: base digit 1–6, order digits 1–4, order ≤ 29) and cross-checks the digit directories against the leaf id when the chain is present.
- [x] **Phase 4 — golden fixtures.** The string layer is pinned through the existing `packed_u64_golden.json` (all 1,080 records): `to_decimal`/`_word_repr`/scalar `str` equal the pinned `decimal_repr` strings, `to_legacy_i64` equals the pinned decimal read as an integer for orders 0–18, and `hive_path`→`from_hive_path` round-trips every fixture word, with literal layout pins for representative ids. Plus the documented non-injectivity test: an order-29 *point* renders identically to the *area* on the same path (words differ, strings don't, and parsing the string lands on the area word).
- [x] **Phase 5 — folded self-review findings (round 1).** (a) `MortonIndexScalar.__reduce__` so pickling preserves the wrapper across process boundaries (multiprocessing/dask) + pinning test; (b) tests pinning `__format__` for string specs, numeric specs (raise), and old-style `%`-formatting (bypasses `__format__`, emits the raw word — documented in-line); (c) `from_hive_path` accepts a single `os.PathLike`; (d) first cut of the directory cross-check anchor. The NA-boxing and per-call-FFI review notes are answered on their threads as recorded decisions.
- [x] **Phase 6 — folded self-review findings (round 2).** The converging review demonstrated phase 5's anchor (equality with the leaf's own sign+base) regressed the wrong-base case (`-4/1/1/2/3/-31123.zarr` silently parsed). The anchor is now the sign+base **shape** (optional `-` + digit 1..6) at its slot: wrong-base and wrong-descent mis-filings raise again (pinned), arbitrary roots still skip, and a real root above a full chain parses (pinned).
- [x] **Phase 7 — Windows-flavored paths.** `from_hive_path` splits `PurePath` inputs via their own `.parts` (a `WindowsPath` honors backslashes; raw strings keep exact `"/"`-split semantics), pinned with a `PureWindowsPath` round-trip that runs on the ubuntu CI.

## How it was tested

- `maturin develop --release` (no Rust change; ext rebuilt at current `main`), `flake8 mortie --select=E9,F63,F7,F82` clean; the 88-col style pass is clean on all touched files.
- `pytest -v` — **578 passed, 11 skipped** (30 new: `TestDecimalDisplay`, `TestHivePath`, and 4 golden string-layer pins in `test_packed_golden.py`).
- Three adversarial self-review rounds ran (phase 1; phases 2–4; phase 5) and converged — every diff-scoped finding is folded (phases 5–7) or answered on its thread as a recorded decision.

## Questions for review

- **Hive path shape:** the zagg design record (`docs/design/sparse_coverage.md`) gives `{sign+base}/{d1}/{d2}/…/` with a `{full_id}.zarr` leaf; with "every order is a node" I read the leaf as sitting *inside its own* digit directory — `-31123` → `-3/1/1/2/3/-31123.zarr`, and order-0 `-3` → `-3/-3.zarr`. If you read the ellipsis as stopping at `d_{order-1}` instead, say so and I'll flip it (it's one line + the pinned fixtures).
- `__format__` formats the *decimal string* under any format spec (so `f"{key:>10}"` pads the string; numeric specs like `:d` raise — `int(key)` is the escape; `"%d" % key` bypasses `__format__` and emits the raw word). All three behaviors are pinned by tests; flagging in case you'd rather have numeric specs fall through to the word.
- `__getitem__` on a missing entry returns `MortonIndexScalar(0)` (renders `<NA>`, but `pd.isna()` is `False`) — pre-existing deviation from the strict ExtensionArray contract, kept deliberately; see the review thread if you want it switched to `pd.NA`.
- `from_hive_path`'s anchored cross-check accepts that a root component coincidentally shaped like `{sign+base}` at exactly its slot with a non-matching descent raises (strict-on-suspicious); flip to lenient if you'd rather never false-reject.

### DIFF
--- a/mortie/morton_index.py
+++ b/mortie/morton_index.py
@@ -30,6 +30,40 @@ __all__ = ["MortonIndexScalar"]
 MAX_ORDER = 29
 
 
+def _decimal_to_word(s):
+    """Parse a decimal Morton string back to its packed word (issue #104).
+
+    The inverse of the decode-through-kernel repr: sign + leading base digit
+    (``1..6``) then one ``1..4`` digit per order. Raises ``ValueError`` on any
+    malformed id. Order-29 strings parse to the *area* word -- the point/area
+    flag is not part of the decimal form, so the point word is unreachable
+    from its repr (the documented non-injectivity).
+    """
+    body = s[1:] if s.startswith("-") else s
+    if not (body.isdigit() and body.isascii()):
+        raise ValueError(f"malformed decimal Morton id {s!r}")
+    lead, digits = int(body[0]), body[1:]
+    if not 1 <= lead <= 6:
+        raise ValueError(
+            f"decimal Morton id {s!r}: base digit {lead} outside 1..6"
+        )
+    order = len(digits)
+    if order > MAX_ORDER:
+        raise ValueError(
+            f"decimal Morton id {s!r}: order {order} exceeds {MAX_ORDER}"
+        )
+    within = 0
+    for d in digits:
+        if not "1" <= d <= "4":
+            raise ValueError(
+                f"decimal Morton id {s!r}: digit {d} outside 1..4"
+            )
+        within = (within << 2) | (int(d) - 1)
+    base = lead + 5 if s.startswith("-") else lead - 1
+    nested = np.asarray([(base << (2 * order)) | within], dtype=np.uint64)
+    return int(_rustie.rust_mi_from_nested(nested, order)[0])
+
+
 class MortonIndexScalar(np.uint64):
     """A packed ``morton_index`` word that displays as its decimal string.
 
@@ -501,6 +535,63 @@ def _build_classes():
                     f"{int(orders.max())}); use to_decimal() for orders 19-29"
                 )
             return np.asarray([int(s) for s in strings], dtype=np.int64)
+
+        def hive_path(self, root="", suffix=".zarr"):
+            """Hive-layout path per element (issue #104; spec lands on #62).
+
+            The ``morton-hive/1`` convention (zagg's sparse-coverage design
+            record): one decimal digit per directory level, the full id as the
+            leaf inside its own node --
+            ``{root}/{sign+base}/{d1}/.../{d_order}/{full_id}{suffix}``, e.g.
+            ``-31123`` -> ``-3/1/1/2/3/-31123.zarr`` and the order-0 ``-3`` ->
+            ``-3/-3.zarr``. Every order is a node, so mixed shard orders nest
+            naturally: a coarser shard's leaf sits in the same directory its
+            finer siblings descend through. Returns a list of ``str``; raises
+            ``ValueError`` on any empty / invalid word.
+            """
+            prefix = root.rstrip("/") + "/" if root else ""
+            paths = []
+            for s in self.decimal_repr():
+                head = 2 if s.startswith("-") else 1
+                levels = "/".join([s[:head], *s[head:]])
+                paths.append(f"{prefix}{levels}/{s}{suffix}")
+            return paths
+
+        @classmethod
+        def from_hive_path(cls, paths, suffix=".zarr"):
+            """Parse hive-layout paths back to words (inverse of hive_path).
+
+            ``paths`` is a single path (``str``) or an iterable of them. The
+            leaf basename carries the full decimal id; when the path also
+            carries the digit directories, they are checked against the leaf
+            and a mis-filed leaf raises ``ValueError`` (a shorter path -- a
+            bare ``{full_id}.zarr`` or one under an arbitrary root -- skips
+            the check for the components it does not have). Order-29 ids
+            parse to the *area* word (see :func:`_decimal_to_word`).
+            """
+            if isinstance(paths, str):
+                paths = [paths]
+            words = []
+            for p in paths:
+                parts = str(p).rstrip("/").split("/")
+                leaf = parts[-1]
+                if suffix and not leaf.endswith(suffix):
+                    raise ValueError(
+                        f"hive leaf {leaf!r} does not end with {suffix!r}"
+                    )
+                dec = leaf[: len(leaf) - len(suffix)] if suffix else leaf
+                word = _decimal_to_word(dec)
+                head = 2 if dec.startswith("-") else 1
+                levels = [dec[:head], *dec[head:]]
+                if len(parts) - 1 >= len(levels):
+                    got = parts[-1 - len(levels):-1]
+                    if got != levels:
+                        raise ValueError(
+                            f"hive path {p!r} directories {'/'.join(got)!r} "
+                            f"do not match leaf id {dec!r}"
+                        )
+                words.append(word)
+            return cls(np.asarray(words, dtype=np.uint64))
 
         # -- repr ------------------------------------------------------------
 
--- a/mortie/morton_index.py
+++ b/mortie/morton_index.py
@@ -17,6 +17,8 @@ numpy installed. The pandas machinery here is built lazily on first use, and a
 clear ``ImportError`` is raised if the ExtensionArray is touched without pandas.
 """
 
+import os
+
 import numpy as np
 
 from . import _rustie
@@ -93,8 +95,15 @@ class MortonIndexScalar(np.uint64):
         # numpy's numeric __format__ would print the packed word; the display
         # form of a morton_index is its decimal string, so f"{shard_key}" (and
         # any string spec, e.g. ">10") formats that instead. int(self) remains
-        # the escape hatch to format the raw word numerically.
+        # the escape hatch to format the raw word numerically. Old-style
+        # "%d" % key bypasses __format__ entirely and emits the raw word.
         return format(str(self), spec)
+
+    def __reduce__(self):
+        # numpy scalars pickle through multiarray.scalar, which rebuilds the
+        # bare np.uint64 and would silently drop the decimal display on any
+        # process boundary (multiprocessing/dask); rebuild the wrapper instead.
+        return (type(self), (int(self),))
 
 
 def _require_pandas():
@@ -561,15 +570,17 @@ def _build_classes():
         def from_hive_path(cls, paths, suffix=".zarr"):
             """Parse hive-layout paths back to words (inverse of hive_path).
 
-            ``paths`` is a single path (``str``) or an iterable of them. The
-            leaf basename carries the full decimal id; when the path also
-            carries the digit directories, they are checked against the leaf
-            and a mis-filed leaf raises ``ValueError`` (a shorter path -- a
-            bare ``{full_id}.zarr`` or one under an arbitrary root -- skips
-            the check for the components it does not have). Order-29 ids
-            parse to the *area* word (see :func:`_decimal_to_word`).
+            ``paths`` is a single path (``str`` / ``os.PathLike``) or an
+            iterable of them. The leaf basename carries the full decimal id;
+            when the path also carries the digit directories -- recognized by
+            the ``{sign+base}`` component sitting at its slot above the leaf
+            -- the whole chain is checked against the leaf and a mis-filed
+            leaf raises ``ValueError``. A bare ``{full_id}.zarr``, or one
+            under an arbitrary root without the digit chain, skips the check.
+            Order-29 ids parse to the *area* word (see
+            :func:`_decimal_to_word`).
             """
-            if isinstance(paths, str):
+            if isinstance(paths, (str, os.PathLike)):
                 paths = [paths]
             words = []
             for p in paths:
@@ -583,7 +594,14 @@ def _build_classes():
                 word = _decimal_to_word(dec)
                 head = 2 if dec.startswith("-") else 1
                 levels = [dec[:head], *dec[head:]]
-                if len(parts) - 1 >= len(levels):
+                # Enforce the directory cross-check only when the chain is
+                # anchored: the {sign+base} component at its expected slot.
+                # (Root components are indistinguishable from digit dirs by
+                # count alone, so an unanchored tail is treated as root.)
+                if (
+                    len(parts) - 1 >= len(levels)
+                    and parts[-1 - len(levels)] == levels[0]
+                ):
                     got = parts[-1 - len(levels):-1]
                     if got != levels:
                         raise ValueError(
--- a/mortie/morton_index.py
+++ b/mortie/morton_index.py
@@ -24,10 +24,43 @@ from . import _rustie
 # ``MortonIndexDtype`` / ``MortonIndexArray`` are provided via module-level
 # ``__getattr__`` (built lazily so a numpy-only install can import this module),
 # so they are intentionally not named in ``__all__`` here.
-__all__ = []
+__all__ = ["MortonIndexScalar"]
 
 # HEALPix orders this datatype reaches (0 = base cell, 29 = max resolution).
 MAX_ORDER = 29
+
+
+class MortonIndexScalar(np.uint64):
+    """A packed ``morton_index`` word that displays as its decimal string.
+
+    Element access and iteration on a ``MortonIndexArray`` yield this type
+    (issue #104), so a downstream ``f"{shard_key}"`` prints the decimal Morton
+    id (``-31123`` style) rather than the raw packed word. It subclasses
+    ``numpy.uint64``: comparisons, hashing, and ``int()`` (the packed word)
+    behave exactly like the word itself; only ``str``/``repr`` differ. The
+    empty sentinel renders ``"<NA>"``; a word with an invalid prefix renders
+    ``"<invalid 0x...>"`` rather than raising (a repr must never raise).
+    """
+
+    def __str__(self):
+        word = int(self)
+        if word == 0:
+            return "<NA>"
+        try:
+            return _rustie.rust_mi_decimal_repr(
+                np.asarray([word], dtype=np.uint64)
+            )[0]
+        except ValueError:
+            return f"<invalid {word:#018x}>"
+
+    __repr__ = __str__
+
+    def __format__(self, spec):
+        # numpy's numeric __format__ would print the packed word; the display
+        # form of a morton_index is its decimal string, so f"{shard_key}" (and
+        # any string spec, e.g. ">10") formats that instead. int(self) remains
+        # the escape hatch to format the raw word numerically.
+        return format(str(self), spec)
 
 
 def _require_pandas():
@@ -272,7 +305,7 @@ def _build_classes():
         def __getitem__(self, item):
             result = self._data[item]
             if np.isscalar(result) or isinstance(result, np.integer):
-                return np.uint64(result)
+                return MortonIndexScalar(result)
             return type(self)(result)
 
         def __setitem__(self, key, value):
@@ -435,13 +468,8 @@ def _build_classes():
         # -- repr ------------------------------------------------------------
 
         def _word_repr(self, word):
-            """Compact ``base/order`` label for one packed word."""
-            w = np.asarray([word], dtype=np.uint64)
-            if word == int(self._SENTINEL):
-                return "<NA>"
-            base = int(_rustie.rust_mi_base_cell_of(w)[0])
-            order = int(_rustie.rust_mi_order_of(w)[0])
-            return f"base={base} order={order}"
+            """Decimal-string label for one packed word (issue #104)."""
+            return str(MortonIndexScalar(word))
 
         def __repr__(self):
             n = len(self)
--- a/mortie/morton_index.py
+++ b/mortie/morton_index.py
@@ -581,11 +581,18 @@ def _build_classes():
             Order-29 ids parse to the *area* word (see
             :func:`_decimal_to_word`).
             """
+            import pathlib
+
             if isinstance(paths, (str, os.PathLike)):
                 paths = [paths]
             words = []
             for p in paths:
-                parts = str(p).rstrip("/").split("/")
+                if isinstance(p, pathlib.PurePath):
+                    # honor the path's own flavor: a WindowsPath splits on
+                    # backslashes too, which a raw "/"-split would not
+                    parts = list(p.parts)
+                else:
+                    parts = str(p).rstrip("/").split("/")
                 leaf = parts[-1]
                 if suffix and not leaf.endswith(suffix):
                     raise ValueError(
--- a/mortie/morton_index.py
+++ b/mortie/morton_index.py
@@ -461,9 +461,46 @@ def _build_classes():
             produced by *decoding* each word (the canonical render-only repr;
             backward-compatible with the legacy ``str(legacy_i64)`` for orders
             0..=18, the natural extension for 19..=29). Raises ``ValueError`` on
-            any empty / invalid word.
+            any empty / invalid word. Note the repr is not injective across
+            kinds: an order-29 *point* renders identically to the order-29
+            *area* on the same path (the point/area flag is not part of the
+            decimal form).
             """
             return _rustie.rust_mi_decimal_repr(self._data)
+
+        def to_decimal(self):
+            """Vectorized decimal-string emit as a fixed-width numpy array.
+
+            The always-strings interchange convention from issue #48: returns
+            the decode-through-kernel decimal strings as a ``"<U31"`` numpy
+            array (sign + base digit + 29 order digits is the widest form, so
+            the width is order-independent and stable across arrays). Raises
+            ``ValueError`` on any empty / invalid word. Same non-injectivity
+            note as :meth:`decimal_repr`: point and area render identically at
+            order 29.
+            """
+            return np.asarray(self.decimal_repr(), dtype="<U31")
+
+        def to_legacy_i64(self):
+            """Emit the legacy signed decimal ``int64`` form (orders <= 18).
+
+            The named escape hatch from issue #48's emit conventions --
+            interchange is always the decimal *string* (:meth:`to_decimal`),
+            storage the packed ``uint64``; this exists solely for testing new
+            output against old pinned values (pairing with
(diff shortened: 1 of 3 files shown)


mortie/morton_index.py
```diff
@@ -24,12 +24,79 @@
...
+        if word == 0:
+            return "<NA>"
+        try:
+            return _rustie.rust_mi_decimal_repr(
+                np.asarray([word], dtype=np.uint64)
+            )[0]
+        except ValueError:
+            return f"<invalid {word:#018x}>"
+
+    __repr__ = __str__
```
espg, line 92: 🤖 *from Claude (review)*

**[low] Pickle drops the wrapper: a `MortonIndexScalar` comes back as a plain `np.uint64` and prints the raw packed word.**

Verified on this branch:

```python
>>> s = a[0]                      # MortonIndexScalar, str(s) == "-31123"
>>> p = pickle.loads(pickle.dumps(s))
>>> type(p).__name__, str(p)
('uint64', '10403315139225845764')
```

numpy scalars reduce to `multiarray.scalar(dtype, bytes)`, which rebuilds the base type (`copy.copy`/`deepcopy` do preserve the subclass — it is only pickle). So a shard key crossing a process boundary (multiprocessing/dask, i.e. exactly the zagg-style consumers this display layer targets per issue #104) silently reverts to raw-word display. A two-line `__reduce__` returning `(MortonIndexScalar, (int(self),))` restores it; worth a pinning test either way.

---

mortie/morton_index.py
```diff
@@ -24,12 +24,79 @@
...
+
+    def __format__(self, spec):
+        # numpy's numeric __format__ would print the packed word; the display
+        # form of a morton_index is its decimal string, so f"{shard_key}" (and
+        # any string spec, e.g. ">10") formats that instead. int(self) remains
+        # the escape hatch to format the raw word numerically.
+        return format(str(self), spec)
```
espg, line 100: 🤖 *from Claude (review)*

**[low] `__format__` is a behavioral change with no direct test.**

Nothing in the suite pins a nonempty format spec. Verified behavior on this branch:

- `f"{s:>10}"` → `'    -31123'` (works as the docstring claims)
- `f"{s:d}"` → `ValueError: Unknown format code 'd' for object of type 'str'` (the behavior the PR body flags as an open question)

Since the numeric-spec question is posed under "Questions for review", a test pinning whichever answer is chosen belongs in this PR (CLAUDE.md §4: every behavioral change needs tests in the same PR) — right now only the empty spec is exercised via `f"{s}"`.

Also verified but undocumented: old-style `%`-formatting bypasses `__format__` entirely — `"%d" % s` emits the raw word `10403315139225845764` while `f"{s}"` gives `-31123`. Worth one docstring line so that asymmetry reads as a documented choice rather than an oversight.

---

mortie/morton_index.py
```diff
@@ -272,7 +339,7 @@ def __len__(self):
         def __getitem__(self, item):
             result = self._data[item]
             if np.isscalar(result) or isinstance(result, np.integer):
-                return np.uint64(result)
+                return MortonIndexScalar(result)
```
espg, line 351: 🤖 *from Claude (review)*

**[note, pre-existing] The boxed missing element now *prints* as NA without *being* NA.**

`__getitem__` on a sentinel row returns `MortonIndexScalar(0)` rather than `dtype.na_value` (`pd.NA`). Verified:

```python
>>> s2  # Series with one missing entry
0    -31123
1      <NA>
dtype: morton_index
>>> pd.isna(s2.iloc[1])
False
```

The mismatch predates this diff (it returned `np.uint64(0)` before), but the new `<NA>` repr makes the scalar look like a missing marker that `pd.isna` / NA-propagation will not honor. The strict ExtensionArray contract is that `__getitem__` returns `self.dtype.na_value` for missing entries. If sentinel-boxing is deliberate (likely, given the all-zero-word design), fine — flagging so it is a recorded decision rather than an accident; out of scope to change in this phase.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of **phase 1** (commit 2b495d9: element display = decimal string, `MortonIndexScalar` wrapper). Phases 2 and 3 landed on the branch while this review was in flight and are **not** covered here — they need their own pass.

**Issue #104 scope items 1 and 3: met.** `_word_repr`/`_formatter`/`__repr__` render the decode-through-kernel decimal string, the `len=N, order=K` header stays, and element access *and* iteration return `MortonIndexScalar` with `str`/`repr` = decimal string and `int()` = packed word (all verified by running the branch).

Verified safe (claims I attacked and could not break):
- The `except ValueError` catch in `__str__` is total: `rust_mi_decimal_repr` maps every empty/invalid word to `PyValueError` via `Option` (`src_rust/src/lib.rs` L1111-1127) — no panic path, so "a repr must never raise" holds. `_word_repr(0)` still returns `<NA>` (pre-existing test passes).
- Arithmetic on the scalar degrades to plain `np.uint64` (`s + 1`, `-s`, `s * 2` all verified), so derived words never carry a misleading decimal repr.
- pandas paths: Series repr renders decimal strings with `<NA>` for missing; `to_csv` emits the decimal string (empty for NA rows); `astype(str)` gives decimal strings — consistent with the issue #48 always-strings interchange convention.
- Per-element FFI in `_word_repr` is bounded: at most 6 calls in `__repr__`; measured ~1.6 us/element (about 5x the vectorized `to_decimal()` path), fine for display.
- Conventions: flake8 blocking selectors and the 88-col style pass are clean, module was 516 lines at phase 1, no new dependencies, docstrings match the surrounding freeform style, tests ship in the same commit, and all 21 relevant tests pass locally.

Findings, ranked (inline comments):
1. **[low]** pickle drops the subclass — a shard key round-tripped through multiprocessing/dask reverts to raw-word display; `__reduce__` fixes it.
2. **[low]** `__format__` has no direct test (neither the working `>10` spec nor the documented `ValueError` on numeric specs), and `"%d" % s` silently emits the raw word — pin whichever numeric-spec answer is chosen.
3. **[note, pre-existing]** the boxed sentinel scalar prints `<NA>` but `pd.isna()` on it is `False`; flagged for a recorded decision, not a change.

Nothing blocking the next phase.

---

mortie/morton_index.py
```diff
@@ -428,20 +495,109 @@ def decimal_repr(self):
...
+                word = _decimal_to_word(dec)
+                head = 2 if dec.startswith("-") else 1
+                levels = [dec[:head], *dec[head:]]
+                if len(parts) - 1 >= len(levels):
+                    got = parts[-1 - len(levels):-1]
+                    if got != levels:
+                        raise ValueError(
+                            f"hive path {p!r} directories {'/'.join(got)!r} "
+                            f"do not match leaf id {dec!r}"
+                        )
```
espg, line 622: 🤖 *from Claude (review)*

**Medium — the directory cross-check false-positives on exactly the case the docstring promises to accept.** The docstring says a bare `{full_id}.zarr` "or one under an arbitrary root" skips the check, but the gate is component-*count*-based (`len(parts) - 1 >= len(levels)`), so any bare leaf whose root has at least `order + 1` components is misread as a full digit chain and raises. Verified against this commit:

```python
MIA.from_hive_path("-31123.zarr")                  # ok (documented)
MIA.from_hive_path("s3://bucket/-3.zarr")          # ValueError: directories 'bucket' do not match leaf id '-3'
MIA.from_hive_path("x/-3.zarr")                    # ValueError: directories 'x' do not match leaf id '-3'
MIA.from_hive_path("s3://bucket/data/-31123.zarr") # ok (root shorter than chain, check skipped)
```

Order-0 ids are the worst case: *any* root at all trips it. The flip side of the same heuristic is that partially-contradicting digit dirs pass silently (`MIA.from_hive_path("9/9/-31123.zarr")` parses fine), which the PR body flags — but the false-*reject* direction isn't flagged anywhere and contradicts the docstring as written.

Root components are inherently indistinguishable from digit dirs, so the count heuristic can't be made airtight; concrete options: (a) only run the check when the expected chain actually matches positionally, i.e. treat a mismatch at `parts[-1-len(levels)]` (the sign+base slot, which never looks like a bare `1..4` digit... except it can equal another id's base — so realistically) — (b) accept the ambiguity but make the behavior match the docs: keep the check strictly opt-in for paths produced by `hive_path` by requiring the *base-level* component `dec[:head]` to be present at its slot before enforcing, or (c) keep the code and rewrite the docstring to state the real rule ("checked whenever the path has at least order+1 leading components — do not pass bare leaves under a root"). Note `test_single_path_and_bare_leaf` only covers the zero-root bare leaf, so the promised under-a-root case is untested.

Related edge (same slice): a doubled slash inside an otherwise-correct full chain also raises (`"-3//1/1/2/3/-31123.zarr"` → mismatch on the empty component). Strict is defensible, just noting it falls out of the same arithmetic.

---

mortie/morton_index.py
```diff
@@ -428,20 +495,109 @@ def decimal_repr(self):
...
+            bare ``{full_id}.zarr`` or one under an arbitrary root -- skips
+            the check for the components it does not have). Order-29 ids
+            parse to the *area* word (see :func:`_decimal_to_word`).
+            """
+            if isinstance(paths, str):
+                paths = [paths]
+            words = []
+            for p in paths:
+                parts = str(p).rstrip("/").split("/")
+                leaf = parts[-1]
```

**Low — a single `pathlib.Path` input fails with a confusing `TypeError`.** The `str(p)` inside the loop suggests PathLike elements are intended to work (and they do, inside a list), but the singleton guard is `isinstance(paths, str)` only, so:

```python
MIA.from_hive_path(Path("-3/1/1/2/3/-31123.zarr"))
# TypeError: 'PosixPath' object is not iterable
```

Verified on this commit. Either widen the guard (`isinstance(paths, (str, os.PathLike))`) or say "str only" in the docstring — as written, "a single path (``str``)" plus per-element `str(p)` reads as if Path should work both ways.

---

mortie/morton_index.py
```diff
@@ -24,12 +24,79 @@
...
+    within = 0
+    for d in digits:
+        if not "1" <= d <= "4":
+            raise ValueError(
+                f"decimal Morton id {s!r}: digit {d} outside 1..4"
+            )
+        within = (within << 2) | (int(d) - 1)
+    base = lead + 5 if s.startswith("-") else lead - 1
+    nested = np.asarray([(base << (2 * order)) | within], dtype=np.uint64)
+    return int(_rustie.rust_mi_from_nested(nested, order)[0])
```
espg, line 66: 🤖 *from Claude (review)*

**Low (non-blocking) — one single-element Rust FFI call per parsed path.** `_decimal_to_word` packs one word per call, so `from_hive_path` pays per-element crossing overhead: measured on this commit, 100k paths parse in ~0.54 s vs ~0.08 s to emit them (`hive_path`) and ~0.03 s for `to_decimal`. Fine for the expected manifest sizes; if bulk parsing ever matters, grouping ids by order and batching `rust_mi_from_nested` per group is the obvious vectorization. Not asking for a change — just pinning the number so the tradeoff is a recorded decision rather than a surprise.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of phases 2–4 (commits c0c90fd..d0ec61a): `to_decimal`, `to_legacy_i64`, `hive_path`/`from_hive_path`/`_decimal_to_word`, and the string-layer golden pins. All claims below were verified by running the code at this head.

**Findings (3 inline):**
1. **Medium** — `from_hive_path`'s directory cross-check raises a false `ValueError` for a bare leaf under a root whenever the root has ≥ order+1 components (`"s3://bucket/-3.zarr"` fails; order-0 ids fail under *any* root), contradicting the docstring's "one under an arbitrary root skips the check". The loose direction (partial contradicting digit dirs accepted) is flagged in the PR body, but this false-reject direction is not.
2. **Low** — a single `pathlib.Path` argument to `from_hive_path` raises `TypeError: 'PosixPath' object is not iterable`.
3. **Low / non-blocking** — `from_hive_path` makes one single-element Rust call per path (~0.54 s / 100k vs ~0.08 s emit); recorded for the file.

**Probed and confirmed correct:** `_decimal_to_word` rejects Arabic-Indic/fullwidth/superscript digits, U+2212 minus, `+`-signed, and whitespace-padded ids (the `isdigit() and isascii()` pair is sufficient); `to_legacy_i64`'s order-18 cap is exact and never data-dependent (southern all-4 order-18 `-6444444444444444444` fits int64 and inverts bit-exactly through `from_legacy`, while the *smallest* order-19 id `1.11e19` already overflows); the `<U31` width claim holds (northern order-29 renders 30 chars, southern 31, kernel-bounded so numpy's silent U31 truncation is unreachable); sentinel words raise through `decimal_repr` before any emit path returns; the order-29 point→area collapse through hive round-trip is documented and pinned by `test_golden_repr_not_injective_point_vs_area`; pandas `Series.array` round-trips all three new methods.

**Conventions:** `mortie/morton_index.py` is 644 lines (under the ~1000 cap), no new dependencies, both flake8 gates clean, all 31 tests in the touched suites pass locally, commit style matches §3.

---

mortie/morton_index.py
espg: 🤖 *from Claude*

Recorded as deliberate, not changed in this PR: the all-zero sentinel *is* the array's storage-level missing value, and boxing it as `MortonIndexScalar(0)` keeps `__getitem__` total over the raw words (the strict `dtype.na_value` return is a pre-existing deviation, as noted). The `<NA>` rendering matches the array formatter's treatment of the same word, so display is at least consistent; making `__getitem__` return `pd.NA` is an ExtensionArray-contract change that belongs to its own decision, standing for review here.

---

mortie/morton_index.py
espg: 🤖 *from Claude*

Accepted as a recorded tradeoff: hive paths are manifest-scale I/O (LIST results), not bulk compute, so ~0.54 s / 100k parses is well inside budget and the per-call clarity wins. If a bulk consumer appears, the vectorization you sketch (group by order, batch `rust_mi_from_nested` per group) is the plan of record.

---

mortie/morton_index.py
```diff
@@ -428,20 +504,118 @@ def decimal_repr(self):
...
+                head = 2 if dec.startswith("-") else 1
+                levels = [dec[:head], *dec[head:]]
+                # Enforce the directory cross-check only when the chain is
+                # anchored: the {sign+base} component at its expected slot.
+                # (Root components are indistinguishable from digit dirs by
+                # count alone, so an unanchored tail is treated as root.)
+                if (
+                    len(parts) - 1 >= len(levels)
+                    and parts[-1 - len(levels)] == levels[0]
+                ):
```

**Anchoring on equality with `levels[0]` regresses the mis-filed-base case (verified).** Phase 4's count-based check raised on `"-4/1/1/2/3/-31123.zarr"` — a full digit chain filed under the wrong base cell. Phase 5 silently accepts it, because the one component used to *detect* the chain (the sign+base slot) is now the one component that can never be *validated*: any wrong value there demotes the whole chain to "root". A wrong-base mis-filing is arguably the most likely one in practice (coarsest split), and it's exactly the class this cross-check exists for.

```
>>> MIA.from_hive_path("-4/1/1/2/3/-31123.zarr")   # phase 5: parses as -31123
>>> # phase 4: ValueError: directories '-4/1/1/2/3' do not match leaf id '-31123'
```

A middle ground that keeps the phase-5 fix for arbitrary roots but recovers this case: engage the check when the anchor slot matches the sign+base *shape* (optional `-` + digit `1..6`) rather than exact equality with `levels[0]` — `"-4"` then anchors and raises, while `"data"`, `"s3:"`, `"archive"` roots still skip. Not raising here may also be an acceptable documented tradeoff, but the current tests (`test_misfiled_leaf_raises`) don't pin either way on a wrong anchor, so the regression is invisible to the suite — worth a decision plus a pinning test whichever way it goes.

Relatedly, the residual false-reject the comment alludes to is real and worth one docstring sentence: a root component coincidentally equal to the anchor token at exactly its slot still rejects, e.g. `"-3/a/b/c/d/-31123.zarr"` raises, and for positive ids the anchor is a bare digit so `"v/3/x/y/z/w/31123.zarr"` raises too (numeric root dirs like dates make single-digit collisions plausible). Verified both against efcc196.

---

mortie/morton_index.py
```diff
@@ -428,20 +504,118 @@ def decimal_repr(self):
...
+            iterable of them. The leaf basename carries the full decimal id;
+            when the path also carries the digit directories -- recognized by
+            the ``{sign+base}`` component sitting at its slot above the leaf
+            -- the whole chain is checked against the leaf and a mis-filed
+            leaf raises ``ValueError``. A bare ``{full_id}.zarr``, or one
+            under an arbitrary root without the digit chain, skips the check.
+            Order-29 ids parse to the *area* word (see
+            :func:`_decimal_to_word`).
+            """
+            if isinstance(paths, (str, os.PathLike)):
```
espg, line 586: 🤖 *from Claude (review)*

Minor, non-blocking: the new `os.PathLike` acceptance splits `str(p)` on `"/"` only, so on Windows a `Path("-3/1/1/2/3/-31123.zarr")` stringifies with backslashes and every hive-layout PathLike raises (`malformed decimal Morton id '-3\\1\\1\\2\\3\\-31123'` — verified via `PureWindowsPath`). It fails loudly rather than mis-parsing, and CI is ubuntu-only so it's untested either way, but since `build-wheels.yml` ships Windows wheels: either normalize (`pathlib.PurePath(p).parts` when the input is a PathLike, or `str(p).replace(os.sep, "/")`) or note posix-separator-only in the docstring.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the phase-5 commit (efcc196), exercised against the built extension in a clean interpreter.

**Checked out clean:**
- `__reduce__` is correct across the full uint64 range: round-tripped words with bit 63 set and `2**64-1` through all pickle protocols 0-5 plus `copy.deepcopy`; `int(self)` never overflows (Python int) and `np.uint64.__new__` accepts the full range without warnings. Type and decimal display survive the boundary.
- `__format__` pinning tests and the `%`-formatting doc line match observed behavior (`"%d" % s` emits the raw word; `f"{s:d}"` raises).
- The two paths the phase targets behave as intended: `"-3/1/1/2/3/-31123.zarr"` validates, `"-3/1/1/2/4/-31123.zarr"` raises, `"data/-3/-31123.zarr"` (short tail) parses as root. Full suite: 58 passed; flake8 E9/F63/F7/F82 clean; commit message and module size conform to conventions.

**Findings (inline):**
1. The exact-equality anchor regresses the wrong-base mis-filing: `"-4/1/1/2/3/-31123.zarr"` raised in phase 4 and silently parses in phase 5 — the anchor component is now unvalidatable by construction. Suggested a shape-match anchor (or a documented decision + pinning test). Also noted the residual coincidental-root false-reject (real, verified for both signs) deserves a docstring sentence.
2. Minor: PathLike support is posix-separator-only; Windows `Path` inputs raise (loudly) and CI won't catch it.

No blocking issues beyond the decision in (1).

---

mortie/morton_index.py

Folded in phase 6 exactly as suggested: the anchor is now the sign+base *shape* (optional `-` + digit `1..6`) rather than equality with the leaf's own token, so `"-4/1/1/2/3/-31123.zarr"` and `"2/1/1/2/3/41123.zarr"` raise again (both pinned in `test_misfiled_leaf_raises`), while `"data"`/`"s3:"` roots still skip, and a real root above the chain stays fine (`test_root_ending_in_digit_chain_still_parses`). The residual single-digit-root collision false-reject is the accepted tradeoff and is implied by the docstring's shape wording; it stays strict-on-suspicious rather than lenient.

---

mortie/morton_index.py
espg: 🤖 *from Claude*

Folded in phase 7: `PurePath` inputs are now split via their own `.parts` (so a `WindowsPath` honors backslashes), with a `PureWindowsPath` round-trip pinned in `test_pathlike_input` — testable on the ubuntu CI since the flavor, not the host, decides the split. Raw strings keep the exact `"/"`-split semantics.

---

espg:
🤖 *from Claude*

All phases complete (1–7), local gates green (flake8 + 578 pytest), review threads addressed or answered; marking `waiting` for your review. One process note: CLAUDE.md §2 says to flip a fully-phased PR out of draft, but this routine runs under a drafts-only constraint (never mark ready-for-review), so per the CLAUDE.md conflict rule I'm leaving it **draft** and flagging it here — flip it when you're satisfied. Four small design questions are under "Questions for review" in the body; none block review.

---

espg:
🤖 *from Claude*

CI note: everything is green (tests 3.10–3.12, `arro3-no-pyarrow`, ruff, codecov patch/project, all wheel builds) except the **CodSpeed gate**, which flags `fast_norm2mort_orders[6]` (−17.8%) and `[10]` (−15.6%). That's runner noise, not a real regression: this PR contains no Rust changes at all (pure Python display layer + tests), the [report itself](https://github.com/espg/mortie/pull/105#issuecomment-4930959405) headlines "will not alter performance" and warns "Different runtime environments detected", and the *same two benchmarks* swung **+12%** on the equally-Rust-free PR #102 ([its report](https://github.com/espg/mortie/pull/102#issuecomment-4885816527)). Acknowledging the flag is a CodSpeed-dashboard action on your account, so I'm leaving it standing rather than dismissing it myself.

---

Issue #90, opened 2026-06-30
Evaluate relaxing node_straddles containment now that on_minor_arc is SoS-hardened


espg:
🤖 *from Claude*

Tracking follow-up — awaiting triage. No action label applied yet; leaving label/triage to the maintainer.

## Context

PR #87 (Refs #78) made `robust_crossing`'s antipodal-disambiguation step `on_minor_arc` exactly SoS-hardened: each wedge sign is now decided on a canonical, identity-sorted evaluation with an identity-keyed symbolic perturbation tie-break, so the predicate is total and reorder-invariant with no residual f64 sign sensitivity at endpoint coincidence.

Before that change, the caveat at `src_rust/src/coverage.rs:442` noted that boundary cells affected by `on_minor_arc`'s residual float sensitivity were caught by the `node_straddles` over-refinement safety net — such cells are refined (descended into) rather than filled whole. Now that the disambiguation is exact, that safety net *may* be loosenable: some cells the descent currently over-refines for safety might be safely fillable, reducing output size / work.

## Re-scope (post PR #106 / issue #103) — supersedes the plan below where they differ

PR #106 replaced the entire crossing pipeline with the uniform symbolic predicate (`arcs_cross_sos` + exact determinant signs), so the `node_straddles` over-refinement is no longer load-bearing for correctness anywhere — it is now **pure conservatism**, and this issue's measure-then-relax program can be run safely. Adjustments:

1. **Baseline = PR #106's branch (or main once it merges), not pre-#87 main.** The measurement is only meaningful against the exact predicate.
2. **Relaxation target = the non-touching straddle over-refinement only.** The closed-set contract from #103 *deliberately* refines and includes boundary-touching cells (`edge_hits_cell_edge`'s exact-incidence path); those refinements are contract, not waste, and must be excluded from the "over-refinement" count.
3. **Guard rail**: PR #106's parity oracle (`debug_assert` in the descent) validates every uniform cell against the independent winding PIP — any relaxation experiment runs under it in the test profile, which the original plan lacked.

## Recommendation

Measure before changing behavior. Specifically:

1. Quantify the over-refinement `node_straddles` currently causes (count of cells refined-not-filled, **excluding closed-set boundary-touch refinements**, across representative polygons) against the exact-predicate baseline.
2. Only then decide whether/how to relax the containment, with the measurement as the baseline to confirm no coverage correctness regression.

[...]

Issue #68, opened 2026-06-25
v1.0: full codebase review for the 1.x freeze


espg:
🤖 *from Claude*

Sub-issue of #48 (v1.0 sweep), Tier-2.

## Scope
A full pass over the codebase to make sure nothing ≤18-era or signed-decimal leaks into the frozen 1.x surface:

- Enumerate the public API surface (`mortie/__init__.py` `__all__`) and confirm each export is intended to be frozen; make anything not-frozen private now.
- Sweep `mortie/*.py` and `src_rust/src/*.rs` for any remaining order-18 caps / decimal-digit string scanning / signed-southern assumptions (this dovetails with #60's order-29 audit and the mixed-order MOC ceiling).
- Check docstrings still describe the current encoding (several still say "1–18").

## Acceptance criteria
A findings list with file:line references; mechanical fixes folded in, design questions raised for @espg.

Refs #48, #60.

espg:
enabled by #118 ; post plan and proceed to implement

espg:
🤖 *from Claude*

Recording the close gate, per espg's direction: **this issue stays open until both the `clip2order` decision and the #135 docstring sweep land.** PR #130 therefore stays `Refs #68`, not `Closes`.

## Where the freeze audit stands

**Landed on PR #130** (CI green, mergeable, awaiting merge):

- Stale signed-decimal docstrings in `prefix_trie.py` corrected — the original phase-1 diff.
- `heal_norm` **removed** (public API removal). Audited first: no caller anywhere in mortie, and zagg does not use it. One example-notebook cell inlined the identical arithmetic rather than being rewired, so its committed outputs stay valid.
- `MortonChild` documented as a **return type** — read surface (`characteristic`, `len`, `children`, `nchildren`, `mantissa_array`, `cell_area`) frozen; `__init__` marked internal, since the production path builds nodes via `__new__` and never calls it.
- `res2display` **reshaped** to return `list[ResolutionLevel(order, value, unit, km)]` instead of printing; USAGE.md rewritten against real output and a numpydoc `Examples` block added.
- `split_children` / `split_children_geo` confirmed frozen, no change needed.

**Split out to their own issues** (each was a behavior change that did not belong in a docstring PR):

- **#136** — the UNIQ helper family: `uniq2geo` silently trusting a caller-supplied `order` (a real correctness trap — wrong order, wrong coordinates, no error), `unique2parent`'s single-resolution limitation, and the `order=18` defaults. Labelled `implement`, in progress.
[...]

PR #106, opened 2026-07-10
Uniform symbolic crossing predicate: fix base-cell-boundary mis-fill

### LINKED ISSUE #103: morton_coverage / from_wkt mis-fill polygons whose edges lie on base-cell boundaries (lon ≡ 0 mod 45°, lat = 0)
🤖 *from Claude*

## Summary

`morton_coverage`, `morton_coverage_moc`, and the public `from_wkt` / `from_geometry` entry points **mis-fill polygons whose edges lie exactly on HEALPix base-cell boundaries** — longitudes that are a multiple of 45°, and latitude 0.

The failure is not subtle. A 5°×5° box at `lat[20,25] lon[0,5]` — whose only unusual property is that its west edge sits on the prime meridian — returns cells spanning **the entire longitude range** (`lon 0.00..359.91`) instead of `lon 0..5`.

The fill both **over-covers** (by up to two orders of magnitude) and **under-covers** (misses cells the polygon genuinely contains), so it can't be defended as a conservative approximation.

Version: mortie **0.8.5**, numpy 2.x, Python 3.12, macOS arm64.

## Reproducer

Depends only on `mortie` + `numpy`.

```python
import numpy as np
from mortie import morton_coverage, mort2geo

def box(lat_lo, lat_hi, lon_w, lon_e):
    return (np.array([lat_lo, lat_lo, lat_hi, lat_hi, lat_lo], float),
            np.array([lon_w,  lon_e,  lon_e,  lon_w,  lon_w ], float))

def extent(keys):
    g  = [mort2geo(int(k)) for k in keys]
    la = np.array([float(x[0][0]) for x in g])
    lo = np.array([float(x[1][0]) for x in g])
    return f"n={len(keys):7d}  lat {la.min():7.2f}..{la.max():7.2f}  lon {lo.min():7.2f}..{lo.max():7.2f}"

for label, a, b, c, d in [
    ("OK   lat[20,25] lon[ 10, 15]  (edges off the 45-deg grid)", 20., 25.,  10., 15.),
    ("BAD  lat[20,25] lon[  0,  5]  (west edge lon=0)",           20., 25.,   0.,  5.),
    ("BAD  lat[20,25] lon[ 40, 45]  (east edge lon=45)",          20., 25.,  40., 45.),
    ("BAD  lat[20,25] lon[ 90, 95]  (west edge lon=90)",          20., 25.,  90., 95.),
    ("OK   lat[-5, 0] lon[100,105]  (top edge lat=0)",            -5.,  0., 100.,105.),
    ("BAD  lat[ 0, 5] lon[100,105]  (bottom edge lat=0)",          0.,  5., 100.,105.),
]:
    lats, lons = box(a, b, c, d)
    print(f"{label:58s} {extent(np.asarray(morton_coverage([lats],[lons], order=9)))}")
```

Output:

```
OK   lat[20,25] lon[ 10, 15]  (edges off the 45-deg grid)  n=   1886  lat   19.95..  25.04  lon    9.93..  15.03
BAD  lat[20,25] lon[  0,  5]  (west edge lon=0)            n=   1687  lat   19.55..  27.20  lon    0.00.. 359.91
BAD  lat[20,25] lon[ 40, 45]  (east edge lon=45)           n=   2419  lat   19.55..  27.20  lon   39.46..  50.54
BAD  lat[20,25] lon[ 90, 95]  (west edge lon=90)           n=  14771  lat   19.87..  41.71  lon   78.84.. 101.16
OK   lat[-5, 0] lon[100,105]  (top edge lat=0)             n=   2003  lat   -5.00..   0.00  lon   99.93.. 105.03
BAD  lat[ 0, 5] lon[100,105]  (bottom edge lat=0)          n=  59579  lat  -19.39..  19.39  lon   95.71.. 134.91
```

Note the last pair: a box whose **top** edge is the equator is fine; the same box moved to sit **on top of** the equator returns 59,579 cells spanning 39° of longitude and ±19.39° of latitude.

`morton_coverage_moc(..., order=13)` → `moc_to_order(..., 9)` reproduces every one of these, and so does the public API:

```python
>>> mortie.from_wkt("POLYGON((0 20, 5 20, 5 25, 0 25, 0 20))", order=9)   # west edge lon=0
n=   1687  lat   19.55..  27.20  lon    0.00.. 359.91
>>> mortie.from_wkt("POLYGON((10 20, 15 20, 15 25, 10 25, 10 20))", order=9)
n=   1886  lat   19.95..  25.04  lon    9.93..  15.03
```

## Under-coverage, not just over-coverage

Ground truth via the point→cell path (`geo2mort` → `clip2order`), sampling constant-latitude rows at ~2 km spacing, for the full latitude ring `lat[-87.0, -86.7]` at order 9 split into 8 × 45° sectors:

| | cells returned | reaches lat | misses (of the 826 true cells) |
|---|---|---|---|
| `morton_coverage`, sector edges at multiples of 45° | 2,173 | **−89.91** (the pole) | **111** |
| `morton_coverage`, sector edges offset by 22.5° | 828 | −87.08 | **0** |

So at the degenerate placement it returns 2.6× too many cells *and* still misses 111 cells that are genuinely inside the polygon.

Worse further from the pole — ring `lat[-83.0, -82.7]`, order 9:

```
sector offset  0.0 deg: n=   9154  lat  -89.91.. -78.38     # true answer: 1560 cells
sector offset 22.5 deg: n=   1560  lat  -83.05.. -82.69     # exact match to ground truth
```

## Why this bites in practice

The natural way to express a full latitude ring (to dodge the known antimeridian collapse of a single −180…180 rectangle) is to split it into 8 × 45° sectors. That puts **every sector edge exactly on a multiple of 45°** — i.e. on a base-cell boundary meridian — so every sector hits this bug at once.

Downstream in zagg this silently produced ring coverages containing polar-cap cells thousands of km from the ring, while also dropping in-ring cells. It surfaced only because a shard 8° from the requested band showed up as the densest.

## Scope / what is *not* affected

- Polygons with irregular vertices that don't land on the 45° grid are fine. I checked 120 ICESat-2 granule swath footprints through `morton_coverage_moc` against independently computed spherical latitude bounds: **0 leaked**, worst excess 0.00°.
- Small AOI boxes away from the 45° meridians and the equator are fine (verified against a point-sampled ground truth).

## Workaround

Offset polygon edges off the 45° grid (e.g. build ring sectors starting at 22.5°), or bypass the fill entirely by mapping sample points through `geo2mort` → `clip2order`.

## Suggested direction

The trigger looks like an exactly-on-boundary classification: a vertex or edge lying on a base-cell boundary is being assigned to the wrong side, after which the fill escapes into the neighbouring base cell (and, in the polar caps, runs to the pole). An inclusive/exclusive tie-break at base-cell edges — or an epsilon nudge of boundary-coincident vertices before the interior test — would likely fix it. A regression test over boxes with edges on `lon ∈ {0, 45, …, 315}` and `lat = 0` would lock it down.

espg:
Closes #103. Refs #78, #87, #44 (the prior per-predicate hardenings this generalizes), #90 (re-scoped on the back of this PR), #107 (the follow-up limitation this PR's oracle discovered).

## What this does

Implements the plan agreed on the issue thread ([approach (1)](https://github.com/espg/mortie/issues/103#issuecomment-4930976666), [sign-off + closed-set semantics](https://github.com/espg/mortie/issues/103#issuecomment-4931072519)): one uniform symbolic crossing predicate for every sidedness decision in the coverage descent. **All the issue's reproducers now come back in-band** (see "How it was tested").

## Phases

- [x] **Phase 1 — the predicate + differential tests.** `arcs_cross_sos` in `src_rust/src/sphere.rs`: arcs cross iff the four orientations `[a c b]`, `[c b d]`, `[b d a]`, `[d a c]` share a sign (the S2 `SimpleCrossing` identity), each through the SoS-hardened `orient_sos`. No constructed intersection point. Five test groups incl. the exact #103 parity-breaking configuration (pinned) and vertex-graze-counts-once.
- [x] **Phase 2 — wiring + the deeper fixes the parity oracle forced.** Call-site swap (old `robust_crossing`/`on_minor_arc` pipeline deleted); closed-set touch semantics with a segment gate (`edge_hits_cell_edge`); the parity oracle `debug_assert` (uniform-cell fill must equal the independent winding PIP); globally unique UNIQ-coded SoS ids for cell centres (positional ids made the fill chain path-dependent at boundary-coincident centres); seed hardening (`seed_fill` geometric ambiguity detection + `base_fills` donor chaining); **exact determinant signs** under `orient_sos` (Shewchuk error-free expansions, std-only, no new dependency, filter-guarded) — near-coplanar noise signs were jointly inconsistent across triples, the residual #103 mechanism off the bit-exact grid.
- [x] **Phase 3 — regression matrix** (`mortie/tests/test_coverage_boundary.py`, 87 tests): all six degenerate meridians incl. 315 × three latitude bands × both sides × both hemispheres; equator edges incl. through (0,0); the non-monotone width family; flagship reproducers at original orders; winding invariance. Asserts superset / no-escape (bulge-aware) / flat == densified MOC.
- [x] **Phase 4 — folded self-review findings.** Major: `base_fills`' donor chain ran ~122° chords through the bare two-straddle fast path (antipodal phantom crossings) — now unconditionally symbolic, reviewer's reproducer pinned. Medium: the closed-set span gate's cos-space epsilon → angular (`sin ρ`-scaled). Plus docstring precondition, rustdoc link, stale docs.
- [x] **Phase 5 — perf recovery for the CodSpeed regression.** The −15..35% on clean-geometry coverage benches traced to `edge_hits_cell_edge` computing all four determinants up front where `arcs_cross_n` exited after two on the common corners-same-side case (the straddle-test hot path), plus per-seed re-normalization of edge normals in `seed_fill`. The two-dot early exit is restored (same shape as the retired `arcs_cross_n`), unit normals are computed once per descent, and `center_id` is hoisted out of the corner loops. One documented closed-set nuance from the fast exit: a pure vertex-**point** touch on a cell edge with the polygon strictly outside is guaranteed only for the vertex's leaf-owning cell; edge-collinear touches — the systematic on-grid family the contract exists for — are always degenerate-path and fully honored. Awaiting CodSpeed re-measure on this head; the exact-arithmetic layer itself is filter-guarded and was never the cost.

## How it was tested

- `cargo test` — 197 passed (parity oracle live in the test profile), 1 deliberately `#[ignore]`d (#107). `cargo clippy` no new warnings; `cargo fmt --check` clean.
- `maturin develop --release` + `pytest` — **635 passed** (87 new matrix tests), 11 skipped; `flake8` clean.
- Issue reproducers, before → after at the issue's orders: belt `lon[0,5]` o9 1687-cells-spanning-the-globe → **1887 in-band** (control 1886); cap `lon[44,45]` o6 829-to-the-pole → **13 in-band**; 315° mirror-symmetric (**13**); southern mirror **13**; ring sector o9 775-reaching-−89.9 → **87**, max dip −87.26 vs the polygon's own chord dip of −87.53.

## Questions for review

1. ~~Known limitation~~ → filed as **#107** with the full diagnosis per your go-ahead (`#[ignore]`d acceptance test in-tree).
2. Module sizes (`sphere/tests.rs`, `coverage.rs` past ~1000 lines) — **deferred to the 1.x freeze sweep (#68)** per your call.
3. #90 — **re-scoped** per your ask ([comment](https://github.com/espg/mortie/issues/90#issuecomment-4932545686)): unblocked by this PR, baseline moves here, relaxation target excludes closed-set touch refinements, parity oracle as guard rail.
4. **New (phase 5):** is the vertex-point-touch nuance above acceptable as the closed-set contract's fine print, or do you want the neighbor cell guaranteed too (costs the two extra dots on the hot path — roughly the difference CodSpeed flagged)?

### DIFF
--- a/src_rust/src/coverage.rs
+++ b/src_rust/src/coverage.rs
@@ -505,19 +505,29 @@ fn edge_crosses_probe(p: &Vec3, q: &Vec3, ip: PointId, iq: PointId, n_pq: &Vec3,
 }
 
 /// Does polygon edge `e` cross **or exactly touch** the cell edge `c1 → c2`
-/// (normal `n_c = c1 × c2`)?  The straddle test's closed-set form (#103):
+/// (normal `n_c = c1 × c2`)?  The straddle test's closed-set form (#103).
 ///
-/// * A **bit-exact zero** determinant is true geometric incidence — a cell
-///   corner on the polygon edge's great circle or a polygon endpoint on the
-///   cell edge's.  HEALPix corners on the lon ≡ 0 mod 45° meridians are
-///   bit-exact (`y == 0.0`), so on-grid input hits this systematically.  It
-///   reports as a hit, making the cell boundary-touching → refined and, at the
-///   target order, included: "a cell whose boundary the polygon touches
-///   *exactly* is always included" (the closed-set contract agreed on #103).
+/// The hot path keeps the exact two-dot shape of the retired `arcs_cross_n`:
+/// corners strictly (beyond [`ORIENT_EPS`]) on one side of the polygon edge's
+/// great circle exit immediately — the CodSpeed-visible cost of computing all
+/// four determinants up front was a ~19% coverage regression.  Degenerate
+/// configurations take the closed-set logic:
+///
+/// * A **bit-exact zero** determinant with the incident point inside the
+///   *segment* cap (angular slack, `sin ρ`-scaled) is true geometric
+///   incidence — the cell is boundary-touching → refined → included ("a cell
+///   whose boundary the polygon touches exactly is always included").
 /// * A near-zero (< [`ORIENT_EPS`]) but nonzero determinant routes to the
-///   symbolic [`arcs_cross_sos`] for an exact crossing verdict (a 1-ulp-off
-///   edge resolves deterministically to one side — not closed-set).
-/// * Otherwise the plain sign test, unchanged (the hot path).
+///   symbolic [`arcs_cross_sos`] for an exact verdict (a 1-ulp-off edge
+///   resolves deterministically to one side — not closed-set).
+///
+/// One documented nuance of the fast exit: a polygon **vertex-point** lying
+/// exactly on this cell edge while the polygon stays strictly outside the
+/// cell (corners strictly one side, no crossing) is not detected here, so
+/// only the vertex's leaf-owning cell (the vertex-in-cell clause) is
+/// guaranteed for a pure point-touch.  Edge-collinear touches — the
+/// systematic on-grid family the closed-set contract exists for — always
+/// make `d1`/`d2` degenerate and are fully honored.
 ///
 /// Cell corners are one-shot derived points ([`CORNER_ID_A`]/[`CORNER_ID_B`]):
 /// only distinctness matters within a call — this feeds the straddle boolean,
@@ -526,43 +536,38 @@ fn edge_crosses_probe(p: &Vec3, q: &Vec3, ip: PointId, iq: PointId, n_pq: &Vec3,
 fn edge_hits_cell_edge(e: &Edge, c1: &Vec3, c2: &Vec3, n_c: &Vec3) -> bool {
     let d1 = dot(&e.n_ab, c1);
     let d2 = dot(&e.n_ab, c2);
+    let degenerate12 = d1.abs() < ORIENT_EPS || d2.abs() < ORIENT_EPS;
+    if !degenerate12 && (d1 > 0.0) == (d2 > 0.0) {
+        return false; // hot exit: corners strictly on one side
+    }
     let d3 = dot(n_c, &e.a);
     let d4 = dot(n_c, &e.b);
-    // Exact incidence ⇒ closed-set touch — but only when the incident point
-    // lies within the *segment* (its bounding cap), not merely on the infinite
-    // great circle: an on-grid meridian edge's circle runs to the poles, and
-    // gating on the circle alone emitted a spurious strip of cells along it
-    // far beyond the polygon.
-    // The span slack is *angular* (cos(ρ+ε) ≈ cos ρ − ε·sin ρ): a bare
-    // cos-space epsilon would floor the angular slack at √(2ε) ≈ 1.4e-6 rad
-    // for short segments, gating vertices hundreds of cell-widths away as
-    // "touching" at fine orders.
-    let span = |cos_rho: f64, sin_rho: f64, d: f64| {
-        d >= cos_rho - sin_rho * ORIENT_EPS - ORIENT_EPS * ORIENT_EPS
-    };
-    if (d1 == 0.0 && span(e.cos_rho, e.sin_rho, dot(&e.mid, c1)))
-        || (d2 == 0.0 && span(e.cos_rho, e.sin_rho, dot(&e.mid, c2)))
-    {
-        return true;
-    }
-    if d3 == 0.0 || d4 == 0.0 {
-        let mid = normalize(&[c1[0] + c2[0], c1[1] + c2[1], c1[2] + c2[2]]);
-        let cos_rho = dot(&mid, c1);
-        let sin_rho = (1.0 - cos_rho * cos_rho).max(0.0).sqrt();
-        if (d3 == 0.0 && span(cos_rho, sin_rho, dot(&mid, &e.a)))
-            || (d4 == 0.0 && span(cos_rho, sin_rho, dot(&mid, &e.b)))
+    if degenerate12 || d3.abs() < ORIENT_EPS || d4.abs() < ORIENT_EPS {
+        // Exact incidence ⇒ closed-set touch, gated to the segment span (the
+        // infinite great circle runs to the poles; gating on it alone emitted
+        // a spurious strip of cells far beyond the polygon).
+        let span = |cos_rho: f64, sin_rho: f64, d: f64| {
+            d >= cos_rho - sin_rho * ORIENT_EPS - ORIENT_EPS * ORIENT_EPS
+        };
+        if (d1 == 0.0 && span(e.cos_rho, e.sin_rho, dot(&e.mid, c1)))
+            || (d2 == 0.0 && span(e.cos_rho, e.sin_rho, dot(&e.mid, c2)))
         {
             return true;
         }
-    }
-    if d1.abs() < ORIENT_EPS
-        || d2.abs() < ORIENT_EPS
-        || d3.abs() < ORIENT_EPS
-        || d4.abs() < ORIENT_EPS
-    {
+        if d3 == 0.0 || d4 == 0.0 {
+            let mid = normalize(&[c1[0] + c2[0], c1[1] + c2[1], c1[2] + c2[2]]);
+            let cos_rho = dot(&mid, c1);
+            let sin_rho = (1.0 - cos_rho * cos_rho).max(0.0).sqrt();
+            if (d3 == 0.0 && span(cos_rho, sin_rho, dot(&mid, &e.a)))
+                || (d4 == 0.0 && span(cos_rho, sin_rho, dot(&mid, &e.b)))
+            {
+                return true;
+            }
+        }
         return arcs_cross_sos(c1, c2, &e.a, &e.b, CORNER_ID_A, CORNER_ID_B, e.ia, e.ib);
     }
-    (d1 > 0.0) != (d2 > 0.0) && (d3 > 0.0) != (d4 > 0.0)
+    // d1, d2 straddle (established above); the plain sign test decides.
+    (d3 > 0.0) != (d4 > 0.0)
 }
 
 /// A descent node: cell `(pixel, depth)`, its centre, even-odd fill state, and
@@ -629,9 +634,9 @@ fn covers_complement(rings: &[Vec<Vec3>], cap: &Cap) -> bool {
 /// great-circle plane makes the seed ambiguous, and [`base_fills`] classifies
 /// it through the SoS parity chain, whose answer is exact in the symbolically
 /// perturbed world the descent's own probes live in.
-fn seed_fill(x: &Vec3, edges: &[Edge], rings: &[Vec<Vec3>]) -> Option<bool> {
-    for e in edges {
-        if dot(&normalize(&e.n_ab), x).abs() < ORIENT_EPS {
+fn seed_fill(x: &Vec3, unit_normals: &[Vec3], rings: &[Vec<Vec3>]) -> Option<bool> {
+    for n in unit_normals {
+        if dot(n, x).abs() < ORIENT_EPS {
             return None;
         }
     }
@@ -661,10 +666,13 @@ fn antipodal_base(a: u64, b: u64) -> bool {
 /// passes through all 12 — pathological), the raw winding verdicts are kept.
 fn base_fills(edges: &[Edge], rings: &[Vec<Vec3>]) -> [bool; 12] {
     let centers: Vec<Vec3> = (0..12).map(|b| cell_center_vec(0, b)).collect();
+    // Unit edge normals once (not per seed): the plane-proximity test is the
+    // only consumer and 12 × E normalizations showed up in the profile.
+    let unit_normals: Vec<Vec3> = edges.iter().map(|e| normalize(&e.n_ab)).collect();
     let mut fill = [false; 12];
     let mut known = [false; 12];
     for b in 0..12 {
-        if let Some(f) = seed_fill(&centers[b], edges, rings) {
+        if let Some(f) = seed_fill(&centers[b], &unit_normals, rings) {
             fill[b] = f;
             known[b] = true;
         }
@@ -831,16 +839,12 @@ fn node_straddles(node: &Node, edges: &[Edge], order: u8) -> bool {
     let quad_straddles = node.relevant.iter().any(|&i| {
         let e = &edges[i];
         (0..4).any(|ci| edge_hits_cell_edge(e, &corners[ci], &corners[(ci + 1) % 4], &n_quad[ci]))
-    }) || corners.iter().any(|c| {
-        arc_crossing_parity(
-            &node.center,
-            c,
-            center_id(node.depth, node.pixel),
-            CORNER_ID_A,
-            &node.relevant,
-            edges,
-        )
-    });
+    }) || {
+        let cid = center_id(node.depth, node.pixel);
+        corners
+            .iter()
+            .any(|c| arc_crossing_parity(&node.center, c, cid, CORNER_ID_A, &node.relevant, edges))
+    };
     if quad_straddles {
         return true;
     }
@@ -859,16 +863,11 @@ fn node_straddles(node: &Node, edges: &[Edge], order: u8) -> bool {
     node.relevant.iter().any(|&i| {
         let e = &edges[i];
         (0..n).any(|ci| edge_hits_cell_edge(e, &bnd[ci], &bnd[(ci + 1) % n], &n_bnd[ci]))
-    }) || bnd.iter().any(|b| {
-        arc_crossing_parity(
-            &node.center,
-            b,
-            center_id(node.depth, node.pixel),
-            CORNER_ID_A,
-            &node.relevant,
-            edges,
-        )
-    })
+    }) || {
+        let cid = center_id(node.depth, node.pixel);
+        bnd.iter()
+            .any(|b| arc_crossing_parity(&node.center, b, cid, CORNER_ID_A, &node.relevant, edges))
+    }
 }
 
 /// The four children of a node, each with re-culled edges and propagated fill.
--- a/src_rust/src/coverage.rs
+++ b/src_rust/src/coverage.rs
@@ -38,8 +38,8 @@ use crate::cell_geom::{cell_center_vec, cell_corners, Cap};
 use crate::geo2mort::{ang2pix_scalar, boundaries_step_scalar};
 use crate::morton::nested2mort;
 use crate::sphere::{
-    arcs_cross_n, cross, dot, latlon_to_unit_vec, norm, normalize, parity_filled_robust,
-    ring_winding_sign_at, robust_crossing, PointId, Vec3,
+    arcs_cross_sos, cross, dot, latlon_to_unit_vec, norm, normalize, parity_filled_robust,
+    ring_winding_sign_at, PointId, Vec3,
 };
 
 // ── public entry points ──────────────────────────────────────────────────
@@ -331,27 +331,41 @@ struct Edge {
     leaf: u64,
     /// Stable Simulation-of-Simplicity identities of the endpoints `a`, `b`
     /// (their global vertex index across the ring-set).  Feed the descent's
-    /// robust crossing test ([`robust_crossing`]) so a probe arc whose endpoint
+    /// symbolic crossing test ([`arcs_cross_sos`]) so a probe arc whose endpoint
     /// lies exactly on this edge's great circle — e.g. a HEALPix base-cell centre
-    /// on a base-cell-centre meridian (issue #11) — resolves to a definite,
-    /// traversal-order-independent side instead of a sign-unstable `arcs_cross_n`.
+    /// on a base-cell-centre meridian (issues #11/#103) — resolves to a definite,
+    /// traversal-order-independent side instead of a sign-unstable plain test.
     ia: PointId,
     ib: PointId,
 }
 
-/// SoS identities reserved for the two endpoints of a descent **probe** arc
-/// (cell centre → centre/corner).  Ring vertices are numbered from
-/// [`VERTEX_ID_BASE`] upward, so the probe endpoints take ids `0` and `1` — i.e.
-/// they sort **before** every ring vertex.  This matters when a probe endpoint
-/// lies exactly on an edge's great circle (the issue #11 base-cell-centre-on-a-
-/// meridian degeneracy): SoS resolves the tie by perturbing the lowest-id point
-/// first, so the probe point — not a ring vertex — is the one nudged decisively
-/// off the edge, and it is nudged the **same** way every time it anchors a walk.
-/// That keeps the descent's incremental fill parity consistent with itself.
-const PROBE_ID_P: PointId = 0;
-const PROBE_ID_Q: PointId = 1;
-/// Ring vertices are numbered from here so their ids never collide with the two
-/// reserved probe ids above.
+/// Symbolic identity of a cell centre: its HEALPix UNIQ code with the top bit
+/// set, so it is unique per `(depth, pixel)`, stable across every probe arc
+/// that touches the centre, and disjoint from the ring-vertex id range.
+///
+/// Stability is load-bearing (#103): the even-odd fill parity is chained along
+/// centre→centre probe arcs, and when a centre lies **exactly** on a polygon
+/// edge its SoS-resolved virtual side must be the same in the arc that arrives
+/// at it and in every arc that leaves it — otherwise the chain is
+/// path-dependent and one boundary-coincident centre inverts its whole
+/// subtree.  Positional ids (source=0 / target=1, the pre-#103 scheme) broke
+/// exactly this.  With per-point ids the symbolic perturbation is one global,
+/// consistent perturbation of the configuration; centres carry the top-bit
+/// range, so ring vertices (low ids) perturb first — the polygon is nudged off
+/// the degenerate grid once, the same way for every probe.
+#[inline]
+fn center_id(depth: u8, pixel: u64) -> PointId {
+    (1u64 << 63) | ((4u64 << (2 * depth as u32)) + pixel)
+}
+
+/// SoS ids for one-shot derived endpoints (cell corners, densified near-pole
+/// boundary points).  These feed only the straddle *boolean* — never the
+/// chained fill parity — so they need distinctness within a call, not global
+/// stability.  Top-of-range keeps them clear of vertex and centre ids.
+const CORNER_ID_A: PointId = PointId::MAX;
+const CORNER_ID_B: PointId = PointId::MAX - 1;
+/// Ring vertices are numbered from here (their global index across the
+/// ring-set), below every centre / corner id.
 const VERTEX_ID_BASE: PointId = 2;
 
 /// A straddle determinant (a scalar triple product of unit vectors) below this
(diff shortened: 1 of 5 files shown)


src_rust/src/sphere.rs
```diff
@@ -421,6 +421,58 @@ pub fn robust_crossing(
...
+/// through [`orient_sos`]'s canonical, identity-keyed evaluation, so the same
+/// physical point resolves to the same side in **every** test that consults it
+/// (e.g. the shared vertex of two incident edges appears as `[p v q]` in one
+/// edge's test and `[q v p]` in the other — exact negations by construction).
+/// A probe passing exactly through a ring vertex therefore counts **exactly
+/// one** crossing across the two incident edges when the boundary passes
+/// through the probe circle (and zero or two when it grazes), keeping the
+/// even-odd fill parity consistent: the half-open `[a, b)` convention emerges
+/// instead of being hand-maintained.  Total and reorder-invariant; `ia, ib,
+/// ic, id` are the SoS identities of the four endpoints.
```
espg: 🤖 *from Claude (review)*

**Low — state the pairwise-distinct-id precondition here.** "Total and reorder-invariant" holds only when `ia, ib, ic, id` are pairwise distinct. With a duplicated id, `orient_sos`'s canonical sort can no longer order the two points and the guarantee fails *silently* in the near-coplanar noise zone: in a 500k fuzz where `c` reused `a`'s id (points distinct, one near the AB great circle), `arcs_cross_sos(a,b,c,d,…) != arcs_cross_sos(c,d,a,b,…)` in 184,761 cases (~37%). [`PointId`]'s doc does say identities must be "distinct and consistently ordered", but this predicate is the surface phase 2 wires call sites to, so a one-line "requires pairwise-distinct ids" in this docstring would make the contract greppable at the point of use.

The converse is solid and worth noting for the same reason: bit-identical **points** with *distinct* ids stay exactly reorder-invariant (verified over 125k shared-endpoint quadruples, 0 failures) — that's the property the incident-edge vertex-graze argument rests on.

---

src_rust/src/sphere/tests.rs
```diff
@@ -887,3 +887,206 @@ fn rotate_about(v: &Vec3, axis: &Vec3, theta: f64) -> Vec3 {
         v[2] * c + k_cross_v[2] * s + axis[2] * k_dot_v * (1.0 - c),
     ]
 }
+
+// ── arcs_cross_sos: the uniform symbolic crossing predicate (issue #103) ──
```
espg, line 891: 🤖 *from Claude (review)*

**Low (convention) — `sphere/tests.rs` is now 1092 lines,** past CLAUDE.md §4's "~1000 lines without prior discussion" guideline, and phases 2–4 plan to add more suites here (parity oracle, seed-PIP, regression matrix). Flagging per the convention rather than as a defect — a quick call from espg would settle it: (a) split the crossing-predicate suites into their own file (e.g. `sphere/tests_crossing.rs`) before phase 2 lands, or (b) treat test modules as exempt from the guideline. Either answer is fine; it just shouldn't grow another 500 lines silently.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of phase 1 (`arcs_cross_sos` + tests) at head 163efcf, verified with an out-of-tree fuzz harness built from this commit's `sphere.rs` against an independent angular-betweenness reference.

**Verified correct:**
- **The 4-sign identity** (S2 `SimpleCrossing`, i.e. `det[a c b], [c b d], [b d a], [d a c]` all one sign): 2,000,000 random quadruples, arc lengths 1e-4 … ~π rad, 1/3 with CD deliberately placed near the AB-midpoint antipode — **0 disagreements** with the geometric reference and 0 with `robust_crossing` on non-degenerate inputs. The antipodal-separation class that fools the bare straddle is rejected by the identity, as the new test claims.
- **Early-return structure** is equivalent to "all four signs equal" because `orient_sos` is total (never 0) — signs are ±1, so the equality chain is exact.
- **Vertex graze counts once**: 200,000 bit-exact meridian (`y == 0`) vertex passages with randomized *distinct* ids — opposite-side incident edges always total exactly 1 crossing; same-side (graze) always even.
- **Collinear-edge parity** (the issue-#103 shape): 200,000 boxes with the west edge exactly on the probe circle, random latitudes and random id assignments — 4-edge crossing count always even. The pinned `[true, false, true, false]` is id-assignment-specific, but the adjacent relabeling-parity test covers the invariant that matters, so the pin works as a regression fixture.
- **Reorder invariance** exact under arc swap/reversal, including bit-identical shared endpoints with distinct ids (625k checks) — see the one inline caveat about duplicated ids.
- The differential grid test exercises all 5,040 quadruples (584 crossing-true), so `checked > 1000` holds with wide margin; its 1e-9 degeneracy filter never fires on this point set (harmless, purely protective).
- Conventions: commit style matches the repo (`phase 1 of issue #103`), no new deps, `cargo fmt --check` clean, clippy introduces no new warnings, full `cargo test` at this commit: 196 passed.

**Findings — both Low, inline:** (1) document the pairwise-distinct-id precondition on `arcs_cross_sos` (silent consistency loss demonstrated when violated); (2) `sphere/tests.rs` crossed the ~1000-line CLAUDE.md §4 guideline and phases 2–4 will add more — needs a split-or-exempt call.

No blocking issues found in phase 1.

---

src_rust/src/coverage.rs
```diff
@@ -535,6 +608,93 @@ fn covers_complement(rings: &[Vec<Vec3>], cap: &Cap) -> bool {
...
+            .expect("a non-antipodal known donor base centre");
+        let parity = arc_crossing_parity(
+            &centers[d as usize],
+            &centers[b as usize],
+            center_id(0, d),
+            center_id(0, b),
+            &all,
+            edges,
+        );
+        fill[b as usize] = fill[d as usize] ^ parity;
```
espg: 🤖 *from Claude (review)*

**[1] Major (verified end-to-end): the donor chain counts phantom antipodal crossings — an ambiguous seed can be chained to the *inverted* fill, flipping its whole base subtree in release builds.**

`edge_crosses_probe`'s fast path is the bare two-straddle
```rust
(d_pq_a > 0.0) != (d_pq_b > 0.0) && (d_ab_p > 0.0) != (d_ab_q > 0.0)
```
which is satisfied not only by a real crossing but also when the edge sits on the **far** intersection of the two great circles (edge contains `y`, probe contains `-y`). The descent never sees this because its probes are short (centre→child/corner) and the edge list is relevance-culled, so no surviving edge is antipodal to a probe. `base_fills` breaks both shields at once: it feeds base-centre chords up to ~122° long over **all** edges (`let all = (0..edges.len()).collect()`).

Per-edge, the phantom fires exactly when the edge crosses the *antipodal image* of the probe arc; summed over a closed ring, the chained parity is inverted iff the polygon **separates the antipodes of the two chain endpoints** — for the `c1 → c0` chain, iff exactly one of base centres 10 and 11 is inside the polygon.

Reproducer (verified against 4412cfd; ring vertices `(lat, lon)`):
```
(10,45) (50,45) (-10,170) (-70,225) (-10,280)
```
- the `(10,45)→(50,45)` meridian edge makes `seed_fill(c0)` return `None` → donor chain from centre 1;
- `arc_crossing_parity(c1, c0, …, &all, …)` returns `false`, while the same count through `arcs_cross_sos` per edge is `true` (two edges of this ring cross the probe's antipodal arc region; one of them phantoms);
- **debug build**: `polygon_to_morton_coverage(&lats, &lons, 4, true)` panics on this phase's own parity oracle — `parity oracle diverged at uniform cell (15, 2)` (and others) — the oracle works exactly as designed;
- **release build**: silently wrong output. Same polygon with the meridian nudged to `lon = 45.000001` (no ambiguity, no chain): symmetric difference of **8221 cells out of ~22k** at order 6.

The phase-3 matrix can't catch this: its reproducers are small boxes, whose boundaries never reach the antipodal arc of a donor chord (`M = 0`, even, parity survives).

Suggested fix: in the donor chain, route every edge through `arcs_cross_sos` unconditionally instead of `edge_crosses_probe` — worst case 12 seeds × E edges, once per call, off the hot path, and `arcs_cross_sos` handles the antipodal configuration correctly (I verified its verdicts on these probes against dense arc sampling). Two related notes while in here: (a) `known[b] = true` after chaining makes a chained seed eligible as a *donor* for later seeds, so an inverted chain compounds; (b) `seed_fill` measures ambiguity against each edge's **infinite** great circle, so seeds get chained more often than strictly necessary — harmless once the chain itself is exact, but today it widens the exposure to this bug.

---

src_rust/src/coverage.rs
```diff
@@ -478,12 +498,65 @@ fn edge_crosses_probe(p: &Vec3, q: &Vec3, n_pq: &Vec3, e: &Edge) -> bool {
...
+        return true;
+    }
+    if d3 == 0.0 || d4 == 0.0 {
+        let mid = normalize(&[c1[0] + c2[0], c1[1] + c2[1], c1[2] + c2[2]]);
+        let cos_rho = dot(&mid, c1);
+        if (d3 == 0.0 && dot(&mid, &e.a) >= cos_rho - ORIENT_EPS)
+            || (d4 == 0.0 && dot(&mid, &e.b) >= cos_rho - ORIENT_EPS)
+        {
+            return true;
+        }
```
espg: 🤖 *from Claude (review)*

**[2] Medium (gate-level verified): the segment gate reuses `ORIENT_EPS` — a triple-product magnitude — as a cos-of-angle slack, and the angular window blows up as `sin_rho → 0`.**

`dot(mid, x) >= cos_rho - ORIENT_EPS` admits points up to `Δ` *beyond* the segment endpoint, where `Δ·sin_rho + Δ²/2 ≤ 1e-12`, i.e. `Δ ≈ 1e-12/sin_rho` for long segments but a floor of `Δ ≈ √(2e-12) ≈ 1.4e-6 rad` for short ones. Measured against the committed formulas:

- **d3/d4 branch** (polygon endpoint vs cell-edge cap): for an order-29-scale cell edge (~1.5e-9 rad), `cos_rho` rounds to `1.0` and `sin_rho` to `0.0`, so the gate degrades to a pure absolute window — a polygon vertex bit-exactly on the meridian passes as "within the segment" from **1.4e-6 rad away, ≈ 10³ cell-widths**.
- **d1/d2 branch**: a cell corner ~0.8e-6 rad past the endpoint of a 1.7e-6-rad polygon edge still gates as a touch.

Two mitigating facts I confirmed: bit-exact `d == 0.0` incidence needs `lon ≡ 0 (mod 90°)` (at lon 45, `cos45°`/`sin45°` differ by 1 ulp, `d ≈ 1e-22 ≠ 0`, and those correctly route to the `arcs_cross_sos` branch below); and `edge_relevant`'s own 1e-12 cos-slack culls the distant cells again unless the *polygon* edge is also short (`≲ 2e-6 rad ≈ 13 m`). So the end-to-end exposure is: finely segmentized boundaries lying on an exact 0/90/180/270 meridian at order ≳ 21 — narrow, but squarely inside the on-grid family this PR is about, and it inflates the agreed closed-set contract ("touches *exactly*") into "touches, or is within ~1.4e-6 rad along the great circle past the endpoint". I verified the gate arithmetic only, not a full-coverage strip.

Suggestion: keep the slack in angle space — e.g. require `dot(mid, x) >= cos_rho - K·sin_rho` with `K` an angular epsilon (plus an exact tie at the endpoint itself), or decide segment membership with the two endpoint half-plane signs through the exact `orient` machinery this PR already adds, which needs no epsilon at all.

---

src_rust/src/sphere.rs
espg, line 1: 🤖 *from Claude (review)*

**[3] Minor: deleting `robust_crossing` left a broken intra-doc link.** `cargo doc --no-deps` now emits a new warning (`rustdoc::broken_intra_doc_links`):

```
warning: unresolved link to `robust_crossing`
   --> src_rust/src/sphere.rs:484:55
484 | /// The companion SoS predicates [`orient_sos`] and [`robust_crossing`] are the
```

That paragraph on `point_in_ring_robust` also still describes the retired pipeline as a companion ("building blocks the descent's per-cell parity flips will use in a later phase") — worth rewording to `arcs_cross_sos` while fixing the link. Cosmetic and optional: the `test_robust_crossing_*` test names in `sphere/tests.rs` now exercise `arcs_cross_sos`; a rename would keep future greps honest.

Everything else I attacked in this file held up — details in the review summary: `two_sum`/`two_product`/`exact_sum_sign` verified against an exact rational oracle on 20k adversarial trials (0 mismatches), the `1e-15` filter bound verified analytically (~6.7e-16 worst case for this evaluation order) and empirically (max observed noise 1.85e-16·mag over 200k trials).

---

espg (commented):
🤖 *from Claude (review)*

**Fresh-context adversarial review of the phase-2 commit (`163efcf..4412cfd`, "phase 2 of issue #103").** I attacked the exact arithmetic, the SoS id scheme, the seed chaining, the closed-set gate, and the parity oracle, verifying every claim by running code (scratch worktree at 4412cfd; nothing in the repo was modified). Three findings, ranked; the inline comments carry the reproducers.

**Findings**

1. **Major — `base_fills` donor chain counts phantom antipodal crossings** (inline on `base_fills`). `edge_crosses_probe`'s fast path is the bare two-straddle, sound in the descent only because probes are short and edges relevance-culled; the donor chain feeds it ~122° base-centre chords over *all* edges. A polygon that separates the antipodes of the chain endpoints gets an inverted seed → whole base subtree mis-filled in release; in debug your own parity oracle catches it (which is exactly the oracle earning its keep). Verified end-to-end with a 5-vertex reproducer: debug panics at `parity oracle diverged at uniform cell (15, 2)`; release differs from the 1e-6°-nudged control by 8221 of ~22k cells at order 6. Fix is cheap: use `arcs_cross_sos` unconditionally in the chain.
2. **Medium — `edge_hits_cell_edge` segment gate**: `ORIENT_EPS` (a triple-product magnitude) reused as cos-of-angle slack; angular window floors at ~1.4e-6 rad for short segments (≈10³ cell-widths at order 29). Gate-level verified; end-to-end exposure limited to short on-meridian edges at high order (inline).
3. **Minor — stale doc**: new `rustdoc` broken-intra-doc-link warning for the deleted `robust_crossing` at `sphere.rs:484` (inline).

**What held up under attack** (all verified, not just read):

- `two_sum` is the branch-free Knuth TwoSum, exact. `two_product` via `f64::mul_add` is exact (Rust guarantees single rounding; residual representable barring denormal underflow, unreachable for unit vectors).
- `exact_sum_sign` is GROW-EXPANSION (Shewchuk Thm. 10) with zero elimination — expansion kept nonoverlapping and ascending, sign = last component. **20,000 adversarial trials** (exact-cancelling pairs, mixed exponents −60..60, up to 24 terms) against an exact rational oracle: **0 mismatches**. Same for the 24-term `orient_exact_sign` fallback on near-coplanar triples.
- The `mag * 1e-15` filter: for `dot(a, cross(b, c))` evaluation order the worst-case rounding error is ~6u·mag ≈ 6.7e-16·mag (u = 2⁻⁵³); empirical max over 200k trials 1.85e-16·mag. So 1e-15 is a safe overestimate — margin ~1.5×, so "comfortably above" in the comment is a mild overstatement, but the accept side is sound and the reject side is exact.
- `antipodal_base`: verified against `cell_center_vec(0, ..)` — all six pairs dot to < −1 + 1e-12, all 60 other pairs are far from antipodal.
- `center_id`: exhaustively unique over depths 0–6, top bit set, and the depth-29 maximum stays clear of `CORNER_ID_A/B`. The stable-id design (rings perturb first, one global perturbation) is the right call and the doc explains why.
- Parity oracle: no side effects in the assert expression; fully compiled out in release; I found no spurious-fire mechanism (a uniform cell's centre is ≥ its inradius from the boundary — any closer and the closed-set straddle test refines the cell first — so the winding verdict there is trustworthy; hemisphere+/multipart use the same `parity_filled_robust` backend on both sides). It caught finding (1) instantly.
- `seed_fill`'s 1e-12 plane gate: winding noise at the trust boundary is ~1e-16 rad-equivalent, four orders inside the gate — no false-trust window. 12 seeds × E normalizes is negligible.
- `cargo fmt --check` clean; `cargo clippy` exits 0 (all warnings pre-existing in `lib.rs`/old tests); 196 tests pass; commit style matches; no new dependencies (the ~90-line expansion kernel honors the no-new-deps agreement).

**Conventions**: `coverage.rs` is now **1136 lines**, past the ~1000 soft limit — and `coverage/tests.rs`'s own header says the tests were split out precisely to stay under it. The PR body's "Questions for review" flags `sphere/tests.rs` but not this; same treatment question applies (the `seed_fill`/`base_fills`/`antipodal_base` block or the edge predicates are natural split seams). Also: the new kernels (`exact_sum_sign`, `orient_exact_sign`, `minor_exact_sign`, `edge_hits_cell_edge`, `seed_fill`/`base_fills`, `center_id`) have no direct Rust unit tests — the phase-3 matrix exercises them only through small-box geometries, which is exactly why finding (1) slipped past 196 green tests. A regression test derived from the inline reproducer would pin both the fix and the oracle.

Findings (2) and (3) are non-blocking for the next phase; (1) is a correctness bug in shipped release behavior and worth a phase commit before the PR leaves draft.

---

src_rust/src/coverage.rs
espg: 🤖 *from Claude*

Folded in phase 4: the donor chain now routes every edge through `arcs_cross_sos` unconditionally (no `edge_crosses_probe`, no fast path), and your reproducer ring is pinned as `test_base_fills_chain_no_antipodal_phantom` — every winding-classifiable seed must agree with the chain. On your two notes: (a) donor eligibility of chained seeds is now benign — the chain is exact, so a chained seed carries the exact verdict and compounds nothing; (b) the infinite-great-circle ambiguity test stays deliberately wide since chaining is now always correct and the cost is 11 arcs at most.

One thing your reproducer surfaced *beyond* this finding: even with the chain fixed, the full descent on that ring still trips the parity oracle — the 40°-long `(10,45)→(50,45)` edge is collinear with the probe lattice, and the vertex-graze bookkeeping between the collinear edge and its transversal sibling desynchronizes on the hemisphere+ ring. That is the long-arc territory `sphere.rs`'s own comments defer, it predates this PR (the oracle is what made it visible), and it's documented as an `#[ignore]`d test (`test_descent_hemisphere_ring_collinear_edge_oracle`) plus a question for review on the PR.

---

src_rust/src/coverage.rs
espg: 🤖 *from Claude*

Folded in phase 4: the slack is now angular — `dot ≥ cos_ρ − sin_ρ·ε − ε²` (both branches, with the cell-edge branch computing its own `sin_ρ`), so the window stays ~1e-12 rad for short segments instead of flooring at √(2e-12). The endpoint-exact tie still gates (the −ε² term covers cos rounding at ρ→0). Your no-epsilon alternative (endpoint half-plane signs through the exact machinery) is the cleaner endgame; noted on the PR as the follow-up shape if the closed-set contract ever needs to be tolerance-free end-to-end.

---

src_rust/src/sphere.rs
espg: 🤖 *from Claude*

Folded in phase 4: the docstring now states the pairwise-distinct-id precondition at the point of use (and notes that bit-identical *points* with distinct ids remain exactly invariant — the property the vertex-graze argument rests on).

---

src_rust/src/sphere.rs
espg: 🤖 *from Claude*

Doc link and the stale "companion pipeline" paragraph fixed in phase 4 (now points at `arcs_cross_sos`). The `test_robust_crossing_*` names are left standing deliberately — they pin behavioral continuity across the swap; happy to rename in the 1.x freeze sweep if you'd rather keep greps clean.

---

espg:
🤖 *from Claude*

All four phases complete; local gates green (197 cargo tests with the parity oracle live, 635 pytest incl. the 87-test boundary matrix, clippy/fmt/flake8 clean); both self-review rounds' findings folded or answered on-thread. Marking `waiting` for your review — kept **draft** per the routine's drafts-only rule (same note as PR #105). Three items under "Questions for review": the deliberately-out-of-scope hemisphere+/collinear-edge limitation the oracle uncovered (recommend a follow-up issue — yours to open), the module-size call, and the #90 connection. Watch the CodSpeed report for the exact-predicate perf gate; the exact path is filter-guarded so clean geometry shouldn't move.

---

espg:
> Known limitation discovered by the oracle, deliberately out of scope (documented as the #[ignore]d test_descent_hemisphere_ring_collinear_edge_oracle): on a hemisphere+ ring whose 40°-long edge is collinear with the probe lattice (the phase-2 reviewer's adversarial ring (10,45),(50,45),(-10,170),(-70,225),(-10,280)), the descent's vertex-graze bookkeeping between the collinear edge and its transversal sibling can still desynchronize. This predates the PR — the old noise-sign machinery was strictly weaker on the same family; the new oracle is what made it visible — and it's the long-arc consistency work sphere.rs's own comments defer. Recommend a follow-up issue (opening one is yours to call); happy to write it up with the full diagnosis.

yes, write it up with full diagnosis as a new issue

> Module sizes: sphere/tests.rs (~1250 lines) and coverage.rs (~1180 lines) are past the ~1000 guideline. Both grew from this fix's tests/kernels. Split now (mechanical) or defer to the 1.x freeze sweep (https://github.com/espg/mortie/issues/68)? Your call either way.

we can defer on this

> Re https://github.com/espg/mortie/issues/90: with the disambiguation now exact end-to-end, the node_straddles over-refinement is the only remaining conservatism; the measurement https://github.com/espg/mortie/issues/90 calls for can now be done meaningfully. Left out of scope here per the original agreement.

I'm confused-- what are you saying here? Is this something that's made possible due to the work in this PR? or is #90 no longer needed and be closed because of this PR? or... is it irrelevant and orthogonal? What's your specific recommendation?

espg:
🤖 *from Claude*

> What's your specific recommendation [on #90]?

Concretely: **#90 stays open, and this PR is what makes it actionable — recommend keeping it, re-scoped one notch.** Before this PR, the `node_straddles` over-refinement was load-bearing: it was the safety net that papered over the crossing predicate's residual float sensitivity (the old `coverage.rs:442` caveat), so relaxing it risked exposing wrong fills. After this PR the predicate is exact end-to-end, so the net is no longer hiding correctness bugs — it's now *pure conservatism*, and #90's "measure the over-refinement, then decide whether to relax" program can be run safely, with the parity oracle as the guard rail the measurement never had. Two adjustments to #90's text when you next touch it: (a) the measurement baseline should be this branch, not pre-#87 main; (b) one nuance — the closed-set contract from #103 *deliberately adds* boundary-touch refinement, so the relaxation target is the non-touching straddle over-refinement only. Not closable, not orthogonal: unblocked.

Separately: CodSpeed's −19% on clean-geometry coverage benches looks real (25 benchmarks, consistent direction), which contradicts my "filter-guarded, clean geometry shouldn't move" claim — investigating now; prime suspect is `edge_hits_cell_edge` losing `arcs_cross_n`'s two-dot early exit on the quad-test hot path. Fix lands as phase 5.

---

espg:
🤖 *from Claude*

**Final perf report** (from the [merge-commit benchmark run](https://github.com/espg/mortie/actions/runs/29074457518) + the branch head's CodSpeed analysis): the merged state measures **−17.67% overall** — 14 regressed coverage benches, worst `coverage_triangle_polar[8]` at −31.7%, `coverage_circle500[6]` −16.1%. Phase 5 recovered only part of the original −19.4%.

Callgrind attribution found the two real causes — neither is the exact-arithmetic layer (filter-guarded; absent from every profile):
1. `base_fills` classified **all 12** base seeds per call (12 winding PIPs + a 12×E plane scan) where the old `base_node` paid only for cap-surviving bases (1–3 typical);
2. `edge_hits_cell_edge` grew past the inline threshold, so the quad-test hot loop paid a real function call per test.

Both fixed in follow-up **PR #109** (draft, `implement`): seed classification respects the cap cull, and the hot shell is `#[inline(always)]` with the closed-set logic outlined `#[cold]`. Local callgrind puts the triangle case at **+1%** vs pre-#106 and the 500-vertex circle at **−5%** (below baseline); CodSpeed on #109 will confirm. Semantics untouched — the full #103 boundary matrix passes unchanged.

---

Issue #107, opened 2026-07-10
Descent parity desynchronizes on hemisphere+ rings with a long edge collinear to the probe lattice


espg:
🤖 *from Claude*

Follow-up from PR #106 (Refs #103), filed on espg's go-ahead ([comment](https://github.com/espg/mortie/pull/106#issuecomment-4932513946)). Pre-existing behavior made *visible* by #106's parity oracle — the old noise-sign machinery was strictly weaker on the same family. Not triggered by any workload reported so far; discovered by the phase-2 adversarial self-review.

## Reproducer

Ring vertices `(lat, lon)`: `(10,45), (50,45), (-10,170), (-70,225), (-10,280)` — hemisphere+ (winding taken as-given), with the first edge running **40° along the lon-45 meridian**, i.e. collinear with the HEALPix probe lattice (base-0 centres and their descendants sit on that same near-plane, 1 ulp off coplanar).

- **Debug build** (with #106's parity oracle): `polygon_to_morton_coverage(&lats, &lons, 6, true)` panics — `parity oracle diverged at uniform cell (51, 3)` (and others).
- **Release build**: silently wrong coverage; vs the same ring with the meridian edge nudged to `lon = 45.000001`, the symmetric difference is thousands of cells at order 6 (measured 8221/~22k on the pre-phase-4 head; same order of magnitude after).
- Pinned as the `#[ignore]`d test `test_descent_hemisphere_ring_collinear_edge_oracle` in `src_rust/src/coverage/tests.rs` — un-ignore when this lands.

## Diagnosis (from the PR #106 debugging session)

The seed layer is **not** the problem — after PR #106 phase 4, `base_fills`' donor chain is fully symbolic and every winding-classifiable seed agrees with it (pinned by `test_base_fills_chain_no_antipodal_phantom`).

The break is in the descent's fill chain, localized by leg-by-leg replay:

- Seed `base 0` (centre `(41.81, 45)`, exactly on the collinear edge's segment) chains to **outside** — consistent: the donor chord from centre 1 (winding +2π, inside) crosses edge 1 once.
- Leg `centre(0,0) → centre(3,1)` — both endpoints on the lon-45 near-plane, i.e. the probe is **collinear with the 40° edge** — counts exactly **one** crossing (edge 1, the transversal sibling departing the shared vertex `(50,45)`); the collinear edge 0 reports no crossing.
[...]

PR #109, opened 2026-07-10
Recover coverage perf after the #103 predicate swap

espg:
Refs #103, PR #106. Follow-up to the merged predicate swap: CodSpeed measured **−17.67%** on the phase-5 head (14 regressed coverage benches, worst `coverage_triangle_polar[8]` −31.7%). This PR recovers it; callgrind instruction counts are back at — and for polygon-heavy cases below — the pre-#106 baseline.

## Diagnosis (callgrind, symbolized builds, instruction counts)

Two independent fixed costs, neither the exact-arithmetic layer (which is filter-guarded and never appeared in any profile):

1. **`base_fills` classified all 12 base seeds unconditionally.** Pre-#106, `base_node` computed the O(V) winding PIP only for bases surviving the vertex-cap cull (1–3 for typical polygons); the #106 seed hardening moved classification to `base_fills`, which paid 12 winding PIPs + a 12×E plane scan every call. Profile: 667k instr/call on a 500-vertex circle — the dominant regression for every small/medium polygon.
2. **`edge_hits_cell_edge` fell out of inlining.** The closed-set degenerate branch pushed the function past the inline threshold, so the quad-test hot loop paid a real call per (edge × cell-edge) test where `arcs_cross_n` had been fully inlined. Visible as a standalone symbol costing ~2× the old inlined test.

## Fixes

- `base_fills` now takes the cap/complement and **skips cap-culled bases** (marked known, fill unused — `base_node` returns `None` for them regardless; and a culled base used as a donor carries `fill = false`, which is exact: its centre is provably outside a non-complement polygon's vertex cap). Non-complement small polygons drop from 12 PIPs back to the pre-#106 1–3.
- `edge_hits_cell_edge` split: an `#[inline(always)]` hot shell in the exact two-dot shape of the retired `arcs_cross_n`, with the closed-set/symbolic logic outlined into a `#[cold] #[inline(never)]` helper. Semantics unchanged (same tests pin both paths).

## Measured (callgrind, mortie-attributed instructions)

| case | pre-#106 | post-#106 (merged) | this PR |
|---|---|---|---|
| triangle order 4 (×20) | 334k | 401k (+20%) | **338k (+1%)** |
| 500-vtx circle order 6 (×3) | ~7.4M | ~9.6M (+30%) | **~7.0M (−5%)** |

(The circle lands *below* baseline because the phase-5 hot-exit restoration already outperformed the old always-4-dot `arcs_cross_n` shape in the straddle loop.) CodSpeed on this PR is the authoritative confirmation — expecting the coverage benches to clear.

## How it was tested

- `cargo test` — 197 passed (parity oracle live), 1 `#[ignore]`d (#107); clippy/fmt clean.
- `maturin develop --release` + `pytest` — 670 passed, 11 skipped (includes the 87-test #103 boundary matrix — closed-set semantics unchanged); `flake8` clean.

## Questions for review

None — behavior-preserving; the one semantic already documented on #106 (vertex-point-touch fine print) is untouched.

---
_Generated by [Claude Code](https://claude.ai/code/session_01X3JyUn7YUxZ294nxbu8Jw8)_

### DIFF
--- a/src_rust/src/coverage.rs
+++ b/src_rust/src/coverage.rs
@@ -532,7 +532,7 @@ fn edge_crosses_probe(p: &Vec3, q: &Vec3, ip: PointId, iq: PointId, n_pq: &Vec3,
 /// Cell corners are one-shot derived points ([`CORNER_ID_A`]/[`CORNER_ID_B`]):
 /// only distinctness matters within a call — this feeds the straddle boolean,
 /// never the chained fill parity.
-#[inline]
+#[inline(always)]
 fn edge_hits_cell_edge(e: &Edge, c1: &Vec3, c2: &Vec3, n_c: &Vec3) -> bool {
     let d1 = dot(&e.n_ab, c1);
     let d2 = dot(&e.n_ab, c2);
@@ -543,31 +543,51 @@ fn edge_hits_cell_edge(e: &Edge, c1: &Vec3, c2: &Vec3, n_c: &Vec3) -> bool {
     let d3 = dot(n_c, &e.a);
     let d4 = dot(n_c, &e.b);
     if degenerate12 || d3.abs() < ORIENT_EPS || d4.abs() < ORIENT_EPS {
-        // Exact incidence ⇒ closed-set touch, gated to the segment span (the
-        // infinite great circle runs to the poles; gating on it alone emitted
-        // a spurious strip of cells far beyond the polygon).
-        let span = |cos_rho: f64, sin_rho: f64, d: f64| {
-            d >= cos_rho - sin_rho * ORIENT_EPS - ORIENT_EPS * ORIENT_EPS
-        };
-        if (d1 == 0.0 && span(e.cos_rho, e.sin_rho, dot(&e.mid, c1)))
-            || (d2 == 0.0 && span(e.cos_rho, e.sin_rho, dot(&e.mid, c2)))
-        {
-            return true;
-        }
-        if d3 == 0.0 || d4 == 0.0 {
-            let mid = normalize(&[c1[0] + c2[0], c1[1] + c2[1], c1[2] + c2[2]]);
-            let cos_rho = dot(&mid, c1);
-            let sin_rho = (1.0 - cos_rho * cos_rho).max(0.0).sqrt();
-            if (d3 == 0.0 && span(cos_rho, sin_rho, dot(&mid, &e.a)))
-                || (d4 == 0.0 && span(cos_rho, sin_rho, dot(&mid, &e.b)))
-            {
-                return true;
-            }
-        }
-        return arcs_cross_sos(c1, c2, &e.a, &e.b, CORNER_ID_A, CORNER_ID_B, e.ia, e.ib);
+        return edge_touches_cell_edge_degenerate(e, c1, c2, d1, d2, d3, d4);
     }
     // d1, d2 straddle (established above); the plain sign test decides.
     (d3 > 0.0) != (d4 > 0.0)
+}
+
+/// Cold half of [`edge_hits_cell_edge`]: the closed-set / symbolic logic for
+/// degenerate configurations.  Outlined (`#[cold]`) so the hot shell above
+/// stays small enough to inline into the quad loop — as a single function the
+/// degenerate branch pushed it past the inlining threshold and the per-test
+/// call overhead showed up as a CodSpeed regression on every coverage bench.
+#[cold]
+#[inline(never)]
+#[allow(clippy::too_many_arguments)]
+fn edge_touches_cell_edge_degenerate(
+    e: &Edge,
+    c1: &Vec3,
+    c2: &Vec3,
+    d1: f64,
+    d2: f64,
+    d3: f64,
+    d4: f64,
+) -> bool {
+    // Exact incidence ⇒ closed-set touch, gated to the segment span (the
+    // infinite great circle runs to the poles; gating on it alone emitted
+    // a spurious strip of cells far beyond the polygon).
+    let span = |cos_rho: f64, sin_rho: f64, d: f64| {
+        d >= cos_rho - sin_rho * ORIENT_EPS - ORIENT_EPS * ORIENT_EPS
+    };
+    if (d1 == 0.0 && span(e.cos_rho, e.sin_rho, dot(&e.mid, c1)))
+        || (d2 == 0.0 && span(e.cos_rho, e.sin_rho, dot(&e.mid, c2)))
+    {
+        return true;
+    }
+    if d3 == 0.0 || d4 == 0.0 {
+        let mid = normalize(&[c1[0] + c2[0], c1[1] + c2[1], c1[2] + c2[2]]);
+        let cos_rho = dot(&mid, c1);
+        let sin_rho = (1.0 - cos_rho * cos_rho).max(0.0).sqrt();
+        if (d3 == 0.0 && span(cos_rho, sin_rho, dot(&mid, &e.a)))
+            || (d4 == 0.0 && span(cos_rho, sin_rho, dot(&mid, &e.b)))
+        {
+            return true;
+        }
+    }
+    arcs_cross_sos(c1, c2, &e.a, &e.b, CORNER_ID_A, CORNER_ID_B, e.ia, e.ib)
 }
 
 /// A descent node: cell `(pixel, depth)`, its centre, even-odd fill state, and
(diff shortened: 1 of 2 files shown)


src_rust/src/coverage.rs
```diff
@@ -687,6 +721,11 @@ fn base_fills(edges: &[Edge], rings: &[Vec<Vec3>]) -> [bool; 12] {
         }
         return fill;
     }
+    // NOTE: a culled base is marked known (fill unused), so the donor search
+    // below can only pick a *classified* seed when at least one un-culled,
+    // unambiguous seed exists; if every un-culled seed was ambiguous, the
+    // donors include culled entries whose fill is false — the correct verdict
+    // for a base whose cap the polygon's vertices cannot reach.
```
espg, line 728: 🤖 *from Claude (review)*

**(1) The culled-donor exactness claim is airtight only for `cap.radius ≤ 90°` — suggest gating the skip on that.**

"the correct verdict for a base whose cap the polygon's vertices cannot reach" needs *interior ⊆ vertex cap*. That holds iff the cap is geodesically convex, i.e. `radius ≤ 90°`: then every edge minor arc stays inside the cap, the cap-complement is connected and boundary-free, and `covers_complement`'s antipode probe returning false makes it all-exterior — so an `excludes`-culled centre (> `cr` beyond the cap) is *macroscopically* outside, exact in both the numeric and the SoS-perturbed world. Airtight.

For `radius > 90°` a minor arc between two in-cap vertices can exit the cap by up to 90° — more than the `radius + cr` cull margin (base `cr` = 45–48.19°) — so an edge **segment** can pass exactly through a culled base centre. Its "outside" then rests on `covers_complement`'s winding probe of that centre (an on-boundary tie, the very tie #103 exists for), and the donor path now trusts it, where the old code re-derived every non-`seed_fill`-classified base through the SoS chain.

I attacked this concretely and could **not** break it:

- targeted family: a 170° edge running exactly through the base-0 and base-1 centres (that great circle also contains the centres of bases 4, 6, 10, 11 — six ambiguous seeds), ballast cluster near base-0's antipode so `radius ≈ 95–105°`; base 0 is culled and, being index 0, is always the donor. 63/300 configs exercised the culled-donor path with `complement == false`: zero divergence from `main`'s `base_fills` on all un-culled entries;
- realistic family: grid-aligned small boxes with one edge on the equator (far bases plane-ambiguous *and* culled): 360 exercised configs, zero divergence;
- end-to-end: `rust_polygon_coverage` from a main-built wheel vs this branch on 60 adversarial rings — bit-identical outputs.

Why it survives: with lat/lon inputs the incidence is ~1e-17 but never bit-exact, so `arcs_cross_sos` decides by the true sign and the winding backend's ±π tie follows the same geometric side. The residual exposure is the **bit-exact** family (`dot(n̂, centre) == 0.0` under a hemisphere+ vertex cap — the on-grid inputs #103 came from), where the symbolic resolution decouples from the winding probe. Cheap hardening that keeps 100% of the measured win (the 12→1–3 PIP saving is a small-polygon effect): take the `continue` only when `cap.radius <= FRAC_PI_2`. That makes this NOTE unconditional. Since the PR is merged, fine as a follow-up issue — flagging so it's recorded.

---

src_rust/src/coverage.rs
```diff
@@ -664,17 +684,31 @@ fn antipodal_base(a: u64, b: u64) -> bool {
...
+    for b in 0..12u64 {
+        if !complement {
+            let center = &centers[b as usize];
+            let corners = cell_corners(0, b);
+            let (cos_cr, sin_cr) = cell_cos_radius(center, &corners);
+            if cap.excludes(center, cos_cr, sin_cr) {
+                known[b as usize] = true; // culled: fill value is never read
+                continue;
+            }
+        }
```
espg, line 708: 🤖 *from Claude (review)*

**(2) Nit — net-new panic sliver in the donor `.expect()` below.**

Pre-marking culled bases `known` can leave exactly one known entry where the old code had zero: a base that is culled **and** plane-ambiguous no longer falls through to the `all(|&k| !k)` raw-winding fallback. If that single known base is base 10 while base 0 is ambiguous, the donor search for `b = 0` sees only an antipodal candidate and `.expect("a non-antipodal known donor base centre")` panics; the old code took the fallback instead. Reachable only when all 12 centres are plane-ambiguous (e.g. two edges on the two great circles that together carry all 12 base centres) with exactly base 10 culled — deep pathological territory, but a panic crosses the PyO3 boundary as an abort of the user call, so a one-line guard (fall back to raw winding when `find` comes up empty) or a comment noting the reasoning would be cheap insurance.

---

src_rust/src/coverage/tests.rs
```diff
@@ -572,7 +572,9 @@ fn test_base_fills_chain_no_antipodal_phantom() {
     let lons = vec![45.0, 45.0, 170.0, 225.0, 280.0];
     let rings = build_rings(&[lats], &[lons], true);
     let edges = build_edges(&rings, 6);
-    let fills = base_fills(&edges, &rings);
+    let cap = Cap::of_rings(&rings);
+    let complement = covers_complement(&rings, &cap);
+    let fills = base_fills(&edges, &rings, &cap, complement);
```
espg, line 577: 🤖 *from Claude (review)*

**(3) No test pins the new cull-skip branch.** The only test change here adapts an existing call site to the new signature; nothing exercises the culled-base-as-donor behavior, so a future edit to the donor selection (e.g. preferring a different `known` index) would go uncaught. A cheap, deterministic pin: assert `base_fills(edges, rings, cap, complement)` agrees with the classification computed without culling on all **un-culled** entries, over a family that actually takes the skip — grid-aligned small boxes with one edge on the equator do (far bases are simultaneously cap-culled and plane-ambiguous, so the culled base is chosen as donor; 360/360 such configs agreed with `main`'s implementation in my fuzz, <0.1 s runtime). Happy to hand the fuzz over for a follow-up PR.

Also for the record: this ring (lat span −70°..50°) lands `complement == true` or leaves the relevant bases un-culled, so the existing assertions retain their original strength — verified the whole matrix still passes (`cargo test` debug + release: 197 passed; `pytest mortie/tests/test_coverage_boundary.py`: 87 passed).

---

espg (commented):
🤖 *from Claude (review)*

Adversarial post-merge review of 3a5fffa (fresh context). Verdict: **no verified correctness bug**; the perf recovery is sound as merged. Findings ranked in the inline comments — (1) is the one worth a follow-up.

**What was attacked and verified:**

- **Culled-donor claim** — the load-bearing claim ("fill = false is exact for a cap-culled base"). Proven airtight for `cap.radius ≤ 90°` (convex cap ⇒ boundary ⊆ cap ⇒ culled centre macroscopically exterior in both the numeric and SoS worlds). For hemisphere+ vertex caps there is a theoretical window (edge arcs can exit the cap past the `radius + cr` margin and run exactly through a culled centre); I built that geometry deliberately — a 170° edge through the base-0 and base-1 centres with the cap axis near base-0's antipode — and differential-tested this branch's `base_fills` against `main`'s verbatim implementation: 63 exercised targeted configs + 360 exercised realistic grid-aligned configs, zero divergence on un-culled entries; plus bit-identical `rust_polygon_coverage` output between a main-built wheel and this branch across 60 adversarial rings. Residual exposure is confined to bit-exact plane incidence under a hemisphere+ cap; inline comment (1) has a one-line gate that closes it with no perf cost.
- **All-ambiguous fallback** — with culled bases pre-marked known, the `all(!known)` branch is unreachable whenever any base is culled; the replacement behavior (ambiguous seeds chained via SoS from a culled `fill = false` donor) is *stronger* than the old raw-winding fallback, given (1). One net-new pathological panic sliver in the donor `.expect()` — inline comment (2).
- **Inline split** — `edge_touches_cell_edge_degenerate` is a verbatim outlining of the old degenerate block: same span closure, same `d1==0.0`/`d2==0.0` then `d3/d4` order, same `arcs_cross_sos(c1, c2, e.a, e.b, CORNER_ID_A, CORNER_ID_B, e.ia, e.ib)` tail; arguments pass through in order. No behavior drift; `#[inline(always)]`/`#[cold]` are attributes only.
- **Suites** — `cargo test` in **debug** (debug_asserts live) and release: 197 passed, 1 pre-existing ignore; `pytest mortie/tests/test_coverage_boundary.py`: 87 passed; `cargo clippy --all-targets`: no lints in the touched code (all 37+ warnings pre-existing elsewhere); `cargo fmt --check` clean.
- **Conventions** — commit is title-only matching repo style; no new dependencies; module stays under the size threshold.

---

PR #110, opened 2026-07-10
Gate the seed cull to sub-hemisphere caps

espg:
Refs PR #109 (merged before its adversarial self-review finished; this folds the review's findings — [review thread](https://github.com/espg/mortie/pull/109)).

## What this does

The #109 review verified the cull-skip's "culled ⇒ provably outside ⇒ `fill = false` is an exact donor verdict" argument is airtight **only for vertex caps ≤ 90°**: cap convexity bounds the polygon boundary inside the cap for sub-hemisphere caps, but a minor arc between in-cap vertices can exit a >90° cap by up to a quadrant and pass exactly through a culled base centre. The review's differential fuzz (423 adversarial configs vs the pre-#109 `base_fills`, plus 60 end-to-end rings vs a main-built wheel) found **zero live divergence** — the exposure is the bit-exact incidence family only — but the fix is free:

1. **The cull-skip now engages only when `cap.radius ≤ π/2`** (and not complement, as before). Oversize-cap polygons classify all 12 seeds — the pre-#109 behavior — which costs nothing in practice: polygons with >90° vertex caps are the rare, already-expensive case.
2. **The donor `.expect()` panic sliver is gone**: if every candidate donor is ambiguous and the only known seed is antipodal to `b` (a pathological configuration the review constructed), the seed falls back to the raw winding verdict instead of panicking.
3. **Regression test**: `test_base_fills_oversize_cap_classifies_all_seeds` pins a 98.5°-cap ring — every winding-classifiable seed must carry its own verdict under the disengaged cull.

## How it was tested

- `cargo test` — 198 passed (parity oracle live), 1 `#[ignore]`d (#107); clippy/fmt clean.
- `maturin develop --release` + `pytest` — 670 passed, 11 skipped; `flake8` clean.

## Questions for review

None — a strictly-tightening guard; perf-neutral for every benchmarked shape (the gate is one comparison per descent).

---
_Generated by [Claude Code](https://claude.ai/code/session_01X3JyUn7YUxZ294nxbu8Jw8)_

### DIFF
--- a/src_rust/src/coverage.rs
+++ b/src_rust/src/coverage.rs
@@ -696,8 +696,15 @@ fn base_fills(edges: &[Edge], rings: &[Vec<Vec3>], cap: &Cap, complement: bool) 
     // winding PIP for it was the dominant fixed cost CodSpeed flagged on
     // small polygons (12 PIPs where pre-#103 code paid ~2).  Culled bases
     // keep fill = false; base_node returns None for them regardless.
+    // The cull-skip's exactness argument (an excluded centre is provably
+    // outside the polygon, so fill = false is a valid donor verdict) needs
+    // the boundary contained in the vertex cap, which cap convexity only
+    // gives for sub-hemisphere caps: a minor arc between in-cap vertices can
+    // exit a > 90° cap by up to a quadrant.  Gate the skip accordingly —
+    // oversize caps classify all 12 seeds, the pre-#109 behavior.
+    let cullable = !complement && cap.radius <= std::f64::consts::FRAC_PI_2;
     for b in 0..12u64 {
-        if !complement {
+        if cullable {
             let center = &centers[b as usize];
             let corners = cell_corners(0, b);
             let (cos_cr, sin_cr) = cell_cos_radius(center, &corners);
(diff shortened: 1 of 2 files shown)


espg:
🤖 *from Claude*

**Verdict: the CodSpeed failure (−15.33% `coverage_triangle[4]`, −12.02% `test_coverage_triangle_order4`) is a cross-environment measurement artifact, not a real regression.** No code change is needed; evidence below.

## Why the diff cannot slow this benchmark

The 10°×10° mid-latitude triangle has a vertex cap of ~5° radius and is non-complement, so on this branch the hot path is byte-for-byte the same decision sequence as `main`:

1. The new gate is **hoisted to one comparison per `base_fills` call** (once per polygon, not per seed or per descent step), and `Cap.radius` is a precomputed field — nothing is recomputed:
   ```rust
   let cullable = !complement && cap.radius <= std::f64::consts::FRAC_PI_2;
   ```
   (`src_rust/src/coverage.rs:705`). For this shape `cullable` is `true`, i.e. the pre-existing `!complement` cull behavior, unchanged.
2. The donor `let-else` fallback sits **after** the `known.iter().all(|&k| k)` early return (`coverage.rs:721`), which fires for this triangle (no seed is boundary-ambiguous) — the changed code is never reached.

Both flagged benchmarks drive the same Rust `polygon_to_morton_coverage`, so a shared runner-environment shift moves them together — exactly what CodSpeed's own ":warning: Different runtime environments detected" banner on the [report](https://github.com/espg/mortie/pull/110#issuecomment-4933215806) says happened (base `2f0439d` and head `35b34df` were measured in different runtime environments).

## Measurement protocol and numbers

Wall-clock on this (shared) box was too noisy to resolve anything — 20 tightly interleaved criterion pairs of `coverage_triangle/4` gave per-pair branch/main ratios spanning 0.53×–2.15×, median ≈ 0.96 (the baseline itself drifted 55→112 µs across runs). So I used **instruction counting, the same methodology as CodSpeed's Simulation mode**: a fixed 1000-iteration harness calling `polygon_to_morton_coverage` on the bench triangle at order 4, compiled identically (`--release`, same toolchain 1.94.1) from each source state, run under `valgrind --tool=cachegrind`:

| run | `main` (2f0439d) | branch (35b34df) |
| --- | --- | --- |
| 1 | 241,097,209 I-refs | 240,036,047 I-refs |
| 2 | 239,887,183 I-refs | 239,818,236 I-refs |

The branch executes **marginally fewer instructions** (−0.1% to −0.4%, within run-to-run jitter). There is no 15% — or any — instruction-count regression to fix; the PR body's "perf-neutral" claim stands, now with direct measurement behind it.

## Current CI state and next step

Every check except CodSpeed is green (tests 3.10/3.11/3.12, Rust + Python benchmark jobs, ruff, codecov patch/project, wheels). Clearing the red CodSpeed check requires either **acknowledging the flagged results on the CodSpeed dashboard** ([link](https://app.codspeed.io/espg/mortie/branches/claude%2F109-review-followup)) or a **re-run of the base measurement in a matching environment** — both are actions on external state I'm leaving for a human rather than taking myself. No new commit was pushed, since any commit would only re-roll the same environment dice without changing the code's cost.

---

Issue #108, opened 2026-07-10
Consolidated follow-ups from the June/July sweeps (#73, #88, #97)


espg:
🤖 *from Claude*

Consolidation of the three rolling sweep issues, filed on espg's direction; #73, #88, and #97 are closed in favor of this issue. Every item from all three was re-verified against current `main` (post PR #106 merge). Resolved items are listed once for the record; everything still open is grouped by what it needs next, with a priority tag per item.

## Resolved since the sweeps (no action — for the record)

- `on_minor_arc` FP sensitivity → implemented in PR #87, then superseded by the uniform predicate in PR #106; #78 carries a close recommendation.
- CI versioning automation → PRs #91/#95 (tag-driven changelog/version carryover; #92 closed).
- `mort2norm` FORCE_PYTHON remnant + stale FORCE_PYTHON benchmark scripts → removed.
- `from_latlon` point-encode (`points=True`) → landed via issue #96 (phases 1–5 on `main`).
- Arrow → pandas `__from_arrow__` hook → wired (`mortie/morton_index.py`).
- Pre-emptive flat-cover size guard → `moc_to_order` guards up front (issue #80 / PR #85).
- `smallvec` for `Node.relevant` → signed off and in (`src_rust/src/coverage.rs`).
- arro3-core CI leg for the pyarrow-free consumer → `arro3-no-pyarrow` job added (issue #98) and green on every PR since.
- Morton `__repr__`, `split_base_cells` ordering + wrapper decisions → resolved per #97's re-verification.
- PR #91 `mktemp+mv` vs `sed -i` portability sign-off → moot (shipped deliberately; noting closes it).

## Group A — small-fixes, bundleable into ONE `claude/small-fixes-<date>` PR

*(each has a standing recommendation; a 👍 on this issue + `small-fix` labels activate the bundle)*

- **[low]** `from_geometry(..., normalize=False)` is silently a no-op for LineString/MultiLineString — raise `ValueError` in the linear branch, symmetric with how `moc`/`tolerance`/`max_cells` are rejected there (`mortie/geometry.py` ~262) + test. *(from #97 item 7)*
- **[low]** Clippy `useless_conversion` false positives (31 at the #97 count, ~38 now) — add `#[allow(clippy::useless_conversion)]` at the crate root with a one-line comment that the pyo3/numpy `?` conversion is load-bearing. Has bounced through #57 → #73 → #88 → #97 undecided; this is the recommended (a). *(from #73 item 8)*
[...]

Issue #97, opened 2026-07-04
Follow up items for the week of July 4, 2026


espg:
🤖 *from Claude*

Sweep of closed PRs from the last 10 days (June 24 – July 4, 2026): #59, #70, #72, #76, #84, #85, #86, #87, #89, #91, #94, #95. Items from #59/#70/#72/#76/#84 were already surfaced by the prior sweep (#88); this issue covers new gaps from #85/#86/#87/#89/#91/#94/#95, plus a re-verification of #88's 10 carry-forward items against current `main`. Six of #88's items are now resolved and excluded: Morton `__repr__` (#76), `split_base_cells` ordering + Python-wrapper decision (#84), `mort2norm` FORCE_PYTHON remnant (#49), CI versioning automation (#91/#95, tracked to completion via now-closed #92), and the stale benchmark scripts (removed entirely). Four carry-forwards are still open and are folded into the numbered list below. Issue #90 (`on_minor_arc`/`node_straddles`, from #87) and #96 (`points=` geo2mort, filed today by espg) are separately tracked and excluded here.

### Items

**(1) `res2display()` order cap still wrong above order 20 — carry-forward from #88** — PR #76

`mortie/tools.py::res2display()` still loops `for res in range(20)`, so it silently returns nothing for any order above 19 even though the packed-u64 kernel supports order 29. Unchanged by any of the 7 new PRs.

**(2) arro3-core CI leg for the pyarrow-free consumer was designed but never wired up** — PR #94 (Refs #92/#93)

PR #94 built the Arrow C Data Interface specifically so zagg's pyarrow-free Lambda worker can consume `morton_index` (`TestArro3Interop`, currently `skipif`-gated and local-only — arro3-core is deliberately kept out of the `test` extra). A concrete `arro3-no-pyarrow` CI job was proposed with exact YAML in a comment on issue #92 ([#92#issuecomment-4859163096](https://github.com/espg/mortie/issues/92#issuecomment-4859163096)), explicitly because the routine lacks the `workflow` OAuth scope needed to push `.github/workflows/` changes itself. Issue #92 was subsequently closed by PR #95 — but #95 only fixed the changelog/version-carryover and `codecov.yml`; the CI leg was never added. The one real pyarrow-absent consumer this feature exists for currently has zero CI coverage.

**(3) `//` floor-division Morton coarsening operator still not implemented — carry-forward from #88** — PR #72

[...]

PR #111, opened 2026-07-10
Group A small-fix bundle from issue #108

### LINKED ISSUE #140: Enforce numpydoc mechanically with numpydoc validation (PR01/RT01/SS01)
**Blocked by #135.**

Split out of #133 → #135 on espg's direction. The `numpydoc` dependency is **pre-approved** ([#135 comment](https://github.com/espg/mortie/issues/135#issuecomment-5080824438)); this issue is only about wiring it, and it must not land until the conversion sweep is finished.

## Why ruff is not enough

CLAUDE.md §7 now mandates numpydoc, and `main` enforces part of it: ruff selects `"D"` with `[tool.ruff.lint.pydocstyle] convention = "numpy"`, with `"mortie/tests/*" = ["D"]` in `per-file-ignores`.

That covers docstring **structure and presence** only. Ruff never requires a `Parameters` section to exist, and never compares a docstring against its signature. Demonstrated — this function reports **nothing** under the current config:

```python
def f(a, b):
    """Do a thing.

    Parameters
    ----------
    a : int
        First.
    """
    return a + b
```

`b` is undocumented and there is no `Returns`; ruff is satisfied. What it does catch is `D400`/`D403` (first-line format), `D202`/`D205`/`D209` (blank lines), and `D1xx` (docstring missing entirely).

The divergence is stark in practice. Measured on `main` before the sweep:

| module | ruff `D` violations | numpydoc `Parameters` blocks / defs |
|---|---|---|
| `mortie/morton_index.py` | **8** | **1 / 59** |
| `mortie/geometry.py` | 7 | 2 / 32 |
| `mortie/arrow.py` | 2 | 0 / 17 |
| `mortie/tools.py` | 24 | 18 / 23 |
| `mortie/coverage.py` | 18 | 11 / 16 |

`morton_index.py` has nearly the cleanest ruff score in the repo and nearly the least conversion, because its docstrings are well-formed prose without sections. **So ruff goes green long before the standard is actually met, and cannot tell you when it is.** That gap is what this issue closes.

## Scope

1. Add **`numpydoc`** to the lint/dev dependency group — approved, docs/lint tooling only, no runtime footprint, no effect on the wheel. numpy stays the only runtime dependency.
2. Configure `[tool.numpydoc_validation]` in `pyproject.toml` with a **curated** check set. The rules worth having are the ones ruff cannot do:
   - `PR01` parameter not documented, `PR02` documented parameter does not exist, `PR03` wrong order, `PR04`/`PR09` type and description hygiene
   - `RT01` no `Returns` section
   - `SS01` no summary
   Deliberately **not** the noisy completeness rules — `SA01` (requires a *See Also* on everything), `EX01` (requires an *Examples* on everything), `ES01` (requires an extended summary). Those would force filler onto trivial functions.
3. Exclude tests, matching the ruff `per-file-ignores` already on `main`.
4. Wire it into CI. Two routes worth weighing: a step in `lint.yml`, or the `numpydoc` pre-commit hook. **Note the routine cannot push `.github/workflows/*`** — that is blocked by OAuth `workflow` scope — so if it goes in `lint.yml` the maintainer applies that hunk by hand, as with #133 phase 3.
5. Decide whether it blocks. `lint.yml` currently runs with `fail_level: none`, so the ruff bot comments but never fails a build. Should numpydoc validation be advisory the same way, or a hard gate? Advisory is consistent with today; a hard gate is what "standard" usually implies. Worth an explicit answer rather than inheriting whichever falls out.

## Why this is blocked, not just sequenced

Turning validation on before the sweep completes would flag essentially every public function at once — a wall of failures nobody reads, which teaches people to ignore the tool. Landing it *after* means it goes green as the sweep finishes and stays green, which is the point of an enforcement gate.

Note #135's first phase covers `morton_index.py`, `geometry.py` and `arrow.py` only; `tools.py`, `coverage.py`, `prefix_trie.py` and `linestring.py` are deferred there to avoid colliding with #136 and PR #130, which are editing them concurrently. **All of those need converting before this can land** — so the block is on #135 as a whole, not on its first PR.

## Acceptance criteria

- `numpydoc` validation runs in CI over `mortie/` excluding tests, with the curated check set.
- It reports **zero** findings on the converted tree — if it does not, the sweep is not finished and this issue is still blocked.
- CLAUDE.md §7 updated to name the enforcement mechanism alongside the standard it already mandates.
- No new runtime dependency.

Refs #135, #133, #68.

espg:
Refs #108 (Group A only — Group B items remain open on that issue, so no `Closes`). Closes #140 (layered into this bundle on espg's direction to reduce churn).

One bundled `small-fix` PR per §5, one commit per item, plus one follow-up commit addressing the adversarial self-review ([review](https://github.com/espg/mortie/pull/111#pullrequestreview-4671113093)):

- [x] **(1) `from_geometry(..., normalize=False)` linestring guard** — the linear branch of `from_geometry` now raises `ValueError("normalize applies only to polygonal geometry")` instead of silently ignoring `normalize=False`, symmetric with the existing `moc`/`tolerance`/`max_cells` rejection two lines above (`mortie/geometry.py`). Docstring updated; test `test_ingest_linear_rejects_normalize_false` covers LineString and MultiLineString via `from_wkt`. *(#108 Group A item 1, from #97 item 7)*
- [x] **(2) crate-root clippy allow for `useless_conversion`** — `#[allow(clippy::useless_conversion)]` at the top of `src_rust/src/lib.rs` with a one-line comment that the pyo3/numpy `?` conversions are load-bearing. `cargo clippy --all-targets` warning count: **43 → 6** (exactly the 37 `useless_conversion` instances suppressed). The 6 remaining warnings (3 `cloned_ref_to_slice_refs`, 2 `useless_vec`, 1 `needless_range_loop` — all pre-existing, mostly in test code) stay visible, confirming nothing else is masked. *(#108 Group A item 2, from #73 item 8)*
- [x] **(3) dissolve hemisphere winding guard** — `dissolve` (Rust `src_rust/src/dissolve.rs`, surfaced as a Python `ValueError` from `rust_dissolve`, mirrored in the Python oracle `_dissolved_rings_py` via `_reject_hemisphere_cover`) now fails loud on covers spanning near or over a hemisphere instead of silently swapping exteriors and holes. Tests: Rust `hemisphere_cover_fails_loud` / `sub_hemisphere_cover_still_dissolves`, Python `test_dissolve_hemisphere_cover_fails_loud` (exact-hemisphere and over-hemisphere polar cap, both engines, plus the `dissolve=False` fallback). *(#108 Group A item 3, from #97 item 5)*
- [x] **(4) numpydoc validation as a hard lint gate** *(issue #140, layered in post-merge with `main`)* — `[tool.numpydoc_validation]` in `pyproject.toml` with the issue's curated check set (PR01/PR02/PR03/PR04/PR09, RT01, SS01; the filler-forcing SA01/EX01/ES01 deliberately off), excluding tests, private members, dunders and private modules per CLAUDE.md §7's public-surface mandate. `numpydoc>=1.10` added to the `test` extra (lint tooling only; numpy stays the sole runtime dep). New `numpydoc` job in `lint.yml` running `numpydoc lint mortie/*.py` as a **hard gate** (espg's explicit call on the issue's question 5) — pushed directly since this session runs with local git credentials, closing the OAuth `workflow`-scope gap the issue flagged. CLAUDE.md §7 names the mechanism. Eight residual findings from the #135 sweep fixed in the same commit (three missing `Returns` on public `MortonIndexArray` members, four `PR09` trailing periods); the tree validates at **zero findings**, meeting the acceptance criterion. Commit `238fbf5`.
- [x] **(review follow-up)** commit `4f43894` addresses the self-review findings:
  - **Σ-vs-exact-area wrap cross-check** ([finding](https://github.com/espg/mortie/pull/111#discussion_r3558639668)): the covered-area guard alone did not make the winding sign trustworthy — the mod-4π fan wrap is driven by the area each *ring* encloses, so the reviewer's ~1.3 sr equatorial band (320 order-4 cells, verified reproducing on both engines) wrapped both rings to −5.63, fired the global flip, and died in the antimeridian stitcher's cryptic assert. `classify_and_split` now takes the exact covered area and rejects with a curated `ValueError` when `||Σ ring areas| − area| > π` (an honest fan matches the exact area within chord discretization ≲0.1 sr; any wrap is off by ~4π, so π has no false-positive surface). Mirrored in the Python oracle (same failure mode there, `RuntimeError` in `_next_segment`); the "sign is trustworthy" comments are now backed by the cross-check. Tests: Rust `equatorial_band_fails_loud_not_cryptic`, Python `test_dissolve_hemisphere_enclosing_ring_fails_loud` (both engines).
  - **Truncated dangling comment** above `classify_and_split` ([finding](https://github.com/espg/mortie/pull/111#discussion_r3558640673)): removed.
  - **`cover_area` disjoint-cells precondition** ([finding](https://github.com/espg/mortie/pull/111#discussion_r3558641915)): documented in both mirrors (`cover_area` doc comment and `_reject_hemisphere_cover` docstring) rather than deduping — duplicate words already break the edge-cancellation dissolve itself, so disjoint input is the function's existing contract.

### Note on the guard key for (3)

The issue's standing recommendation was to raise when `|Σ ring areas| ≈ 2π`. While implementing it I found the literal check does not fire on the flagship ambiguous input: for a cover of exactly half the sphere (24 order-1 cells, base cells 0–5), the anchor-fan signed-area formula wraps mod 4π per-ring and returns `Σ = −1.85`, not `±2π` — and because ring chaining starts from a `HashMap` iteration, the wrapped value's *sign* is not even deterministic at that scale. Probing polar caps: `Σ` tracks the true area faithfully below ~2π (e.g. 6.02 for a 6.02 sr cap) and only breaks at/past the hemisphere.

So the guard is keyed on the **exact covered area** instead — cells are equal-area, so `Σ π/(3·4^depth)` over the input words is exact — raising when `area > 2π − margin` with `margin = 2% of 2π` (`HEMISPHERE_MARGIN`). This honors the item's intent (fail loud when the winding sign becomes untrustworthy), fires on the exact-hemisphere case the literal check misses, and additionally catches all over-hemisphere covers (which today silently emit the complement). The review follow-up adds the complementary per-ring wrap detection (Σ cross-checked against the exact area), so hemisphere-enclosing *rings* on small covers now also fail loud instead of panicking. The error messages tell the user to split the cover or pass `dissolve=False`.

### How tested

- Post-merge with `main` at PR #146 (2026-07-26): `cargo test --lib` **256 passed, 0 failed, 1 ignored**; `cargo fmt --check` clean; `cargo clippy --lib --tests`: **6 pre-existing warning sites, 0 `useless_conversion`** (the crate-root allow retires the whole class, including the two instances PR #146 had to carry as declared deviations).
- `maturin develop --release && pytest -q`: **874 passed, 12 skipped** (this PR's 3 new Python tests included).
- `flake8 mortie --select=E9,F63,F7,F82`: clean.

### Questions for review

1. **Threshold for (3):** the 2% margin (`2π·0.02 ≈ 0.126 sr`) is a judgment call — it keeps the guard clear of the fan-formula breakdown while still admitting a 96%-of-hemisphere polar cap (which dissolves correctly today). The wrap cross-check tolerance (π) is separately justified above. Happy to tighten/loosen either.
2. **Guard key for (3):** keyed on exact covered area rather than the literal `|Σ ring areas|` for the reasons above — flagging the deviation from the issue's wording explicitly in case the literal form was wanted anyway.
3. Over-hemisphere covers (area > 2π) now raise where they previously returned a silently inverted outline, and hemisphere-enclosing-ring covers (e.g. equatorial bands) now raise a curated `ValueError` where they previously surfaced a cryptic stitcher panic — behavior changes beyond "near 2π", but the old outputs were wrong or unactionable. The full winding-free classifier from #97 item 5 remains a possible follow-up if polar-scale dissolves become a real workload — and it is now mostly on the shelf: since PRs #138/#146 the coverage layer trusts `winding_sign_in_cap` / `ring_turning_sign` (Gauss–Bonnet sign with a derived rounding bound) with `ring_is_simple` / `ring_set_validity` as the checkable precondition, so a dissolve classifier would reuse those instruments rather than #97's sketch.
4. **Uniformity call — answered by events.** The asymmetry this question described no longer exists: PR #138 phase 4 threaded `normalize` through every MOC entry point and PR #138's phase-4 fold threaded `from_geometry`'s `moc=True` branch, so the polygonal path honors `normalize=False` uniformly under decision (A) (issue #144). The merge with `main` resolves this PR's docstring accordingly (the linear-branch `ValueError` from item (1) is retained; the "ignored when `moc=True`" text is gone).


src_rust/src/dissolve.rs
```diff
@@ -41,10 +41,40 @@ pub struct ClassifiedRings {
...
+pub fn dissolve(morton: &[u64], step: u32) -> Result<ClassifiedRings, String> {
+    let area = cover_area(morton);
+    if area > std::f64::consts::TAU - HEMISPHERE_MARGIN {
+        return Err(format!(
+            "dissolved cover spans {area:.6} sr — within 2% of a hemisphere \
+             (2π sr) or beyond — so its exterior/hole winding is ambiguous; \
+             split the cover into sub-hemisphere parts or pass dissolve=False \
+             for per-cell polygons"
+        ));
+    }
```
espg, line 78: 🤖 *from Claude (review)*

**The exact-area key is sound for what it checks, but it does not deliver what the new comments claim.** The mod-4π wrap is a function of the area *each ring encloses*, not of the *covered* area — so a small cover whose rings individually enclose more than a hemisphere still wraps, sails through this guard, and still breaks classification.

Verified end-to-end on this branch (throwaway clone, scratch test in this module): an equatorial band of 320 order-4 cells covering |lat| ≲ 3.7°:

```
band cells=320 area=1.3089969389957445 (guard threshold=6.157521601035994)
boundary rings=2 ring_areas=[-5.628269441174673, -5.628269441174666] sum=-11.25653888234934
thread ... panicked at src_rust/src/dissolve.rs:372:5:
unbalanced antimeridian segments but no pole enclosed
```

Each boundary ring encloses a ~2π+0.65 sr cap, so the fan returns the wrapped complement (−5.628 instead of +6.94), `Σ = A − 4π < 0`, the global flip fires on a correctly-wound cover, and the stitcher dies on its assert — surfaced to Python as the cryptic `dissolve panicked: unbalanced antimeridian segments but no pole enclosed` instead of this item's curated, actionable `ValueError`. (This family is pre-existing on `main` — `classify_and_split` is unchanged — so it's not a regression; but the PR's whole point per issue #108 was "fail loud instead of silently inverting", and this polar-adjacent shape still fails cryptic.)

Cheap hardening that would close it: `dissolve` already has the exact `area` in hand — pass it into `classify_and_split` and, after computing the fan areas, reject when `(total.abs() - area).abs()` is large (any wrapped ring is off by ~4π; a tolerance of, say, π has no false-positive surface). That converts every wrap — near-2π covers *and* hemisphere-enclosing-ring covers — into the same actionable error, and makes the "sign is trustworthy" comments below actually true. At minimum, the comments at lines 432–433 (and the `geometry.py` mirror) should be softened back to an assumption, because as written they overclaim.

---

src_rust/src/dissolve.rs
```diff
@@ -387,6 +417,8 @@ fn ring_signed_area_lonlat(ring: &[(f64, f64)]) -> f64 {
     spherical_signed_area(&v)
 }
 
+// Winding-guard margin (issue #108): spherical signed area is defined mod 4π,
+// so a cover whose |Σ ring areas| lands near 2π reads the same as its
```
espg: 🤖 *from Claude (review)*

Leftover edit artifact: this comment is truncated mid-sentence ("…reads the same as its") and duplicates the `HEMISPHERE_MARGIN` block at the top of the file — it also dangles over `classify_and_split`, which defines no "winding-guard margin". Delete these two lines.

---

src_rust/src/dissolve.rs
```diff
@@ -41,10 +41,40 @@ pub struct ClassifiedRings {
...
+// covers within 2% of half the sphere.
+const HEMISPHERE_MARGIN: f64 = std::f64::consts::TAU * 0.02;
+
+/// Exact covered area (steradians) of a morton cover: Σ π/(3·4^depth).
+fn cover_area(morton: &[u64]) -> f64 {
+    morton
+        .iter()
+        .map(|&w| std::f64::consts::PI / (3.0 * 4f64.powi(mort2nested(w).1 as i32)))
+        .sum()
+}
```
espg, line 61: 🤖 *from Claude (review)*

Nit: "exact" holds only for disjoint, non-duplicated words — duplicates and parent/child overlaps double-count, so a legitimate sub-threshold cover containing duplicates can spuriously raise. That precondition is real in practice: raw `morton_coverage` output can carry duplicates (the test helper `_polar_cap` that feeds the new `over` test needs its `np.unique`). Worth one line in this doc comment (mirrored in `_reject_hemisphere_cover`) stating the disjoint-cells precondition, or a dedup in both engines — whichever, keep the two mirrors in lockstep.

---

mortie/geometry.py
```diff
@@ -663,8 +690,8 @@ def _dissolved_rings_py(morton, step):
     # holes) is the covered area, always positive.  HEALPix orders boundary
     # points one way for step==1 and the other for step>1, so key the
     # exterior/hole sign off this invariant rather than a fixed convention.
-    # (Spherical signed area is defined mod 4π, so this assumes the cover stays
-    # well under a hemisphere — true for every realistic emit input.)
+    # (Spherical signed area is defined mod 4π; the hemisphere guard above
+    # keeps the cover well under 2π, so the sign is trustworthy.)
```

Same overclaim as the Rust twin (see the main comment on `dissolve` in `src_rust/src/dissolve.rs`): the guard bounds the *covered* area, but the wrap is driven by the area each *ring encloses* — an equatorial band of ~1.3 sr (well under 2π) still wraps both rings to −5.63, flips the winding here, and dies in `_next_segment`'s `RuntimeError` instead of the new curated `ValueError`. "keeps the cover well under 2π, so the sign is trustworthy" should either become true (cross-check `sum(areas)` against the exact covered area the guard just computed) or stay phrased as the assumption it was before this change.

---

mortie/geometry.py
```diff
@@ -260,6 +261,8 @@ def from_geometry(geom, order=18, moc=False, normalize=True,
         raise ValueError(
             "moc / tolerance / max_cells apply only to polygonal geometry"
         )
+    if not normalize:
+        raise ValueError("normalize applies only to polygonal geometry")
```
espg, line 404: 🤖 *from Claude (review)*

Nit / question for espg: the fix matches issue #108 Group A item 1 exactly and the test pins both LineString and MultiLineString. But the same "silent no-op" rationale from #97 item 7 now applies one branch up: `moc=True, normalize=False` on *polygonal* geometry is still silently ignored (`morton_coverage_moc` has no orientation auto-correct at all), while the linear branch fails loud. The updated docstring enshrines the asymmetry ("Ignored when ``moc=True``"). Defensible as-is since the issue scoped this to the linear branch — flagging in case fail-loud was meant to be the uniform policy.

---

espg (commented):
🤖 *from Claude (review)*

Solid bundle overall: fix (1) matches the issue and is properly tested; fix (2)'s inventory claim verified empirically (clippy on `main`: exactly 37 `useless_conversion`; on this branch: 0 suppressed lints, precisely the 6 claimed pre-existing warnings remain); and fix (3)'s deviation from the issue's literal `|Σ ring areas|` trigger is well-reasoned — the fan sum does wrap mod 4π exactly where the literal check would need it, and keying on exact covered area is the right call for that failure mode. Findings, ranked: **(should-fix)** the guard does not make the winding sign trustworthy as the new comments claim — a ~1.3 sr equatorial band (verified end-to-end on this branch) wraps both ring areas to −5.63, fires the global flip, and dies in a cryptic stitcher panic instead of the new actionable ValueError; a cheap cross-check of Σ fan areas against the exact covered area would close the hole (details inline on `dissolve`). **(should-fix, trivial)** a truncated leftover comment dangles above `classify_and_split` (dissolve.rs 420–421). **(nit)** `cover_area`'s disjoint-cells precondition is undocumented (duplicates double-count toward the threshold); **(nit)** `moc=True, normalize=False` on polygonal geometry remains a silent no-op while the linear branch now fails loud — worth an explicit espg call on uniformity.

---

espg:
🤖 *from Claude*

Audit of this bundle against the geometry work that has landed since it was written (PR #138 — the reference-layer repair, decision (A) on issue #144, and the id-threading — plus PR #146's ring-validity machinery), as requested by espg. Item by item: what still stands, what is stale, and where the ground moved under it.

## Still fully relevant

- **Item (1), the linear-branch `normalize` guard** — still correct and still wanted: `normalize` remains a polygonal-only concept and the loud rejection is right. Two caveats below (docstring + Q4).
- **Item (2), the crate-root `useless_conversion` allow** — *more* relevant than when written: the count has grown with the crate (now ~35 instances; PR #146's phase 3 added one more of the identical idiom and had to flag it as a declared deviation from the clippy baseline). Merging this item retires that whole nuisance class. The "43 → 6" numbers are stale but the direction is unchanged.
- **Item (3)'s intent** — fail loud where the dissolve winding sign is untrustworthy — still stands, and the exact-covered-area key remains sound (cells are equal-area; that reasoning is untouched by anything we changed).

## Stale or superseded

1. **Question (4)'s asymmetry is gone.** "`moc=True, normalize=False` on polygonal geometry is a silent no-op (`morton_coverage_moc` has no orientation auto-correct at all)" — no longer true: PR #138 phase 4 threaded `normalize` through every MOC entry point (Rust core, `_rustie` wrappers, `morton_coverage_moc`), and the phase-4 fold threaded `from_geometry`'s `moc=True` branch specifically ([the finding](https://github.com/espg/mortie/pull/138#discussion_r3652007041)). The polygonal branch now honors `normalize=False` uniformly, so the docstring text this item adds about the asymmetry would document a state that no longer exists — it needs a rebase-time rewrite, not just a merge.
2. **Item (3)'s rationale section describes the pre-repair winding world.** The anchor-fan signed-area formula whose mod-4π wrap drove the guard's design is the same family of machinery #138 replaced at the coverage layer with the Gauss–Bonnet turning sign plus a derived rounding bound (`ring_turning_with_error`), and #146 adds `ring_is_simple` / `ring_set_validity`. The *guard itself* stays valid — dissolve's internal fan formula is untouched by our changes, so its failure modes are real — but the closing note ("the full winding-free classifier from #97 item 5 remains a possible follow-up") is now mostly *available on the shelf*: a dissolve classifier keyed on `winding_sign_in_cap` / `ring_turning_sign` with the validity check as its precondition would replace the wrap cross-check with the same instrument the coverage layer now trusts. Worth a line in the PR body so the follow-up path points at the new machinery rather than at #97's sketch.
3. **The "How tested" numbers are two eras old** (200 Rust / 673 Python; current baselines are 252 / 871 on the #146 branch) — expected for a `waiting` PR, noted only so the rebase refreshes them.
4. **Merge-conflict surface**: `mortie/geometry.py` (the phase-4 fold edited the same `from_geometry` region this PR touches) and `src_rust/src/lib.rs` (many additions since) will both conflict textually; `dissolve.rs` itself is untouched by the geometry work and should rebase clean.

## Net recommendation

The bundle survives: rebase it, rewrite the Q4/docstring text for the now-uniform `normalize` behavior, refresh the counts, and add one sentence pointing item (3)'s future-work note at `ring_turning_sign`/`ring_is_simple`. Nothing in it is invalidated; one of its open questions has been answered by events (Q4 — uniformity arrived from the other direction).

---

espg:
🤖 *from Claude*

**Refreshed for merge on top of PR #146, per espg's go-ahead** — `main` merged into this branch as `8a9d00e` (merge commit; no force-push, no rebase).

**The one conflict was exactly the audit's prediction**: `from_geometry`'s `normalize` docstring, where this PR's pre-decision-(A) text ("ignored when `moc=True`; `morton_coverage_moc` has no orientation auto-correct") collided with the uniform behavior that landed since. Resolved to the current semantics — `normalize` honored on both polygonal paths under decision (A), with this PR's own feature retained verbatim: `normalize=False` on linear geometry raises `ValueError`. Everything else auto-merged (`lib.rs`, `test_geometry.py`); `dissolve.rs` merged clean as predicted.

**Body updated in place**: "How tested" refreshed to post-merge numbers, question (4) marked answered-by-events, and item (3)'s follow-up note now points at the #138/#146 machinery (`ring_turning_sign` / `ring_is_simple`) instead of #97's sketch.

**Suites on the merged branch**: `cargo test --lib` **256 / 0 / 1 ignored** · `pytest` **874 passed, 12 skipped** (this PR's three new tests included) · `cargo fmt --check` + both flake8 gates clean · `cargo clippy --lib --tests` **6 pre-existing warning sites, zero `useless_conversion`** — item (2)'s crate-root allow retires the whole class, including the two instances PR #146 carried as declared deviations, exactly as hoped.

One honesty note: the first post-merge `cargo test --lib` run showed a single failure (255+1) that did not reproduce across four consecutive clean re-runs (256/0 each); the failing test's name was lost to output filtering before it could be recorded. CI's three legs on this push are the arbiter — if any leg disagrees, that flake is real and I'll chase it.

Ready for review/merge; `waiting` stands.

---

espg:
🤖 *from Claude*

**Issue #140 layered into this bundle as item (4), per espg — commit `238fbf5`.**

- `[tool.numpydoc_validation]` configured with the issue's curated set (signature-agreement + section-presence checks; the filler-forcing completeness rules deliberately off). Exclusions: tests (mirroring ruff), plus private members, dunders and private modules — CLAUDE.md §7 mandates the standard for the *public* surface, which is also why most of the 30 raw findings were curated out rather than padded with boilerplate.
- The eight findings that were genuinely public got fixed: `Returns` sections on `MortonIndexArray.dtype`/`nbytes`/`copy`, and four `PR09` trailing periods in `tools.py`. `numpydoc lint mortie/*.py` now reports **zero findings** — the issue's acceptance criterion.
- New `numpydoc` job in `lint.yml` as a **hard gate** (espg's explicit answer to the issue's question 5; ruff's advisory mode is untouched). Pushed directly from the local session, closing the OAuth `workflow`-scope gap the issue flagged for routine runs.
- `numpydoc>=1.10` in the `test` extra — lint tooling only, numpy remains the sole runtime dependency. CLAUDE.md §7 now names the enforcement mechanism beside the standard.
- Blocker check: #135 is closed and the tree validated clean after the eight residual fixes, so the "still blocked" condition in the acceptance criteria is satisfied rather than bypassed.

Suites: `pytest` **874 passed, 12 skipped** · `numpydoc lint` zero findings · flake8 gates clean · no Rust changes. The new lint job will run on this very push — its first green is the gate proving itself.

---

PR #112, opened 2026-07-10
Cause-tagged node_straddles instrumentation + over-refinement measurement (issue #90, phases 1-2)

espg:
Refs #90

Phases 1 and 2 of the measure-then-relax program agreed in the [issue plan](https://github.com/espg/mortie/issues/90#issuecomment-4932919988) and the [maintainer's answers](https://github.com/espg/mortie/issues/90#issuecomment-4932984344): a cargo-feature instrumentation (usable from Python dev builds), both KPIs (output cell count flat + MOC, and wall time, MOC size the priority), go threshold >=10% MOC-size reduction or comparable wall-time percentage reduction on at least one realistic polygon class. **Phase 3 (the actual relaxation) is explicitly a separate future PR** and is not part of this one.

### Phase 1 — cause-tagged instrumentation (`descent-stats` cargo feature)

- `Cargo.toml` declares a `descent-stats` feature — a cfg flag only, **no new dependency**. Off by default; release builds compile none of the instrumentation (all sites are `#[cfg(feature = "descent-stats")]`).
- `src_rust/src/coverage/descent_stats.rs` (new, feature-gated): the issue #90 cause taxonomy — `VertexLeaf`, `QuadCross`, `QuadTouch` (the closed-set exact-incidence branch, contract per #103, excluded from over-refinement), `CornerParity`, `NearPoleBulge` (the densified #32 path) — plus a global collector (stdlib `Mutex`; counters and per-leaf records updated under one lock so `sum(leaf) == leaves.len()` holds under rayon interleaving) and a thread-local cause/touch hand-off from `node_straddles` to its callers.
- `src_rust/src/coverage.rs`: each `node_straddles` true-return path tags its cause; `edge_hits_cell_edge` / `edge_touches_cell_edge_degenerate` tag whether the deciding hit was the exact-incidence touch branch or a crossing; both descent drivers (`descend_parallel`, `consider_node`) record straddle-stopped leaves (morton, depth, cause, centre fill, centre vector, densified-boundary circumradius) vs internal refinements. The only feature-off structural change is splitting the quad clause's `A || B` into two sequential `if`s — semantically identical short-circuit.
- `src_rust/src/lib.rs`: one feature-gated binding, `rust_descent_stats_take()` (take-and-reset), returning the counters and the per-leaf table as numpy arrays.
- Tests: `src_rust/tests/descent_stats.rs` (own integration binary via `[[test]] required-features`, so the process-global collector is isolated from the lib's other descent tests) + `mortie/tests/test_descent_stats.py` (skips unless the extension was built with the feature).

### Phase 2 — the measurement

`benchmarks/measure_straddle_overrefinement.py` — a dev script, **not wired to CI**; run and reported, and may be dropped before merge (the #78 benchmark precedent). For each straddle-stopped leaf it independently re-tests whether the polygon boundary actually intersects the cell: dense great-circle sampling of every edge (spacing 0.2 cells) classified with the exact HEALPix point-in-cell assignment (`geo2mort`), plus a distance proof — a leaf is **provably missed** only when the min angular distance from its centre to every sample exceeds its densified circumradius + half the sample spacing + 1e-6 rad; anything closer that no sample lands in is **ambiguous** and never claimed. `quad_touch` leaves are excluded from the over-refinement count by the #103 contract regardless of the geometric verdict. Achievable reductions are computed exactly: drop provably-missed fill=false leaves via `moc_minus` and renormalize (MOC size), `moc_to_order_count` (flat size).

Suite: the #103 on-grid boxes (belt/lon 0, cap/lon 45, equator), the bench polygons (circles 32/100/500, mid-lat + polar triangles/squares), a synthetic ICESat-2-style swath (~130 deg near-polar track, ~6.6 km width), and a hemisphere+ ring (spherical circle, radius 100 deg), at orders {6, 9, 11}. Wall time is measured on the normal (uninstrumented) release build; cause stats on the `descent-stats` build. Results are posted as a table on issue #90.

## Phases

- [x] Phase 1 — feature-gated cause-tagged instrumentation + tests
- [x] Phase 2 — measurement script, run across the suite; results table + go/no-go recommendation posted on issue #90

## How tested

- `cargo test` (feature off: 197 passed) and `cargo test --features descent-stats` (197 + the new integration test), `cargo fmt`, `cargo clippy --all-targets` both with and without the feature (the one new clippy note is the same macro-generated `useless_conversion` that fires on every existing `#[pyfunction]`).
- `maturin develop --release` + full `pytest -v`: 670 passed, 12 skipped (the new Python test skips on the uninstrumented build, runs on the `descent-stats` build).
- Zero-cost check (feature off): timed `morton_coverage_moc` across triangle / polar square / circle100 / circle500 at orders 6/9/11 before and after; cross-build ratios 0.84–1.12 are entirely inside the same-build run-to-run noise floor on this box (up to 1.47x), i.e. no detectable regression — expected, since the feature-off diff is a semantically-identity restructure.

## Questions for review

- The plan sketched "thread-local counters merged per descent"; the implementation uses one stdlib `Mutex` instead (no new dependency, atomic count+record invariant, contention only at straddle events on a dev-only build). Flagging the deviation for visibility.
- `descent_stats` is `pub` (not `pub(crate)`) so the integration test binary can reach it; it only exists under the dev feature, so nothing new ships in release wheels.
- The measurement script is committed on this branch per the "run and reported" precedent — say the word and it is dropped before merge.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fn3kUE3NLLMN1sSEGeswmG)_

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,20 @@ healpix = "0.3.3"
 smallvec = "1"
 arrow = { version = "53", features = ["ffi"] }
 
+[features]
+# Cause-tagged `node_straddles` instrumentation for the issue #90 measurement.
+# A cfg flag only (no dependency); off by default so release builds compile
+# none of it.  Build dev wheels with:
+#   maturin develop --release --features descent-stats
+descent-stats = []
+
 [dev-dependencies]
 criterion = { version = "4.3.0", package = "codspeed-criterion-compat" }
+
+[[test]]
+name = "descent_stats"
+path = "src_rust/tests/descent_stats.rs"
+required-features = ["descent-stats"]
 
 [profile.release]
 opt-level = 3
--- a/benchmarks/measure_straddle_overrefinement.py
+++ b/benchmarks/measure_straddle_overrefinement.py
@@ -210,6 +210,10 @@ def measure_stats(lats, lons, order):
     smort = np.asarray(mortie.geo2mort(slat, slon, order=order), np.uint64)
 
     morton = st["morton"]
+    # The order-level membership test below (`np.isin(morton, smort)`) is only
+    # sound for the exact descent, where every straddle leaf stops at `order`;
+    # a tolerance/budget stop would record coarser leaves and misclassify.
+    assert np.all(st["depth"] == order), "measurement requires exact-mode leaves"
     cause = st["cause"]
     fill = st["fill"]
     centers = np.stack([st["cx"], st["cy"], st["cz"]], axis=1)
@@ -262,11 +266,12 @@ def measure_stats(lats, lons, order):
         "ambiguous": int(ambiguous.sum()),
         "moc_hyp": int(len(moc_hyp)),
         "flat_hyp": int(flat_hyp),
-        "moc_reduction_pct": 100.0 * (1.0 - len(moc_hyp) / len(moc)),
+        "moc_reduction_pct": 100.0 * (1.0 - len(moc_hyp) / max(len(moc), 1)),
         "flat_reduction_pct": 100.0 * (1.0 - flat_hyp / max(flat_cur, 1)),
         "ceiling_removed": int(len(ceil_removed)),
         "moc_ceiling": int(len(moc_ceil)),
-        "moc_ceiling_reduction_pct": 100.0 * (1.0 - len(moc_ceil) / len(moc)),
+        "moc_ceiling_reduction_pct": 100.0
+        * (1.0 - len(moc_ceil) / max(len(moc), 1)),
         "samples": int(len(samples)),
         "sample_spacing_rad": float(spacing),
         "stats_build_wall_s": wall,
--- a/benchmarks/measure_straddle_overrefinement.py
+++ b/benchmarks/measure_straddle_overrefinement.py
@@ -1,0 +1,376 @@
+"""
+Issue #90 phase-2 measurement: how much of the `node_straddles` boundary
+refinement is provable over-refinement, by cause?
+
+Dev script — not wired to CI ("run and reported", the issue #78 benchmark
+precedent).  Protocol (two builds):
+
+    # 1. wall-time baseline on the normal release build (no instrumentation)
+    maturin develop --release
+    python benchmarks/measure_straddle_overrefinement.py timing -o timing.json
+
+    # 2. cause-tagged stats on the instrumented build
+    maturin develop --release --features descent-stats
+    python benchmarks/measure_straddle_overrefinement.py stats -o stats.json
+
+    # 3. merge into the report table
+    python benchmarks/measure_straddle_overrefinement.py report timing.json \
+        stats.json
+
+For each straddle-stopped leaf the descent recorded, the polygon boundary is
+re-tested independently of the descent's own predicates: every polygon edge
+is densely sampled along its great circle (spacing 0.2 cells) and each sample
+is classified with the exact HEALPix point-in-cell assignment (`geo2mort`).
+
+  * a sample lands in the leaf          -> the boundary genuinely enters it;
+  * min angular distance from the leaf centre to every sample exceeds the
+    leaf's densified circumradius + half the sample spacing (+1e-6 rad
+    safety) -> the boundary **provably misses** the cell = over-refinement;
+  * anything between                    -> ambiguous, counted but not claimed.
+
+`quad_touch` leaves (the #103 closed-set exact-incidence branch) are contract,
+not waste: they are excluded from the over-refinement count and from the
+hypothetical reduction regardless of the geometric verdict.
+
+The achievable reduction assumes provably-missed leaves whose centre fill is
+False (fully outside) are dropped; provably-missed fill=True leaves are fully
+inside and already merge optimally in the MOC, so they change descent work
+but not output size.  Hypothetical covers are computed exactly with
+`moc_minus` + renormalization.
+"""
+
+import argparse
+import json
+import sys
+import time
+
+import numpy as np
+
+import mortie
+from mortie import _rustie, moc_minus, morton_coverage_moc
+
+CAUSES = ["vertex_leaf", "quad_cross", "quad_touch", "corner_parity",
+          "near_pole_bulge"]
+QUAD_TOUCH = CAUSES.index("quad_touch")
+ORDERS = [6, 9, 11]
+SAMPLE_FRACTION = 0.2  # sample spacing as a fraction of the cell resolution
+SAFETY_RAD = 1e-6
+
+
+# ---------------------------------------------------------------------------
+# Polygon suite
+# ---------------------------------------------------------------------------
+
+def _unit(lats, lons):
+    la, lo = np.radians(lats), np.radians(lons)
+    return np.stack(
+        [np.cos(la) * np.cos(lo), np.cos(la) * np.sin(lo), np.sin(la)], axis=-1
+    )
+
+
+def _latlon(v):
+    return (
+        np.degrees(np.arcsin(np.clip(v[:, 2], -1, 1))),
+        np.degrees(np.arctan2(v[:, 1], v[:, 0])),
+    )
+
+
+def box(lat_lo, lat_hi, lon_w, lon_e):
+    return (
+        np.array([lat_lo, lat_lo, lat_hi, lat_hi], float),
+        np.array([lon_w, lon_e, lon_e, lon_w], float),
+    )
+
+
+def circle(n, lat0=-75.0, lon0=0.0, radius=5.0):
+    """The bench circle: lat/lon-offset approximation (matches
+    benchmarks/test_bench_coverage.py and the Rust coverage bench)."""
+    ang = np.linspace(0, 2 * np.pi, n, endpoint=False)
+    return lat0 + radius * np.cos(ang), lon0 + radius * np.sin(ang)
+
+
+def icesat2_swath(width_deg=0.06, n_track=260):
+    """Synthetic ICESat-2-style swath: a ~130 deg near-polar ground-track
+    segment buffered to ~6.6 km total width (the beam-pair span)."""
+    a = _unit(np.array([-65.0]), np.array([-40.0]))[0]
+    b = _unit(np.array([65.0]), np.array([20.0]))[0]
+    omega = np.arccos(np.clip(np.dot(a, b), -1, 1))
+    t = np.linspace(0.0, 1.0, n_track)
+    p = (
+        np.sin((1 - t) * omega)[:, None] * a + np.sin(t * omega)[:, None] * b
+    ) / np.sin(omega)
+    p /= np.linalg.norm(p, axis=1, keepdims=True)
+    # local track tangent and cross-track normal
+    tan = np.gradient(p, axis=0)
+    tan /= np.linalg.norm(tan, axis=1, keepdims=True)
+    nrm = np.cross(p, tan)
+    nrm /= np.linalg.norm(nrm, axis=1, keepdims=True)
+    h = np.radians(width_deg / 2.0)
+    left = np.cos(h) * p + np.sin(h) * nrm
+    right = np.cos(h) * p - np.sin(h) * nrm
+    ring = np.concatenate([left, right[::-1]])
+    return _latlon(ring)
+
+
+def hemisphere_ring():
+    """Hemisphere-plus ring: the pinned #22 world ring (lat -80..80,
+    lon -90..90, vertex sum balanced so ingest trusts the winding), whose
+    interior is the lon-0-facing hemisphere+ region including both poles —
+    the shape `test_coverage_hemisphere.test_complement_world_minus_cap`
+    validates.  Verified here to cover > half the sphere."""
+    lats = np.array([-80.0, -80.0, 80.0, 80.0])
+    lons = np.array([-90.0, 90.0, 90.0, -90.0])
+    cells = morton_coverage_moc(lats, lons, order=3)
+    frac = _rustie.rust_moc_to_order_count(
+        np.asarray(cells, np.uint64), 3
+    ) / (12 * 4 ** 3)
+    if frac <= 0.5:
+        raise RuntimeError(f"hemisphere+ ring covers only {frac:.3f}")
+    return lats, lons
+
+
+def shapes():
+    tri_lat, tri_lon = np.array([20.0, 30.0, 25.0]), np.array(
+        [-120.0, -120.0, -110.0]
+    )
+    sq_lat, sq_lon = np.array([20.0, 20.0, 30.0, 30.0]), np.array(
+        [-125.0, -115.0, -115.0, -125.0]
+    )
+    trip_lat, trip_lon = np.array([-80.0, -88.0, -84.0]), np.array(
+        [-120.0, -120.0, -100.0]
+    )
+    sqp_lat, sqp_lon = np.array([-80.0, -80.0, -87.0, -87.0]), np.array(
+        [-130.0, -100.0, -100.0, -130.0]
+    )
+    return [
+        # (class, name, lats, lons)
+        ("ongrid_box", "box_belt_lon0", *box(20.0, 25.0, 0.0, 5.0)),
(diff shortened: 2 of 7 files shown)


espg:
🤖 *from Claude*

Status: both phases pushed; the [measurement report + no-go recommendation](https://github.com/espg/mortie/issues/90#issuecomment-4934919084) is on issue #90.

**Incidental observation (out of scope here, flagging rather than acting):** while building the hemisphere+ suite shape I hit what looks like an under-coverage anomaly for a hemisphere+ ring whose radius crosses 90°. Reproducer:

```python
import numpy as np
from mortie import morton_coverage_moc, moc_to_order

c = np.array([np.cos(np.radians(45)), 0.0, np.sin(np.radians(45))])  # (45N, 0)
e1 = np.cross([0.0, 0.0, 1.0], c); e1 /= np.linalg.norm(e1)
e2 = np.cross(c, e1)
th = np.linspace(0, 2 * np.pi, 96, endpoint=False)
r = np.radians(97.5 + 12.5 * np.sin(3 * th))   # wobbly ring, radius 85..110 deg
v = np.cos(r)[:, None] * c + np.sin(r)[:, None] * (
    np.cos(th)[:, None] * e1 + np.sin(th)[:, None] * e2)
lats = np.degrees(np.arcsin(v[:, 2])); lons = np.degrees(np.arctan2(v[:, 1], v[:, 0]))
# both windings produce the *identical* 45.6%-of-sphere cover at order 4,
# yet the ring's two complementary interiors are ~55.8% and ~44.2%:
for la, lo in ((lats, lons), (lats[::-1].copy(), lons[::-1].copy())):
    cells = morton_coverage_moc(la, lo, order=4)
    print(len(np.asarray(moc_to_order(cells, 4))) / (12 * 4 ** 4))
```

The ring is simple (azimuth strictly monotone around its centre), its vertex sum axis has min-dot −0.087 ≤ 0, so ingest takes the hemisphere+ "trust the winding" path — yet both windings return the same cover, and ~53% of dense-sampled interior points of the large side are missing from it (checked at order 4 with a 0.02 rad boundary margin). Two complementary interiors cannot both have sound covers of 45.6%. For the measurement I switched to the pinned #22 world ring (`test_complement_world_minus_cap`'s outer ring), which behaves correctly. If you want this chased down, it likely deserves its own issue — leaving that call to you.

---

Cargo.toml
```diff
@@ -19,9 +19,21 @@ healpix = "0.3.3"
...
+#   maturin develop --release --features descent-stats
+descent-stats = []
+
 [dev-dependencies]
 criterion = { version = "4.3.0", package = "codspeed-criterion-compat" }
 
+[[test]]
+name = "descent_stats"
+path = "src_rust/tests/descent_stats.rs"
+required-features = ["descent-stats"]
```
espg, line 35: 🤖 *from Claude (review)*

**CI never exercises the `descent-stats` feature.** `test.yml` builds with plain `pip install -e .` and runs plain `cargo test`, so this `required-features` integration test is always filtered out in CI and `mortie/tests/test_descent_stats.py` always skips (its `skipif` gates on the binding's presence). A future true-return path in `node_straddles` that forgets to tag a cause would pass CI green and only trip the `take_cause` panic on the next manual instrumented build. Not fixable in this PR without touching workflows (out of scope per repo conventions) — flagging so the tradeoff is explicit; an alternative is one CI job step `cargo test --features descent-stats` if the dev feature is kept long-term.

---

benchmarks/measure_straddle_overrefinement.py
```diff
@@ -0,0 +1,376 @@
...
+    slat, slon = _latlon(samples)
+    smort = np.asarray(mortie.geo2mort(slat, slon, order=order), np.uint64)
+
+    morton = st["morton"]
+    cause = st["cause"]
+    fill = st["fill"]
+    centers = np.stack([st["cx"], st["cy"], st["cz"]], axis=1)
+    circ = st["circ"]
+
+    hit = np.isin(morton, smort)  # a boundary sample lands in the leaf
```
espg, line 222: 🤖 *from Claude (review)*

**Latent misclassification if the script is ever run in tolerance/budget mode.** `hit = np.isin(morton, smort)` compares leaf morton words against order-level `geo2mort` sample words, which is only sound because the exact descent stops every straddle leaf at `depth == order`. `record_leaf` stores `nested2mort(pixel, depth)` — under a `tolerance=`/`max_cells=` stop a coarser leaf's word can never match `smort`, so every such leaf would silently fall through to the distance test and could be claimed provably-missed. One-line guard: assert `np.all(st["depth"] == order)` in `measure_stats` so the assumption fails loudly instead.

---

benchmarks/measure_straddle_overrefinement.py
```diff
@@ -0,0 +1,376 @@
...
+        "flat_cells": int(flat_cur),
+        "straddle_leaves": int(len(morton)),
+        "per_cause": per_cause,
+        "over_refined": int(over.sum()),
+        "over_refined_outside": int(len(removed)),
+        "over_refined_inside": int((over & fill).sum()),
+        "ambiguous": int(ambiguous.sum()),
+        "moc_hyp": int(len(moc_hyp)),
+        "flat_hyp": int(flat_hyp),
+        "moc_reduction_pct": 100.0 * (1.0 - len(moc_hyp) / len(moc)),
```
espg: 🤖 *from Claude (review)*

**Inconsistent zero-guard.** `flat_reduction_pct` divides by `max(flat_cur, 1)` but the two MOC ratios (`moc_reduction_pct` here and `moc_ceiling_reduction_pct` below) divide by bare `len(moc)`. An empty cover is not reachable for the fixed suite (a valid polygon always covers its vertex leaves), but the guard was clearly intended — make the three denominators consistent.

---

benchmarks/measure_straddle_overrefinement.py
```diff
@@ -0,0 +1,376 @@
...
+        if omega < 1e-12:
+            continue
+        n = max(1, int(np.ceil(omega / delta)))
+        max_step = max(max_step, omega / n)
+        t = np.arange(n) / n  # endpoint b belongs to the next edge
+        pts = (
+            np.sin((1 - t) * omega)[:, None] * a + np.sin(t * omega)[:, None] * b
+        ) / np.sin(omega)
+        out.append(pts / np.linalg.norm(pts, axis=1, keepdims=True))
+    return np.concatenate(out), max_step
```
espg, line 187: 🤖 *from Claude (review)*

Minor: `np.concatenate(out)` raises `ValueError` if every edge is degenerate (`out == []`, e.g. all-coincident vertices). Unreachable for the committed suite; fine to leave for a dev script, noting it so the failure mode is known rather than mysterious.

---

src_rust/src/coverage.rs
```diff
@@ -27,6 +27,11 @@
 //! is unambiguously the interior); hemisphere-plus rings are never reordered, so
 //! supply those CCW (holes CW) exactly as the right-hand rule prescribes.
 
+/// Cause-tagged straddle instrumentation for the issue #90 measurement; a
+/// dev-only cargo feature (cfg flag, no dependency), absent from release builds.
+#[cfg(feature = "descent-stats")]
+pub mod descent_stats;
```
espg, line 33: 🤖 *from Claude (review)*

**Module size (CLAUDE.md §4).** `coverage.rs` was already ~1194 lines before this PR (over the "~1000 lines without prior discussion" cap) and the cfg hooks add ~47 more (now ~1241). The bulk of the instrumentation was correctly extracted to `coverage/descent_stats.rs`, and the hooks themselves cannot live elsewhere, so the worsening is minimal — but the pre-existing overage should be on the maintainer's radar; if a split is wanted it belongs in its own PR.

---

src_rust/src/coverage/descent_stats.rs
```diff
@@ -0,0 +1,157 @@
...
+    pub cause: Cause,
+    pub fill: bool,
+    pub center: Vec3,
+    pub circ: f64,
+}
+
+/// Everything recorded since the last [`take`].
+pub struct Stats {
+    /// Straddle-stopped leaves per cause.
+    pub leaf: [u64; N_CAUSES],
```
espg, line 58: 🤖 *from Claude (review)*

Minor (simplification): `leaf: [u64; N_CAUSES]` is fully derivable from `leaves[*].cause` — `record_leaf` maintains both under one lock precisely so they can never disagree. Only `internal` carries non-derivable information. Keeping the counter array is a (cheap) second source of truth; dropping it and deriving `leaf_counts` at `take()`/binding time would shrink the invariant surface. Fine either way for a dev-only feature.

---

benchmarks/measure_straddle_overrefinement.py
```diff
@@ -0,0 +1,376 @@
...
+ORDERS = [6, 9, 11]
+SAMPLE_FRACTION = 0.2  # sample spacing as a fraction of the cell resolution
+SAFETY_RAD = 1e-6
+
+
+# ---------------------------------------------------------------------------
+# Polygon suite
+# ---------------------------------------------------------------------------
+
+def _unit(lats, lons):
```
espg, line 64: 🤖 *from Claude (review)*

Minor (reuse): `_unit`/`_latlon` and `cell_resolution_rad` (line 166) re-implement conversions that exist in `mortie/geometry.py` (`_xyz_to_latlon`) and as `sqrt(pi/3)/nside` in Rust `coverage::cell_resolution_rad`. Deliberate here — the measurement should stay independent of the library's own geometry helpers, and the library copies are private — but noting the duplication so a future formula change knows this script pins its own.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of both phases. Verified and NOT flagged: the feature-off restructure of `node_straddles` / `edge_hits_cell_edge` is semantically identity (same short-circuit order and return values; `cargo clippy` clean of `let_and_return` in both configs, CodSpeed reports no perf change); every `node_straddles == true` path tags a cause before either caller consumes it, so the `take_cause` fail-loud cannot fire on the current taxonomy; the "provably missed" classifier is sound against corner grazes between samples — a boundary point inside a leaf forces a sample within `circ + spacing/2` of the centre, so such a leaf can only classify as intersecting or ambiguous, never missed. Seven findings posted inline: one CI-visibility tradeoff (the feature is never built in CI), one latent script assumption worth a loud assert, and five minor/informational.

---

benchmarks/measure_straddle_overrefinement.py
espg: 🤖 *from Claude*

Addressed in dc1e257: `measure_stats` now asserts `np.all(st["depth"] == order)` with a comment explaining why exact mode is required. Reran the full stats sweep after the change — results identical.

---

benchmarks/measure_straddle_overrefinement.py
espg: 🤖 *from Claude*

Addressed in dc1e257: both MOC ratio denominators now use `max(len(moc), 1)`, matching the flat guard.

---

benchmarks/measure_straddle_overrefinement.py
espg: 🤖 *from Claude*

Intentionally left: an all-degenerate-vertex polygon is invalid input for the suite this script pins, and the `ValueError` from `np.concatenate([])` fails loudly at the right spot. Not adding a guard for an unreachable path in a dev script.

---

espg:
please review this (again), and let me know if there's anything for the new plan in #107 that impacts us here... also, let me know on phasing (i.e., should that (c) PR land first)

espg:
🤖 *from Claude*

Reviewed against current `main` (f3b2bcc). Bottom line: **this PR survives (c) and should land before it, not after** — and re-reading it in light of the (c) analysis turned up something worth your attention: the anomaly I flagged on this PR two weeks ago is the same bug I just diagnosed on #107, which I did not realize at the time.

## 1. The anomaly on this PR was the winding defect

My [earlier comment](https://github.com/espg/mortie/pull/112#issuecomment-4934921227) reported a hemisphere+ ring where "both windings produce the identical 45.6%-of-sphere cover, yet the ring's two complementary interiors are ~55.8% and ~44.2%", with ~53% of the large side's interior points missing from the cover. I flagged it as an incidental oddity and switched to a different test ring.

It is a clean symptom of the `ring_winding_at` antisymmetry (`src_rust/src/sphere.rs:434-452`, derived in my [comment on #107](https://github.com/espg/mortie/issues/107#issuecomment-5080416398)). `w(-x) = -w(x)`, so:

- **Both windings give the same cover** because reversing a ring negates the winding everywhere, and the `w > π` test then selects the *same* set — the one region where `w = +2π`. Winding reversal cannot change the answer, which is exactly what I observed and could not explain.
- **The large side is ~53% under-covered** because the missing part is precisely the antipodal lens — the points whose antipode is also interior, where the sum cancels to 0 and a truly-interior point reads outside. The 55.8% interior losing ~53% of its points to land at 45.6% is the right order of magnitude for that lens.

So that reproducer is a second, independently-arrived-at witness for the defect, on a different ring family from PR #113's. Worth keeping — I'd add it to the (c) PR's phase-1 acceptance set alongside the `#[ignore]`d test in `sphere/tests.rs`, since it exercises the *cover* path rather than the bare predicate.

## 2. Does (c) impact this PR's content?

**No.** I checked each piece:

- The **cause taxonomy** (`VertexLeaf`, `QuadCross`, `QuadTouch`, `CornerParity`, `NearPoleBulge`) is a classification of *why `node_straddles` returned true*. (c) changes what the crossing predicate answers at degenerate configurations; it does not remove or rename any of those causes. `QuadTouch` — the closed-set exact-incidence branch excluded from the over-refinement count — is the one (c) touches, and only if we take the fold-in decision from #107 question (5), which changes what that branch *decides*, not that it exists.
- The **tagging sites** are the `node_straddles` true-return paths and the two descent drivers. (c) does not restructure the descent.
- The **`descent-stats` feature, the collector, the binding, and the measurement script** are all orthogonal to the predicate.

The one thing that *would* change is the phase-2 **numbers**: the measurement re-tests each straddle-stopped leaf against sampled boundary geometry, and if (c) alters boundary verdicts, some leaves reclassify. But that argues for landing this first and re-running afterward, not for rewriting it.

## 3. Phasing — this PR first, and here's why it matters

**Recommendation: merge this before the (c) PR.**

1. **It's finished and green; (c) is a multi-phase rewrite of the crossing layer.** Holding a complete, zero-runtime-cost instrumentation PR behind that is a long block for no benefit.
2. **It becomes (c)'s measurement harness.** The output contract you set on #107 — "monitor for divergence in both performance and output, flag it, investigate, discuss" — needs an instrument that reports *which cells changed and why they were refined*. That is exactly what this PR built. If it lands first, (c) inherits a ready-made before/after apparatus and the divergence report is a diff of its output; if it lands after, (c) has to hand-roll something weaker to satisfy its own gate.
3. **It gives a clean pre-(c) baseline.** Cause counts measured on today's predicate are the reference the post-(c) numbers get compared against. Taking that snapshot is only possible before (c) lands.
4. **Merge risk is low in this direction.** This PR's only `coverage.rs` change is feature-gated tagging plus one semantically-identical `A || B` → two-`if` split. (c) will rewrite parts of `coverage.rs`, so it will carry the conflict either way — but resolving feature-gated tag lines into a rewritten predicate is much easier than re-deriving the instrumentation on top of it.

So: **this PR → (c) PR → re-run the measurement → then the #90 phase-3 relaxation decision** on post-(c) numbers. The relaxation is where (c) actually pays off for #90, since a trustworthy boundary verdict is what you'd be relaxing toward.

## 4. Standing items on this PR (unchanged, still needing your call)

- The `Mutex`-vs-thread-local deviation from the plan sketch — flagged for visibility, no action requested.
- `descent_stats` is `pub` so the integration-test binary can reach it; it only exists under the dev feature.
- **`benchmarks/measure_straddle_overrefinement.py` is committed on the branch** per the #78 "run and reported" precedent, and I offered to drop it before merge. Given point (3) above I'd now argue for **keeping** it — it becomes the tool that produces (c)'s divergence report, and re-deriving it later is wasted work. Your call; it's a one-line deletion either way.

No code changes made here — this is a review response. Leaving the PR as-is with `implement`; clearing `waiting` since you've responded, and it's actionable on your merge decision rather than on more work from me.

---

PR #121, opened 2026-07-20
Decimal kind suffix p for point ids (issue #120)

### LINKED ISSUE #120: decimal repr kind suffix: emit p for point ids, accept on parse (spec §2/§4)
🤖 *from Claude*

espg-ruled on the PR #118 review (2026-07-21), extending the #62 spec page's decimal grammar: the render/interchange form gains an **optional kind suffix `p`** on point ids, making point-ness representable in strings and the decimal round-trip **lossless for both kinds**.

## Grammar (normative on `docs/specification.md` §2/§4, PR #118)

```text
morton-decimal = ["-"] base-digit *order-digit [kind-suffix]
kind-suffix    = "p"    ; POINT ids only ⇒ legal only on full order-29 strings
```

- **Render**: `to_decimal_repr` / `decimal_repr` / `to_decimal` emit a terminal `p` on point words (e.g. `-62…21p`); area words render unchanged.
- **Parse**: a `p`-marked string ⇒ the **point** word (suffix `48 + t28·4 + t29`); an unmarked string ⇒ the **area** word, always (the existing tie-break becomes the unmarked-string rule — every pre-suffix string is an area context, fully backward compatible). `p` on a sub-order-29 string is malformed.
- **Paths never carry it**: points don't live in paths; the hive path grammar is unchanged. `hive_path` on a point word should refuse loudly rather than leak a `p` into a path component.
- **Round-trip contract**: `word → string → word` is the identity for both kinds; golden vectors for the marked point round-trip and the unmarked backward-compat case.

Rationale recorded on the spec page: no letter occurs anywhere in the base grammar (sign, base digit, digits 1–4), so a letter suffix is maximally distinguishable, greppable, and inert in shells/markup/URLs; `*` and terminal `.` were rejected (glob/markdown hazards; trailing-dot stripping).

## Implementation surfaces

- `src_rust/src/decimal_morton.rs::to_decimal_repr` — append `p` when `Kind::Point` (the `Decoded.kind` field already carries it).
- `mortie/morton_index.py::_decimal_to_word` — accept the marked form (strip `p`, require order 29, emit the point word: same prefix+body as the area parse, suffix swapped to `48 + t28·4 + t29`).
- `MortonIndexArray.to_decimal` — widen `"<U31"` → `"<U32"` (the marked form is one char wider) and refresh the now-stale non-injectivity docstrings.
- `MortonIndexArray.hive_path` — refuse point words (see above).

## Known consumers pinning the old behavior

- `mortie/tests/test_packed_golden.py` (~312): asserts point and area render **identically** and the point's string parses to the *area* word — both flip under this change; the test updates to the new lossless contract (the unmarked tie-break assertion stays, via the area word's own string).
- zagg: unaffected — every string zagg renders or parses is an area word (verified by grep; shard keys/paths are area words by construction).

Refs #62; spec text in PR #118 (§2 kind-suffix bullet, §4 parse rules).

espg:
Closes #120. Refs #62 (normative grammar in PR #118, `docs/specification.md` §2/§4).

Implements the espg-ruled decimal **kind suffix `p`**: point ids render with a terminal `p` (e.g. `-62…21p`), the parser accepts marked and unmarked forms, and the decimal round-trip becomes **lossless for both kinds**. Unmarked strings keep the area tie-break — fully backward compatible (every pre-suffix string is an area context).

## What changed

- `src_rust/src/decimal_morton.rs::to_decimal_repr` — appends `p` when `Decoded.kind == Kind::Point`; area words render byte-identically to before.
- `mortie/morton_index.py::_decimal_to_word` — accepts the marked form: strip `p`, require order 29 (`p` below 29 is malformed), emit the **point** word (same prefix+body as the area parse; suffix `48 + t28·4 + t29` per the spec §1/§4 formulas). Unmarked strings parse to the area word, unchanged.
- `MortonIndexArray.to_decimal` — fixed width `<U31` → `<U32` (the marked form is one char wider); stale non-injectivity docstrings refreshed (`decimal_repr` is now injective across kinds).
- `MortonIndexArray.hive_path` — **refuses point words** with a pointed error (paths never carry the suffix; points don't live in paths — spec §2). `from_hive_path` likewise rejects a `p`-marked leaf stem.
- Golden vectors: new `mortie/tests/test_point_suffix.py` (marked point round-trip, unmarked backward-compat tie-break, sub-29 `p` rejection, path refusals, `<U32` untruncated emit, scalar `str()` carries the suffix).

## Phases

- [x] Phase 1 — repr emit + parse accept + surface updates + golden vectors

## Consumers pinning the old behavior (updated in this PR)

- `mortie/tests/test_packed_golden.py::test_golden_repr_not_injective_point_vs_area` — pinned identical point/area renders and point-string→area-word; now pins the lossless two-kind round-trip plus the unmarked tie-break.
- `mortie/tests/test_packed_golden.py::test_golden_string_emit_layer` and `test_morton_index.py::test_to_decimal_fixed_width` — pinned `<U31`; now `<U32`.
- zagg: **unaffected** — every string zagg renders or parses is an area word (shard keys/paths by construction), unmarked and byte-identical; verified by grep on the zagg tree.

## Testing

- Full `pytest mortie/tests`: **680 passed, 11 skipped** against a local `pip install -e .` (maturin) build with the Rust change compiled in; `flake8 mortie --select=E9,F63,F7,F82` clean.
- `cargo test --lib`: **198 passed, 0 failed** in the review worktree — the earlier "cannot link locally" caveat no longer applies here. This also caught a Rust unit test (`decimal_repr_point_matches_area_of_same_path`) that still pinned the old identical-render behavior; it is renamed and updated to the lossless point == area + `p` contract (review fold F1). CI's Rust jobs remain the authority.

## Questions for review

1. `hive_path` refusing point words is new behavior (previously a point word silently produced an area-looking path — the pre-suffix ambiguity). Loud refusal matches "points don't live in paths"; confirm vs silently rendering the unmarked area form.

### DIFF
--- a/mortie/morton_index.py
+++ b/mortie/morton_index.py
@@ -74,12 +74,12 @@ def _decimal_to_word(s):
     base = lead + 5 if text.startswith("-") else lead - 1
     nested = np.asarray([(base << (2 * order)) | within], dtype=np.uint64)
     if point:
-        # The point word shares prefix+body with the area parse; only the
-        # suffix differs: area = 28 + t28*5 + t29 + 1, point = 48 + t28*4 +
-        # t29 (spec section 1/section 4 tie-break, stored 0..=3 tuples).
-        area = int(_rustie.rust_mi_from_nested(nested, order)[0])
-        t28, t29 = int(digits[-2]) - 1, int(digits[-1]) - 1
-        return (area & ~0x3F) | (48 + t28 * 4 + t29)
+        # Delegate the point-suffix layout to the kernel rather than
+        # reimplement it here: `nested` is the full order-29 NESTED id (order
+        # == MAX_ORDER is guaranteed above), and the kernel packs the point
+        # word directly (suffix = 48 + t28*4 + t29, spec section 1/section 4
+        # tie-break). Keeping the formula in one place -- the kernel.
+        return int(_rustie.rust_mi_from_nested_point(nested)[0])
     return int(_rustie.rust_mi_from_nested(nested, order)[0])
 
 
--- a/mortie/morton_index.py
+++ b/mortie/morton_index.py
@@ -36,12 +36,16 @@ def _decimal_to_word(s):
     """Parse a decimal Morton string back to its packed word (issue #104).
 
     The inverse of the decode-through-kernel repr: sign + leading base digit
-    (``1..6``) then one ``1..4`` digit per order. Raises ``ValueError`` on any
-    malformed id. Order-29 strings parse to the *area* word -- the point/area
-    flag is not part of the decimal form, so the point word is unreachable
-    from its repr (the documented non-injectivity).
+    (``1..6``), one ``1..4`` digit per order, and an optional terminal ``p``
+    kind suffix (spec section 4, issue #120). Parse rules: a ``p``-marked
+    string (legal only at order 29) yields the POINT word; an unmarked
+    string always yields the AREA word -- the tie-break for the one
+    ambiguous form, and fully backward compatible (every pre-suffix string
+    is unmarked). Raises ``ValueError`` on any malformed id.
     """
-    body = s[1:] if s.startswith("-") else s
+    point = s.endswith("p")
+    text = s[:-1] if point else s
+    body = text[1:] if text.startswith("-") else text
     if not (body.isdigit() and body.isascii()):
         raise ValueError(f"malformed decimal Morton id {s!r}")
     lead, digits = int(body[0]), body[1:]
@@ -54,6 +58,12 @@ def _decimal_to_word(s):
         raise ValueError(
             f"decimal Morton id {s!r}: order {order} exceeds {MAX_ORDER}"
         )
+    if point and order != MAX_ORDER:
+        raise ValueError(
+            f"decimal Morton id {s!r}: the kind suffix 'p' is legal only on "
+            f"full order-{MAX_ORDER} point ids (points exist only at order "
+            f"{MAX_ORDER})"
+        )
     within = 0
     for d in digits:
         if not "1" <= d <= "4":
@@ -61,8 +71,15 @@ def _decimal_to_word(s):
                 f"decimal Morton id {s!r}: digit {d} outside 1..4"
             )
         within = (within << 2) | (int(d) - 1)
-    base = lead + 5 if s.startswith("-") else lead - 1
+    base = lead + 5 if text.startswith("-") else lead - 1
     nested = np.asarray([(base << (2 * order)) | within], dtype=np.uint64)
+    if point:
+        # The point word shares prefix+body with the area parse; only the
+        # suffix differs: area = 28 + t28*5 + t29 + 1, point = 48 + t28*4 +
+        # t29 (spec section 1/section 4 tie-break, stored 0..=3 tuples).
+        area = int(_rustie.rust_mi_from_nested(nested, order)[0])
+        t28, t29 = int(digits[-2]) - 1, int(digits[-1]) - 1
+        return (area & ~0x3F) | (48 + t28 * 4 + t29)
     return int(_rustie.rust_mi_from_nested(nested, order)[0])
 
 
@@ -504,10 +521,9 @@ def _build_classes():
             produced by *decoding* each word (the canonical render-only repr;
             backward-compatible with the legacy ``str(legacy_i64)`` for orders
             0..=18, the natural extension for 19..=29). Raises ``ValueError`` on
-            any empty / invalid word. Note the repr is not injective across
-            kinds: an order-29 *point* renders identically to the order-29
-            *area* on the same path (the point/area flag is not part of the
-            decimal form).
+            any empty / invalid word. Point words carry the terminal ``p``
+            kind suffix (spec section 2, issue #120), so the repr is
+            injective across kinds and the round-trip is lossless.
             """
             return _rustie.rust_mi_decimal_repr(self._data)
 
@@ -515,14 +531,13 @@ def _build_classes():
             """Vectorized decimal-string emit as a fixed-width numpy array.
 
             The always-strings interchange convention from issue #48: returns
-            the decode-through-kernel decimal strings as a ``"<U31"`` numpy
-            array (sign + base digit + 29 order digits is the widest form, so
-            the width is order-independent and stable across arrays). Raises
-            ``ValueError`` on any empty / invalid word. Same non-injectivity
-            note as :meth:`decimal_repr`: point and area render identically at
-            order 29.
+            the decode-through-kernel decimal strings as a ``"<U32"`` numpy
+            array (sign + base digit + 29 order digits + the point ``p`` kind
+            suffix is the widest form, so the width is order-independent and
+            stable across arrays). Raises ``ValueError`` on any empty /
+            invalid word. Point words carry the ``p`` suffix (issue #120).
             """
-            return np.asarray(self.decimal_repr(), dtype="<U31")
+            return np.asarray(self.decimal_repr(), dtype="<U32")
 
         def to_legacy_i64(self):
             """Emit the legacy signed decimal ``int64`` form (orders <= 18).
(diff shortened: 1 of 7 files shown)


src_rust/src/decimal_morton.rs
```diff
@@ -633,6 +634,12 @@ pub fn to_decimal_repr(word: u64) -> Option<String> {
         // decode guarantees every tuple is 0..=3; read as the digit 1..=4.
         s.push(char::from(b'1' + t));
     }
+    if dec.kind == Kind::Point {
+        // Kind suffix (spec section 2, issue #120): point ids carry a
+        // terminal `p` in the render form; paths never do (the hive-path
+        // surface refuses point words instead).
+        s.push('p');
```
espg, line 641: 🤖 *from Claude (review)*

**F1 (blocking) — a Rust unit test still pins the old identical-render behavior; CI's cargo job will be red.**

`decimal_morton.rs:1347` `decimal_repr_point_matches_area_of_same_path` asserts `to_decimal_repr(point) == to_decimal_repr(area)`, which this line (the `p` push) deliberately falsifies. The PR updated the Python golden (`test_golden_repr_not_injective_point_vs_area`) to the lossless two-kind contract but missed the Rust counterpart — understandably, since the PR notes cargo could not link locally. It links fine in a clean checkout here, and the test fails:

```
test decimal_morton::tests::decimal_repr_point_matches_area_of_same_path ... FAILED
  left: Some("541234123412341234123412341234p")
 right: Some("541234123412341234123412341234")
test result: FAILED. 197 passed; 1 failed; 1 ignored
```

Update that test (name + body) to the new contract: point renders as `area + "p"`, mirroring the Python golden. Every other Rust test passes.

espg, line 641: 🤖 *from Claude (review)*

**F3 (minor) — `prefix_trie` is a missed render surface: it consumes `to_decimal_repr`, which now emits `p`, and silently mishandles it.**

`prefix_trie.rs:63` renders every word through `to_decimal_repr`, then `build_slot_grid` maps each char with `(d - b0) as i8`. A point word's trailing `p` (0x70) becomes slot value 64 — outside the documented `0..=9` digit range (`slot_char` happens to round-trip it back to `p`, so no panic). Building a trie over a point word yields a characteristic ending in `p`, and `MortonChild.cell_area` counts the `p` as an extra order digit (order-30 area → 4× too small; reproduced `cell_area ≈ 8.67e-19` for an order-29 point). Since `hive_path` now *loudly refuses* point words, the trie is the one remaining string surface that silently accepts them. Consider a matching guard (reject point words), or at least a doc note that the trie is area-only. Not blocking — feeding points to the coverage trie is off the documented path.

mortie/morton_index.py
```diff
@@ -54,15 +58,28 @@ def _decimal_to_word(s):
...
-    base = lead + 5 if s.startswith("-") else lead - 1
+    base = lead + 5 if text.startswith("-") else lead - 1
     nested = np.asarray([(base << (2 * order)) | within], dtype=np.uint64)
+    if point:
+        # The point word shares prefix+body with the area parse; only the
+        # suffix differs: area = 28 + t28*5 + t29 + 1, point = 48 + t28*4 +
+        # t29 (spec section 1/section 4 tie-break, stored 0..=3 tuples).
+        area = int(_rustie.rust_mi_from_nested(nested, order)[0])
+        t28, t29 = int(digits[-2]) - 1, int(digits[-1]) - 1
+        return (area & ~0x3F) | (48 + t28 * 4 + t29)
```

**F2 (minor / quality) — the point branch reimplements the kernel's point encoding in Python instead of delegating.**

`(area & ~0x3F) | (48 + t28 * 4 + t29)` duplicates the kernel's `POINT_BASE` / `build_point_suffix` bit layout in Python (and pays for an extra area decode to get `area`). The module docstring says the bit layout is the kernel's job, and the kernel already exposes the vectorized `_rustie.rust_mi_from_nested_point(nested)` — which the tests use — returning the point word directly. The whole `if point:` branch collapses to:

```python
if point:
    return int(_rustie.rust_mi_from_nested_point(nested)[0])
```

Verified bit-identical to the current arithmetic across all 12 base cells (northern and southern). This keeps the suffix formula in one place (`decimal_morton.rs`), so a future change to the point encoding can't silently drift the parser. Correct today; flagging as a maintainability/duplication concern, not a bug.

mortie/tests/test_point_suffix.py
```diff
@@ -0,0 +1,80 @@
...
+        s = point.decimal_repr()[0]
+        assert s.endswith("p")
+        # Area words render unchanged -- no suffix at any order.
+        assert not area.decimal_repr()[0].endswith("p")
+
+    def test_scalar_str_carries_suffix(self):
+        point, _ = _pair()
+        assert str(point[0]).endswith("p")
+
+    def test_to_decimal_is_u32_and_untruncated(self):
```
espg, line 43: 🤖 *from Claude (review)*

**F4 (nit) — the width golden uses a northern point, so it can't actually catch a `<U32` → `<U31` regression.**

`_pair()` uses `(45, 45)`, a *northern* order-29 point whose repr is 31 chars (`141…p`) — it fits in `<U31` too, so the "untruncated" content assertion (`out[0].endswith("p")`) passes even at the old width; only the separate `out.dtype == <U32` line actually pins it. The true 32-char widest form is a *southern* order-29 point (`-6…p`), which loses its `p` under `<U31` (verified: 32 chars, truncates). Rendering the `_pair()` (or this test) from a southern lat/lon would make the untruncation check non-tautological and directly exercise the boundary the `<U31→<U32` change exists for. Minor — the dtype assert already guards the regression.

src_rust/src/decimal_morton.rs
espg: 🤖 *from Claude*

Fixed in d34cee5. Renamed `decimal_repr_point_matches_area_of_same_path` -> `decimal_repr_point_is_area_repr_plus_p_suffix` and flipped it to the lossless contract: it now asserts `to_decimal_repr(point) == to_decimal_repr(area) + "p"` (same base + body, point carries the trailing kind mark). `cargo test --lib` links and passes in this worktree — 198 passed, 0 failed (the earlier "cannot link locally" note in the PR body is stale for this environment); the renamed test is green.

mortie/morton_index.py

Fixed in 557276b. The `if point:` branch now collapses to `return int(_rustie.rust_mi_from_nested_point(nested)[0])` — `nested` is the full order-29 NESTED id (order == MAX_ORDER is guaranteed above), so the kernel packs the point word directly and the extra area decode is gone. Brief comment retained citing the spec section 1/section 4 suffix formula. Python suffix tests green.

mortie/tests/test_point_suffix.py
espg: 🤖 *from Claude*

Fixed in c3441e7. `_pair()` now uses a *southern* order-29 point (lat=-45, lon=45) whose repr is the widest 32-char form `-314441444444444111111414111414p` (verified: leading `-`, trailing `p`, len 32). It truncates under `<U31`, so the untruncated-content assertions (`out[0].endswith("p")`, `out[0] == decimal_repr()[0]`) now genuinely exercise the `<U31` -> `<U32` boundary instead of passing tautologically.

src_rust/src/decimal_morton.rs
espg: 🤖 *from Claude*

Fixed in e242fb4 — the trie now refuses point words loudly.

Choice: I refused at the Python binding surface (`split_children_rust` in lib.rs) rather than thread a `Result` through the pure-Rust `split_children_flat` core. Rationale: the reviewer flagged threading an error through the trie API as potentially disproportionate, and it is here — `split_children_flat` returns a bare tuple consumed by ~12 Rust unit tests, so a `Result` flip is pure churn for no added coverage (the only non-test caller is this binding). The binding already returns `PyResult`, and it is the actual Python-facing "trie" surface — the same layer where `hive_path` enforces the identical contract. So a pre-scan there (`data.iter().any(|w| kind_of(w) == Kind::Point)`) raises `ValueError: ... points do not live in paths (spec section 2, issue #120) ...` before any corrupt trie is built. This is stronger than the documented-limitation + follow-up alternative you offered.

Also added: a doc note at the render site (`reprs` in prefix_trie.rs) recording that point words are refused upstream and never reach the slot grid, plus a Python regression test `test_split_children_refuses_point_words` asserting the raise. Full pytest green (680 passed), `cargo test --lib` green (198 passed).

espg:
🤖 *from Claude*

**CodSpeed check (the one red item): false positive for this PR.** The flagged regression is `coverage_triangle[4]` (−12.67%) — the polygon-coverage fill benchmark, whose code path this PR does not touch (the diff is decimal repr/parse: `to_decimal_repr`'s suffix append, the Python parser, the prefix-trie binding guard, and tests). CodSpeed's own report carries the *"different runtime environments detected"* warning for exactly this comparison, and the same run shows an equally unrelated +21.65% "improvement" in `fast_norm2mort_orders[6]` — both are cross-environment noise, not effects of this change. The two benchmarks that COULD see the suffix append (repr-side) are in the untouched set. Acknowledging on the CodSpeed dashboard is left to espg (side-effecting); happy to investigate further if the regression reproduces on a same-environment rerun.

PR #125, opened 2026-07-21
fix test_spec_page order-29 tie-break probe (issue #123)

### LINKED ISSUE #123: main is red: test_spec_page tie-break test predates #120's p-suffix rendering
🤖 *from Claude*

**Pre-existing failure on `main` (@ 4b57a1f), surfaced during PR #122's fold** (evidence: Tests run 29787103143 red on 3.10 + 3.12 with the identical failure; #122's own diff is disjoint and its 23 new tests pass).

`mortie/tests/test_spec_page.py::TestDecimalParseTieBreak::test_order29_string_parses_to_area_word` fails with `ValueError: invalid literal for int() with base 10: 'p'` — issue #120's "delegate point suffix to kernel" made `decimal_repr()` render the terminal `p` kind-suffix on point words (spec §2: the render form MAY carry `p` on point ids), but the test at line ~87 still constructs its probe via `int(dec[-1])` on a repr that is now `p`-terminated.

**Likely fix is test-only** (the behavior matches the spec): build the unmarked order-29 probe string deliberately — e.g. `dec.rstrip('p')` (or render an *area* order-29 word instead of a point) — so the test exercises what it means to pin: *the unmarked string parses to the AREA word per the §4 tie-break*. Worth one check while there: that the tie-break assertion itself still passes once the probe is well-formed, and that a `p`-marked string round-trips to the point word (which #120's own tests may already cover).

**Why it matters beyond hygiene**: it blocks PR #122's test matrix from going green (and therefore the release moczarr#8 adoption waits on). Not fixed inside #122 per the pre-existing-failure discipline; flagged there with the same diagnosis.

Refs: #120, #62 (the spec page + drift-pin), PR #122 (where it surfaced), espg/moczarr#8 (downstream).

espg:
Closes #123

## What / why

`main` was red: `TestDecimalParseTieBreak::test_order29_string_parses_to_area_word` failed with `ValueError: invalid literal for int() with base 10: 'p'`.

The test derives `t28, t29` from `dec[-2]` / `dec[-1]`, but issue #120's "delegate point suffix to kernel" made `decimal_repr()` terminate a point word with the `p` kind suffix (spec §2 — the render form carries `p` on point ids). So `dec[-1] == 'p'` and `int('p')` raises.

**The runtime is correct per spec — this is test-only.** §2 says the render carries `p`; §4 says an *unmarked* order-29 decimal string ties to the AREA word. The pin, not the behavior, was stale, so no kernel / `decimal_repr` change is warranted.

## Approach

Strip the marker to build the *unmarked* probe the tie-break is actually about (`dec = marked.rstrip("p")`), then keep the existing tie-break assertions on that well-formed string. Also, while there:
- assert the point repr `marked.endswith("p")` (documents #120's render form),
- keep the tie-break checks (`parsed & 0x3F == 28 + t28*5 + t29 + 1`, same-path prefix, `parsed != word`, non-injectivity of the *unmarked* string),
- add that the `p`-**marked** string is unambiguous and round-trips back to the POINT word (`_decimal_to_word(marked) == word`) — the marker is what disambiguates.

The non-injectivity assertion now compares the area word's repr against the unmarked `dec` (the area word renders without `p`), which is the correct pin.

## How tested

- `maturin develop --release` (rebuilt the Rust ext; no Rust change)
- `flake8 mortie --select=E9,F63,F7,F82` — clean
- `pytest -v`
  - before: `1 failed, 3 passed` in `test_spec_page.py` (full suite red)
  - after: `4 passed` in `test_spec_page.py`; full suite `684 passed, 11 skipped`

## Questions for review

None — the fix is faithful to the test's stated purpose (unmarked → area, marked → point) and leaves runtime behavior untouched. Flagging only that `main`-red is resolved by this single test-only change.

---

### DIFF
--- a/mortie/tests/test_spec_page.py
+++ b/mortie/tests/test_spec_page.py
@@ -83,22 +83,31 @@ class TestDecimalParseTieBreak:
             np.array([45.0]), np.array([45.0]), points=True
         )
         word = int(np.asarray(arr._data, dtype=np.uint64)[0])
-        dec = arr.decimal_repr()[0]
+        # decimal_repr renders point words with the terminal 'p' kind suffix
+        # (spec §2, issue #120); the §4 tie-break concerns the *unmarked*
+        # order-29 string, so strip the marker to build the ambiguous probe.
+        marked = arr.decimal_repr()[0]
+        assert marked.endswith("p")
+        dec = marked.rstrip("p")
         t28, t29 = int(dec[-2]) - 1, int(dec[-1]) - 1
 
         # The point word sits in the point suffix region, at the documented
         # preorder slot 48 + t28*4 + t29 (§1).
         assert word & 0x3F == 48 + t28 * 4 + t29
 
-        # Tie-break: the parse yields the AREA word — same prefix+body (same
-        # path), area suffix 28 + t28*5 + t29 + 1 (§4).
+        # Tie-break: the unmarked parse yields the AREA word — same prefix+body
+        # (same path), area suffix 28 + t28*5 + t29 + 1 (§4).
         parsed = int(_decimal_to_word(dec))
         assert parsed & 0x3F == 28 + t28 * 5 + t29 + 1
         assert parsed >> 6 == word >> 6
         assert parsed != word
 
-        # Non-injectivity both ways: the area word renders the same string,
-        # so point-ness cannot round-trip through the decimal repr.
+        # The 'p'-marked string is unambiguous and round-trips to the POINT
+        # word (spec §2/§4, issue #120) — the marker is what disambiguates.
+        assert int(_decimal_to_word(marked)) == word
+
+        # Non-injectivity: the area word renders the same *unmarked* string,
+        # so point-ness cannot round-trip through the unmarked decimal repr.
         area = MortonIndexArray.from_words(np.asarray([parsed], dtype=np.uint64))
         assert area.decimal_repr()[0] == dec
 


mortie/tests/test_spec_page.py
```diff
@@ -83,22 +83,31 @@ def test_order29_string_parses_to_area_word(self):
             np.array([45.0]), np.array([45.0]), points=True
         )
         word = int(np.asarray(arr._data, dtype=np.uint64)[0])
-        dec = arr.decimal_repr()[0]
+        # decimal_repr renders point words with the terminal 'p' kind suffix
+        # (spec §2, issue #120); the §4 tie-break concerns the *unmarked*
+        # order-29 string, so strip the marker to build the ambiguous probe.
+        marked = arr.decimal_repr()[0]
+        assert marked.endswith("p")
+        dec = marked.rstrip("p")
```
espg, line 91: 🤖 *from Claude (review)*

Non-blocking nit (optional): `marked.rstrip("p")` strips *all* trailing `p`s, whereas the intent is to drop the single kind marker. It's harmless here — the body digits are `1..4` and the preceding `assert marked.endswith("p")` guarantees exactly one `p` to remove — so `rstrip` and `[:-1]` are equivalent on any well-formed input. `marked[:-1]` would state "drop the one marker" more precisely, but this is purely stylistic; no change required.

---

espg (commented):
🤖 *from Claude (review)*

**Assessment: LGTM (COMMENT only).** Verified locally: on `main` this test dies with `ValueError: invalid literal for int() with base 10: 'p'` at the `int(dec[-1])` line; on this branch `TestDecimalParseTieBreak` is `2 passed` (full `test_spec_page.py` green). The fix targets the exact root cause and is faithful to issue #123's intent.

Correctness checks I ran through:
- **Pins the intent, does not silence it.** `dec = marked.rstrip("p")` correctly rebuilds the *unmarked* order-29 probe after asserting the point repr carries `p` (spec §2 / issue #120). The §4 tie-break assertion (unmarked → AREA word, suffix `28 + t28*5 + t29 + 1`, same-path prefix, `parsed != word`) is preserved on a well-formed string, not weakened.
- **Strengthens, not weakens.** The added `_decimal_to_word(marked) == word` genuinely pins that the `p`-marked string round-trips to the POINT word — new coverage the old test lacked. The non-injectivity assertion now correctly compares the area word's repr against the *unmarked* `dec` (area renders without `p`), which is the right pin now that the marker disambiguates the two renders.
- **No off-by-one.** `t28 = dec[-2]-1`, `t29 = dec[-1]-1` operate on the last two body digits of the unmarked string exactly as before #120 broke it; the point slot `48 + t28*4 + t29` (→ 48..63) and area slot `29 + t28*5 + t29` (→ 29..47) ranges are consistent. No assertion is vacuous or tautological — the arithmetic formulas independently pin the Rust output.
- **Genuinely test-only.** Only `mortie/tests/test_spec_page.py` changed (+14/-5); no kernel / `decimal_repr` / runtime change, no `skip`, no `# noqa`, no lint-ignore. Style matches the surrounding class.
- **Governance.** Commit message is title-only (§3). PR body carries the substance — `Closes #123`, what/why, approach, how-tested, Questions-for-review — with no stray `🤖 *from Claude*` attribution in the description (§6). Draft, `implement` label.

One optional style nit left inline (`rstrip` vs `[:-1]`); non-blocking. Nothing here should hold the PR.

---

Issue #119, opened 2026-07-20
Unify order2res Earth model with the spec page sphere (R=6371.0088)


espg:
🤖 *from Claude*

## Ruling

Per espg, the spec-page resolution/Δ table (`docs/specification.md` §3, PR #118) now derives **both** columns — cell scale and cell area — from the exact HEALPix sphere at mean radius `R = 6371.0088 km`:

- cell area = `4πR² / (12·4^order)` (HEALPix is equal-area by construction);
- cell scale = `sqrt(area)` = `sqrt(4πR²/(12·4^order))`, the RMS cell spacing.

This replaces the historical flat `order2res` constant (`111 km/deg × 58.6323 × 0.5^order`, implied sphere `R ≈ 6366 km`) **on the page**. The page cell-scale column is ~0.2% larger than `order2res` at every order.

## Why the code was not changed in PR #118

`mortie.tools.order2res` is consumed **behaviorally** and changing its constant would shift real outputs, so PR #118 leaves the code untouched and documents the divergence as a code-vs-page note in §3. Consumers found:

- `mortie/tools.py` `res2display()` — prints per-order resolutions from `order2res`;
- `mortie/tests/test_coverage_boundary.py:78` — computes the coverage buffer pad as `pad_cells * order2res(order) / KM_PER_DEG`; a changed constant shifts the boundary-escape tolerance;
- `mortie/tests/test_tools.py` — pins `order2res` values directly (`test_order2res_basic`, `test_order2res_range`, `test_order2res_decreasing`).

## Migration note

Unifying means switching `order2res` to `sqrt(4πR²/(12·4^order))` with `R = 6371.0088 km` (or exposing a single `EARTH_RADIUS_KM` constant in `mortie.tools` and deriving both scale and area from it). That is a behavioral change: it must re-pin `test_tools.py`, re-check the `test_coverage_boundary.py` pad tolerance, and refresh `res2display` expectations. The spec-page drift pin (`test_spec_page.py`) already regenerates both columns from the sphere formula and would then import the unified constant instead of redefining `EARTH_RADIUS_KM` locally.

Tracked out of PR #118 so the spec page ships with the normative sphere values while the code migration is scheduled separately.

espg:
seems small-ish -- post plan here and implement

Issue #65, opened 2026-06-25
v1.0: refresh README benchmarks + add a cross-order benchmark table


espg:
🤖 *from Claude*

Sub-issue of #48 (v1.0 sweep), Tier-2.

## Scope
The README benchmark numbers are stale (pre packed-u64 / pre order-29). Refresh them and add the cross-order dimension espg asked for:

- Re-run the existing benches (`benchmarks/`, `bench_matrix.py`) on current `main` and update the README figures.
- Add a **cross-order benchmark table** (encode/decode/coverage timing across representative orders, e.g. 4 / 12 / 18 / 29) so the order-29 performance story is visible.
- Keep it reproducible — wire the table to the existing bench harness rather than pasting hand-run numbers.

## Acceptance criteria
README benchmark section updated; a documented command regenerates the cross-order table.

Refs #48.

espg:
gated on #122  ; post plan

espg:
🤖 *from Claude*

Plan for refreshing the README benchmarks and adding the cross-order table. Grounded in the current harness:

- **README benchmark table** lives at `README.md:30-38` — a 3-row Rust-vs-Python table (1k / 100k / 1.2M values), stale (pre packed-u64 / pre order-29).
- **`bench_matrix.py`** already regenerates the coverage-method matrix straight into `docs/coverage_methods.md` between `<!-- BENCH_MATRIX:START/END -->` markers (self-contained, drives off the bundled Antarctic drainage data). This is the reproducible-table precedent to mirror.
- **`benchmarks/test_bench_cpu.py`** is the CodSpeed CPU-bench surface (fixtures `coords_1k`, etc.; `geo2mort`/`norm2mort`/`morton_polygon`/`split_children`). There is **no** cross-order (4/12/18/29) encode/decode/coverage table yet — that's the new dimension.

### Acceptance criteria
- README's Rust-vs-Python figures re-measured on current `main` and updated (or replaced with a pointer to the generated tables).
- A **cross-order table** (encode `geo2mort` / decode `mort2geo` / coverage across orders {4, 12, 18, 29}) that is **regenerated by a committed script**, not hand-pasted — same START/END-marker pattern as `bench_matrix.py`, written into the README or a `docs/` page.
- A documented one-line command regenerates the table; cell counts stay deterministic, timings noted as machine/run-dependent (as `bench_matrix.py` already disclaims).

### Phases
[...]

PR #128, opened 2026-07-21
add HEALPix interchange guide (issue #63)

### LINKED ISSUE #63: v1.0 docs: HEALPix-ecosystem interchange guide
🤖 *from Claude*

Sub-issue of #48 (v1.0 sweep), Tier-2 docs.

## Scope
A guide showing how a mortie packed-u64 morton index interchanges with the wider HEALPix ecosystem:

- mortie morton ↔ HEALPix `(order, nested-pixel)` via `mort2healpix` / `norm2mort` / `mort2norm`.
- Round-trip against `cdshealpix` / `healpy` (`to_nested`/`from_nested`, `ang2pix`/`pix2ang`), the same oracle pattern the test-suite uses.
- Worked examples and the gotchas (base-cell prefix offset, unsigned vs signed-southern presentation, order self-encoding).

## Acceptance criteria
A markdown guide (under `docs/`) with runnable snippets; cross-checked against the ecosystem libraries already in the `test`/`docs` extras.

Refs #48.

espg:
Closes #63. Refs #48.

## What

Adds `docs/healpix_interchange.md`, a Tier-2 v1.0 guide for moving a packed
`morton_index` word to and from the wider HEALPix ecosystem, and links it from
the README next to the specification reference.

The guide covers:

- The conversion API — `mort2healpix` (word → `(nested_ipix, order)`),
  `mort2norm` / `norm2mort` (split/join around `base_cell * nside**2 + normed`),
  and `infer_order_from_morton` — grounded in the real signatures in
  `mortie/tools.py`.
- Round-trip against **cdshealpix** (`lonlat_to_healpix` / `healpix_to_lonlat`),
  the same oracle the suite uses in `test_morton_index.py` /
  `test_coverage_hemisphere.py`, plus importing a foreign NESTED id back into a
  word.
- Round-trip against **healpy** (`ang2pix` / `pix2ang`, `nside = 2**order`,
  `nest=True`, `lonlat=True`).
- The three boundary gotchas: (1) the 4-bit prefix is `base_cell + 1` and base
  cells 7–11 set bit 63, so words must stay `uint64` (reading `int64` corrupts
  southern cells); (2) order is self-encoded — never carry it out of band, but
  every ecosystem call needs it explicitly; (3) a bare `geo2mort` yields an
  order-29 point, and the null sentinel is the all-zero word.

## Approach

Every code block was executed against the built extension and the installed
ecosystem libs, and the pasted outputs are the real values. Snippet argument
order is called out where it differs (`geo2mort` is `(lat, lon)`; cdshealpix and
healpy are `(lon, lat)`). numpy stays the only runtime dep — cdshealpix is in
the `test` extra (the suite's oracle); healpy is not in any mortie extra, and
the guide says so and shows `pip install healpy`.

## Phases

- [x] Phase 1: write the guide, verify every snippet, wire into README.

## How tested

- `maturin develop --release` (rebuilt the Rust ext), then installed the `test`
  extra (`cdshealpix==0.8.1`, `astropy==8.0.1`) and `healpy==1.19.0` separately.
- Ran every doc code block verbatim as assertions — all pass (cell ids match
  cdshealpix and healpy exactly; centers agree to ~2e-9 deg).
- `flake8 mortie --select=E9,F63,F7,F82` — clean.
- `pytest -q` — 683 passed, 11 skipped, 1 failed. The single failure is the
  pre-existing `test_spec_page.py::TestDecimalParseTieBreak::test_order29_string_parses_to_area_word`
  (issue #123), unrelated to this docs-only change.

## Questions for review

- README placement: I linked the guide from the specification paragraph;
  `docs/arrow_interchange.md` currently isn't linked from the README at all, so
  there's no single docs index to slot into — happy to move the link or add a
  proper docs index if preferred.
- healpy is shown but is not in any mortie extra (only cdshealpix is, via
  `test`). Kept it as an optional `pip install healpy` block. Fine as-is, or
  should the healpy snippet be dropped to keep every snippet runnable purely
  from the declared extras?

🤖 Generated with Claude Code

---


docs/healpix_interchange.md
```diff
@@ -0,0 +1,186 @@
...
+that same relation.
+
+```python
+import numpy as np
+import mortie
+
+# geo2mort takes (lat, lon) and returns a length-1 array for scalar input,
+# so index [0] for a single word.
+m = mortie.geo2mort(-80.0, 120.0, order=6)[0]
+print(repr(m))                       # np.uint64(11570310392668225542)
```
espg: 🤖 *from Claude (review)*

The `# np.uint64(...)` expected-output comments here (and at line 151, `print(repr(w))`) are numpy-2-only. On numpy 1.x `repr(np.uint64(x))` prints the bare integer (`11570310392668225542`), so a reader on that stack sees a "mismatch" against the guide. numpy is the sole pinned runtime dep and no floor is stated in pyproject, so both are in scope. Cheap fixes: use `print(int(m))` (stable across versions) or note the outputs assume numpy ≥ 2. The scalar `mort2healpix`/`mort2norm` prints below (`# 37010 6`, `# 146 9 6`) are already version-stable since those return Python ints — only the two `repr` lines need it.

---

docs/healpix_interchange.md
```diff
@@ -0,0 +1,186 @@
...
+                    dtype=np.int64)
+print(np.array_equal(cell_ids, oracle))     # True
+# cell_ids == [1275068416, 67108860, 2617245699, 800177021,
+#              2962305922, 1556551876, 2997005193]
+
+# centers, mortie vs cdshealpix
+clon, clat = healpix_to_lonlat(cell_ids, depth=order)
+mlat, mlon = mortie.mort2geo(words)
+print(float(np.max(np.abs(mlat - clat.to_value(u.deg)))))   # ~1.98e-09
+print(float(np.max(np.abs(mlon - clon.to_value(u.deg)))))   # 0.0
```
espg, line 102: 🤖 *from Claude (review)*

Please confirm the pinned numeric outputs in this file were captured from an actual run, since they'll silently drift if not: the `cell_ids == [...]` list (line 84-85), the `# ~1.98e-09` lat tolerances, and especially this `# 0.0` for longitude. The lon-diff-of-exactly-zero relies on `mort2geo` returning longitude in `[0, 360)` (it does not wrap — only `mort2bbox` does), matching cdshealpix's `[0, 360)`. If `mort2geo` ever returned `[-180, 180)`, the eastern test points (lons 200/305/270.2) would diff by ~360 and this would not be `0.0`. Worth a one-line note that these were generated from the current build (as the arrow guide implicitly does), so a future reviewer can regenerate rather than guess. Not blocking — the snippets run regardless of the comment values.

---

docs/healpix_interchange.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Gotchas at the boundary
+
+### 1. The 4-bit prefix is `base_cell + 1`, and the word is unsigned
+
+The top nibble of a word stores the HEALPix base cell as `base + 1` (so the 12
+base cells occupy `1..=12`, and `0` is the empty/null sentinel). Base cells
+**7–11 set bit 63**, making the word a large unsigned value. Store and exchange
+the word as `uint64`; reinterpreting it as signed `int64` reads southern base
+cells as negative and corrupts the id.
```
espg: 🤖 *from Claude (review)*

Minor precision nit (optional): "reads southern base cells as negative" is slightly loose. The set that reads negative as `int64` is exactly the bit-63-set cells **7–11** (stated correctly one sentence up), but base cell 7 is *equatorial*, not south-polar (HEALPix base cells: 0–3 north, 4–7 equatorial, 8–11 south). The int64-negative property is purely "prefix ≥ 8 ⇒ bit 63 set", independent of hemisphere — that's a different mapping from the decimal `-1..-6` southern-sign in §2. Suggest "reads those base cells as negative" (or "base cells 7–11") to match the spec's own §1 wording ("base cells 7–11 set bit 63 and would read back negative"), which pointedly avoids the "southern" label here.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial pass on the HEALPix interchange guide. Overall this is accurate and well-scoped — I verified every claimed signature and worked example against `mortie/tools.py` and the section refs against `docs/specification.md`, and they hold up:

- API table is correct: `mort2healpix` (scalar → `(int, int)`, array → `(ndarray, order)`), `mort2norm` → `(normed, base_cell, order)`, `norm2mort(normed, parent, order)` → scalar `uint64`, `geo2mort(lat, lon)` → length-1 array for scalar input. The `[0]` indexing convention is right.
- The "raises on mixed orders" claim for `mort2healpix` is correct — `mort2norm` raises first (the array-path order branch in `mort2healpix` is effectively unreachable for mixed orders), so the guide describes real behavior.
- Arithmetic checks out: `37010 == 9*4096 + 146` (order 6), the `int64` wrap `-6881500230622117885`, prefix nibble `base+1`, and bit-63 set for base cells 7–11. Operator precedence in `int(w) >> 63 & 1` parses as intended.
- Argument-order pitfall (`geo2mort` = (lat, lon) vs cds/healpy = (lon, lat)) is called out; `healpy` keying on `nside` and `nest=True`/`lonlat=True` is correct.
- healpy being outside the extras is clearly disclosed; cdshealpix correctly identified as the `test`-extra oracle. Relative doc links and README link resolve correctly. Style/tables match `arrow_interchange.md`.

Only minor nits, none blocking (COMMENT, not request-changes) — three inline: (1) `repr()` expected-outputs are numpy-2-only; (2) confirm the pinned numeric outputs (esp. the lon `# 0.0`) were captured from a real run so they don't drift; (3) a precision nit on "southern base cells" vs the bit-63 set 7–11.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of the HEALPix-interchange guide (issue #63). Verified every snippet against `mortie/tools.py`, `mortie/__init__.py`, `pyproject.toml`, and the test suite. **No blocking defects — the API names, signatures, and gotchas are all correct.** Notes below.

**Verified correct**
- All referenced functions are real and exported: `geo2mort`, `mort2healpix`, `mort2norm`, `norm2mort`, `infer_order_from_morton`, `mort2geo`. Argument orders match: `geo2mort(lat, lon, order=)` is lat-first, `mort2norm` → `(normed, parent, order)`, `norm2mort(normed, base_cell, order)`.
- `docs/healpix_interchange.md:89` / `:132` — `mlat, mlon = mortie.mort2geo(words)` correctly handles mortie's unusual lat-first return order (`mort2geo` returns `lat, lon`). Easy to get wrong; got it right.
- The scalar worked example (order 6 → `cell 37010`, word `11570310392668225542`, base 9, normed 146) matches the `mort2healpix` docstring in `tools.py` and is internally consistent (`9*4096+146 == 37010`).
- The cdshealpix block reuses the exact order-14 lat/lon arrays and `lonlat_to_healpix(lons*u.deg, lats*u.deg, depth=order)` oracle call from `test_morton_index.py::test_latlon_matches_cdshealpix_nested` — a faithful mirror of the test-suite oracle pattern.
- Extras claims accurate: `cdshealpix>=0.7` is in the `test` extra; `healpy` is in no extra (guide says install separately). numpy-only runtime respected (§7).
- healpy calls correct: `ang2pix(nside, lons, lats, nest=True, lonlat=True)` and `pix2ang(..., lonlat=True)` both use the `nside`/lonlat convention properly.
- Gotcha 1 arithmetic checks out: prefix nibble `= base+1` (`int(w)>>60 == 10` for base 9), bit 63 set, and `np.int64(w) == -6881500230622117885` (= `w - 2**64`) is exactly the signed-misread value shown. Gotchas 2 (order self-encoding, mixed-order raise via `mort2norm`) and 3 (bare call → order-29 point) are accurate.

**Minor (non-blocking)**
- `docs/healpix_interchange.md:103` — the "Importing a foreign nested id" block uses `oracle`, `order`, and `nside_sq` from the *previous* code block, so it only runs as a continuation. A one-line "(continues the previous snippet)" note, or inlining `order = 14`, would keep it copy-run-safe.
- Wiring: there is no docs index/toctree in the repo (`docs/` is flat markdown linked ad hoc from README), so linking the guide from README is the right pattern. Note `docs/arrow_interchange.md` is not linked from README either — out of scope here, just flagging the inconsistency.
- The exact array cell-ids and the `~1.98e-09` / `0.0` tolerance comments couldn't be re-run here (review-only, no build), but they mirror the test inputs and the shared cds-healpix core, so they're plausible; worth a final `pytest`/manual run before un-drafting.


---

espg:
this doc mentions cdshealpix as verification, but doesn't mention or track https://github.com/matt-cornell/healpix-rs -- which is the actual crate we use in production for transforms to/from NESTED and other healpix structures and wrapped code functions. please update and add.

espg:
🤖 *from Claude*

Folded the three non-blocking self-review nits in `a35253c`:

(1) The two `print(repr(...))` demonstrations (the `geo2mort(...)[0]` word print in the conversion-API block and in gotcha 1) were numpy-2-only — `repr(np.uint64(x))` prints `np.uint64(x)` on numpy 2.x but the bare integer on numpy 1.x. Switched both to `print(int(...))` and updated the expected-output comments to the bare integer, so they're version-stable. The scalar `mort2healpix` / `mort2norm` prints were already stable and were left as-is.

(2) Added a one-line note after the cdshealpix round-trip block that the `cell_ids` list and the tolerance figures are captured from the current build and should be regenerated by rerunning the snippet rather than hand-edited. Numbers unchanged.

(3) Reworded the gotcha-1 sentence from "reinterpreting it as signed `int64` reads southern base cells as negative" to "makes base cells 7–11 read back negative" — the int64-negative set is exactly base cells 7–11 (bit 63 set), and base cell 7 is equatorial, so the "southern" label there was imprecise (matches specification.md §1 wording).

Docs-only, no build needed. Leaving the review threads for the reviewer to resolve.

---

espg:
🤖 *from Claude*

Good catch — the earlier run only folded the self-review nits and missed this. Addressed in `0f7d602`: the guide now names and tracks [`matt-cornell/healpix-rs`](https://github.com/matt-cornell/healpix-rs) (the `healpix` crate, pinned in `Cargo.toml`) as the crate mortie actually wraps for the NESTED transforms in production, and clarifies that `cdshealpix` / `healpy` appear here **only as independent external oracles** for cross-checking exactness, not as the implementation. Docs-only, no build needed.

---

espg:
🤖 *from Claude*

Tacked the main-greening fix into this PR as you asked. The red on main ([run 29847408029](https://github.com/espg/mortie/actions/runs/29847408029/job/88691409528)) was `test_mixed_order.py::TestResolutionCrossCheck::test_orders_of_indexes_order2res` — a pre-existing bug introduced with the `order2res` rewrite in #119, not by anything in this PR (it was surfaced by the #65 run). This branch was 9 commits behind main and still carried the old `order2res`, so the fix had to land against current main.

Two commits added:

1. **`eeebd7b` merge main** — brought this branch up to current `main` (no conflicts; the docs file here is new, main didn't touch it). No rebase/force-push.
2. **`c4d598e`** — `order2res` now accepts array input and exponentiates in float:

```python
order = np.asarray(order, dtype=np.float64)
area = 4 * np.pi * EARTH_RADIUS_KM**2 / (12 * 4.0**order)  # km2
res = np.sqrt(area)
return float(res) if res.ndim == 0 else res
```

Two bugs, one fix: (a) `float(np.sqrt(area))` threw `TypeError: only 0-dimensional arrays can be converted to Python scalars` on the array `orders_of(words)` passes; (b) `4**order` on a `uint8` order array wrapped `4**29` to 0 (the `divide by zero encountered in divide` RuntimeWarning in the same run). Float exponentiation of integer orders ≤29 is exact (powers of two), so scalar results are byte-identical — `res2display` and the existing `test_tools` order2res assertions are unaffected. The failing test is itself the coverage for the array path.

**Green locally** (numpy 2.4.6): `pytest -q` → **709 passed, 11 skipped, 0 failed** (was 708 + 1); the targeted test passes under `-W error::RuntimeWarning` (no div-by-zero warning); `flake8 mortie --select=E9,F63,F7,F82` clean. No Rust changed. CI should now go green on the `synchronize`.

Kept as draft — it's yours to mark ready and merge whenever CI confirms. Left `waiting`.

---

Issue #136, opened 2026-07-25
UNIQ helpers: multi-resolution support, and uniq2geo silently trusts a passed order


espg:
Filed on espg's direction from the #68 freeze audit ([PR #130 thread](https://github.com/espg/mortie/pull/130#issuecomment-5080621272)). Kept off PR #130 because that PR is a 4-line docstring diff and this is a behavior change needing its own tests.

Three related problems in the UNIQ helper family, all verified against `main` @ `ba2782a`.

## (1) `uniq2geo` trusts a caller-supplied order — silent wrong answers

`mortie/tools.py:437`:

```python
def uniq2geo(uniq, order=18):
    nside = 2**order
    nest = uniq - 4 * (nside**2)      # <-- order comes from the argument
    lon, lat = hp.pix2ang(order, nest)
    return lat, lon
```

**UNIQ is self-describing** — the order is recoverable from the value itself, which is exactly what `unique2parent` does 350 lines earlier (`orders = np.log2(np.array(unique)/4.0)//2.0`, `tools.py:82`). Because `uniq2geo` takes the order as a parameter and never cross-checks it, passing the wrong one returns **wrong coordinates with no error raised**.

This is the highest-severity item here: it is a correctness trap rather than a missing feature, and it is reachable from ordinary use since the parameter defaults (see (3)).

It is also *why* the function cannot do multi-resolution: a scalar `order` argument forces one resolution across the whole array.

**Fix:** decode the order per element from the UNIQ value. The `order` parameter then becomes redundant — either drop it (breaking) or keep it as an optional assertion that raises on mismatch (compatible). Decoding per element makes the function multi-res for free.

## (2) `unique2parent` throws away information it already computed

`mortie/tools.py:76-92`:

```python
orders = np.log2(np.array(unique)/4.0)//2.0
# this is such an ugly hack-- does little, will blow up with multi res
orders_ = np.unique(orders)
if len(orders_) == 1:
    order = int(orders_[0])
else:
    raise NotImplementedError("Cannot parse mixed resolution unique cells")
```

`orders` is **already per-element**. It is then collapsed to a scalar and the function refuses mixed input. Keeping it per-element and doing the remaining arithmetic elementwise (`4**(orders-1)` broadcasts fine in numpy) makes this multi-res in a handful of lines.

There is precedent for the pattern: #116 added group-by-order dispatch to `mort2geo` / `mort2polygon` / `mort2bbox` for exactly this situation.

[...]

PR #137, opened 2026-07-25
Docs site (mkdocs-material + mkdocstrings) and build workflow (issue #133)

### LINKED ISSUE #133: need docs workflow to build docs
...in general, this should match what we do in https://github.com/englacial/zagg and https://github.com/espg/moczarr

espg:
Closes #133.

Stands up a documentation site for mortie — **mkdocs + mkdocs-material + mkdocstrings**, built with **uv**, published to GitHub Pages — plus the one deviation neither reference repo needs: mortie is a compiled Rust extension, so the docs build has to compile `_rustie` before mkdocstrings can introspect the package.

Versioned docs are handled by **mike**, per the maintainer's decision. That changes the publishing model from the one originally scoped (see "Publishing model" below).

All seven open questions were answered in [this comment](https://github.com/espg/mortie/issues/133#issuecomment-5080653511); this PR is execution against those answers.

## Phases

- [x] **Phase 1 — scaffold.** `mkdocs.yml`, `[dependency-groups] docs`, the `docs`→`examples` extra rename, `docs/index.md`, nav over the six existing pages, `site/` ignored, `--strict` link audit.
- [x] **Phase 2 — API reference.** Seven mkdocstrings pages under `docs/api/`. Partial on purpose — see (c) below.
- [x] **Phase 3 — mike publishing (config landed; workflow needs hand-application).** `mike` added to the docs group, `extra.version.provider: mike` wired in `mkdocs.yml`, `strict: true` enforced at config level, `uv.lock` ignored. `.github/workflows/docs.yml` is written and verified but **cannot be pushed by the routine** — posted in [this comment](https://github.com/espg/mortie/pull/137#issuecomment-5080789992) to apply by hand.
- [x] **Phase 4 — folded into phase 3.** There is no separate artifact→branch conversion step any more, because this goes branch-model from the start. See below.
- [ ] **Phase 5 — notebooks.** Not implemented. Still gated on #108 Group B.

Adversarial self-review findings and the two fixes folded in are in [this comment](https://github.com/espg/mortie/pull/137#issuecomment-5080826065).

## Publishing model: mike, not the Pages artifact

Phase 3 was originally scoped to mirror moczarr — build to `site/`, publish via `actions/upload-pages-artifact` + `actions/deploy-pages`, Pages source "GitHub Actions". **That is no longer the plan.** mike publishes differently: it commits the rendered HTML to a `gh-pages` branch and Pages serves that branch. The two models are mutually exclusive — opposite Pages sources, opposite permissions — so this is a replacement, not an increment. The earlier YAML in the thread is superseded and **must not be applied**; it is explicitly marked as such.

Folding old phase 4 into phase 3 follows directly: converting from artifact-model to branch-model would have meant standing up a publishing path and immediately dismantling it.

Concretely, versus the artifact model:

| | artifact model (dropped) | mike branch model (this PR) |
|---|---|---|
| Permissions | `pages: write`, `id-token: write` | **`contents: write`** (mike pushes a branch) |
| Publishes to | Pages artifact | `gh-pages` branch |
| Pages source | "GitHub Actions" | **"Deploy from a branch" → `gh-pages`** |
| Versioning | none | `dev` on `main`; `X.Y.Z` + `latest` on tags |

The `contents: write` change is the security-relevant one: the deploy job gets write access to repo contents, which the artifact model did not need. It is scoped to the `deploy` job only — the workflow's top-level `permissions` stays `contents: read`, and the PR build job runs unprivileged.

`gh-pages` already exists as an orphan commit (`209cfa4`) carrying `.nojekyll` and a placeholder `index.html`. The `.nojekyll` matters because mkdocstrings emits `assets/_mkdocstrings.css` (confirmed present in the built site) and Jekyll skips underscore-prefixed paths — but mike writes its own `.nojekyll` on every deploy (`commands.py` ends with `commit.add_file(make_nojekyll())`), so the pre-seeded copy only covers the window before the first deploy. Nothing in this PR touches that branch.

**No tag is created or pushed by any of this.** A `*.*.*` tag triggers the PyPI publish in `build-wheels.yml`; per CLAUDE.md §1/§7 the routine never creates one. The workflow only ever *reacts* to a tag a human cuts.

## Phase 3 changes in this branch

```toml
[dependency-groups]
docs = [
  ...
  "mike>=2.1.3,<3",
]
```

```yaml
strict: true

extra:
  version:
    provider: mike
```

`site_url` was already set (`mkdocs.yml:2`), which mkdocs-material needs for versioning to resolve. `[project.dependencies]` is untouched — numpy remains the only runtime dependency, and PEP 735 groups are never published with the wheel, so an installed mortie is unaffected. `uv.lock` is gitignored, matching moczarr; CI resolves the docs group fresh.

`strict: true` is set in config rather than relying on `mkdocs build --strict` because **mike shells out to a plain `mkdocs build`** — without it, the published build would skip the exact check the PR gate enforces. The upper bound on mike exists for the same reason: with the lock ignored and `deploy` never running on a PR, a breaking major bump would first surface as a broken publish on `main`.

**This is not the inert configuration zagg has.** As flagged on [the issue](https://github.com/espg/mortie/issues/133#issuecomment-5080662823), zagg carries a dangling `provider: mike` with mike absent from its pyproject and every workflow — a version selector nothing populates. Here mike is a real dependency and the workflow deploys with it. Verified in the built HTML:

```json
{"translations": {"select.version": "Select version"}, "version": {"provider": "mike"}}
```

The dropdown itself is rendered client-side from `versions.json`, so it stays hidden until the first deploy — correct, not broken.

## Phase 1 — scaffold

`mkdocs.yml` uses the material theme with the light/dark palette toggle, `search`, and the mkdocstrings python handler with `docstring_style: numpy` as ratified. Nav covers `index.md` plus the six pages that already existed in `docs/`.

The one substantive edit to existing content was a link fix for `--strict`. `docs/benchmarks.md` linked `../BUILDING.md` — a relative link escaping `docs_dir`, which mkdocs cannot resolve and which `--strict` turns into a build failure. Repointed at the canonical blob URL rather than weakening `--strict`.

## The `docs`→`examples` rename

`[project.optional-dependencies] docs` held notebook *runtime* deps (matplotlib, geopandas, cartopy, earthaccess…), and the `docs` name was needed for the toolchain group. Renamed to `examples`, per the ratified answer.

**The rename turned out to be nearly free.** A repo-wide grep for `[docs]` / `mortie[docs]` — across markdown, workflows, notebooks, `binder/`, `environment.yml`, and Python — found **exactly one occurrence: the definition itself in `pyproject.toml`.** Nothing consumed it:

- `binder/environment.yml` lists the notebook deps as conda packages directly and `binder/postBuild` installs `mortie==0.8.2` from PyPI — it never touches the extra.
- No workflow installs it (`test.yml` uses `[test]`, `codspeed.yml` uses `[bench]`).
- `README.md` and `BUILDING.md` document `pip install mortie` and the maturin dev build; neither mentions the extra.

So the only external breakage is a third party running `pip install mortie[docs]`, which is exactly the pre-freeze cost that was accepted. The four protected workflows (`test.yml`, `lint.yml`, `codspeed.yml`, `build-wheels.yml`) are untouched.

## Phase 2 — API reference

Seven pages under `docs/api/`, one per module (`# mortie.` + a `:::` block): `tools`, `coverage`, `prefix_trie`, `geometry`, `linestring`, `morton_index`, `arrow`. **47 objects render in total.**

Each page uses an explicit `members:` list rather than a bare `::: mortie.`. That is deliberate: it is the only way to hold back names whose freeze verdict is still open, and it makes the published surface reviewable as a diff rather than as whatever the module namespace happens to contain.

Two structural notes:

1. **Lazily-built classes are not in the API pages.** `MortonIndexDtype` / `MortonIndexArray` (`morton_index.py:203` `_build_classes`) and `MortonIndexType` / `MortonIndexExtArray` (`arrow.py:54` `_build_type`) are constructed inside functions and exposed through a module `__getattr__`, so a numpy-only install can still import `mortie`. mkdocstrings uses griffe's static analysis and cannot see through that. They remain documented narratively in `morton_index_datatype.md` and `arrow_interchange.md`, both in the nav, and each API page says so and links across.
2. **`mortie.arrow` includes `export_c_array` / `export_c_schema` / `import_c_array`.** These are not in `mortie.__all__` (they are reached as `mortie.arrow.*`), but they are the documented C Data Interface surface from #93 and are already described in `arrow_interchange.md`, so holding pages back from them seemed worse than including them. Flagging in case you disagree.

## How tested

Merged `origin/main` into this branch first (merge, not rebase — the branch is already pushed and §1 forbids force-pushing), picking up `5134984 enforcing numpydocs`. Clean merge, one auto-merge in `pyproject.toml`.

Clean venv, then the uv docs path:

```
$ flake8 mortie --select=E9,F63,F7,F82
(clean)

$ pytest -q
757 passed, 12 skipped, 5 warnings in 14.98s

$ uv sync --group docs          # builds the Rust ext via the maturin backend
$ uv run python -c "import mortie; from mortie import _rustie"
mortie 0.9.1 | rust ext OK

$ uv run mike --version
mike 2.2.0

$ uv run mike list              # exit 0, empty -- nothing deployed yet

$ uv run mkdocs build           # no --strict flag; strict comes from config
INFO    -  Documentation built in 1.47 seconds
```

757 / 12 matches the pre-change baseline exactly. The build is green with **no warnings suppressed and no validation relaxed**. No Rust changed, so the `cargo` gates are not exercised by this diff.

CI on `162a21e` was all green — 15 success, 6 skipped, 0 failures, including `ruff` under the numpydoc `D` rules newly landed on `main`. The CodSpeed check that was red on the previous commit went green on its own; it was the stale-comparison-base artifact flagged earlier, not a real regression.

**`uv sync --group docs` builds the Rust extension on its own** — the part neither reference repo has to solve. Because the project is its own dependency, uv drives the maturin build backend and pulls maturin in as a PEP 517 build requirement, so the workflow needs only `dtolnay/rust-toolchain@stable` on PATH and no separate maturin step. Expect the deploy leg to be slower than a pure-Python docs job, since it compiles the extension from scratch; no cargo caching is included, on the view that it is better added if the time actually bites.

mike was exercised **read-only** (`--version`, `list`). No deploy was run and `gh-pages` was not modified.

## Questions for review

**(a) Repo-admin action required before docs will publish.** Set **Settings → Pages → Source: "Deploy from a branch" → `gh-pages` / `(root)`**. Do **not** choose "GitHub Actions" — that is the artifact model, and it is incompatible with mike. This is the opposite of the setting the original phase-3 scope called for, which is why it is called out rather than assumed.

**(b) `.github/workflows/docs.yml` needs hand-application.** The routine's token lacks `workflow` scope and the push is rejected; the full file is in [this comment](https://github.com/espg/mortie/pull/137#issuecomment-5080789992), with one recommended one-line tweak (`--alias-type redirect`) noted [here](https://github.com/espg/mortie/pull/137#issuecomment-5080826065). Either apply it yourself or grant the routine `workflow` scope and a later run will push it. I did not write it through the GitHub API — that would bypass the control rather than satisfy it.

**(c) The API reference is deliberately incomplete, pending PR #130.** Still held: `heal_norm` (being deleted under #130) and the UNIQ helpers `geo2uniq` / `norm2uniq` / `uniq2geo` / `unique2parent` (signatures in flux under #136). `res2display` and `MortonChild` are now documented following your verdicts. Each remaining one is a one-line addition to a `members:` list once #130 resolves.

**(d) Site root right after the first deploy.** `mike deploy dev` writes `/dev/` and `versions.json` but does not touch the site root, so the root will still serve the placeholder `index.html` until the first release tag runs `mike set-default --push latest`. Run `mike set-default dev --push` once if you want the root live sooner. Flagged so a blank-looking root is not mistaken for a broken build.

**(e) `mike>=2.1.3,<3` is the one new dependency** — docs group only, never published with the wheel, no runtime impact. Pre-approved; noting it here for the record per §4.

**(f) Three judgment calls left standing from self-review**, detailed [here](https://github.com/espg/mortie/pull/137#issuecomment-5080826065): `set-default latest` firing on every tag (a backport tag would demote the root), no manual redeploy lever, and the mike alias type. Each is a policy choice rather than a bug.


espg:
🤖 *from Claude*

Phase 3's workflow, for the record and for whoever applies it. It is verified locally but is **not** in the branch — the push was rejected because the routine's token lacks `workflow` scope:

```
 ! [remote rejected] claude/133-docs-workflow -> claude/133-docs-workflow
   (refusing to allow an OAuth App to create or update workflow
    `.github/workflows/docs.yml` without `workflow` scope)
```

I stopped there per CLAUDE.md §2 rather than retrying, and deliberately did not write the file through the GitHub API instead — the scope restriction exists precisely to keep automation out of workflow files, so routing around it would defeat the control rather than satisfy it. Pasting the content for a human to apply is the honest version of the same outcome.

`.github/workflows/docs.yml`:

```yaml
name: Build and Deploy Docs

on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
  workflow_dispatch:

permissions:
  contents: read
  pages: write
  id-token: write

concurrency:
  group: "pages"
  cancel-in-progress: false

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v5

    # mortie is a compiled Rust extension, so unlike the moczarr docs job the
    # package has to be built before mkdocstrings can introspect it. `uv sync`
    # below installs the project itself, which drives the maturin build backend
    # and needs a Rust toolchain on PATH.
    - name: Install Rust toolchain
      uses: dtolnay/rust-toolchain@stable

    - name: Install uv
      uses: astral-sh/setup-uv@v6

    - name: Set up Python
      run: uv python install 3.12

    - name: Install dependencies
      run: uv sync --group docs

    - name: Build docs
      run: uv run mkdocs build --strict

    - name: Setup Pages
      uses: actions/configure-pages@v5
      if: github.ref == 'refs/heads/main'

    - name: Upload pages artifact
      uses: actions/upload-pages-artifact@v3
      if: github.ref == 'refs/heads/main'
      with:
        path: site

  deploy:
    environment:
      name: github-pages
      url: ${{ steps.deployment.outputs.page_url }}
    runs-on: ubuntu-latest
    needs: build
    if: github.ref == 'refs/heads/main'
    steps:
    - name: Deploy to GitHub Pages
      id: deployment
      uses: actions/deploy-pages@v4
```

It also wants this in `.gitignore`, matching moczarr (which ignores the lock at `.gitignore:16`):

```
# uv resolves the docs group fresh in CI; the lock is not committed (matches
# the moczarr docs setup this workflow mirrors).
uv.lock
```

Two things worth knowing before it runs:

**The Rust step is load-bearing, and there is no separate maturin step on purpose.** `uv sync --group docs` installs the project itself, which drives the maturin build backend; maturin arrives automatically as a PEP 517 build requirement. All the job has to supply is a toolchain. Verified locally:

```
$ uv sync --group docs
Resolved 210 packages in 2.92s
   Building mortie @ file:///.../mortie
      Built mortie @ file:///.../mortie
Prepared 1 package in 27.56s          # maturin==1.14.1 pulled as a build dep

$ uv run python -c "import mortie; from mortie import _rustie; print('rust ext OK')"
mortie 0.9.1
rust ext OK

$ uv run mkdocs build --strict
INFO    -  Documentation built in 1.42 seconds
```

**Expect the build leg to be slower than moczarr's** — it compiles the Rust extension from scratch on every run (~30 s locally for the maturin build alone, likely a few minutes on a cold CI runner). No cargo caching step is included, since neither reference has one and it seemed better to add it if the time actually bites than to speculate now.

Deployment only happens on `main`: `configure-pages` and `upload-pages-artifact` are both gated on `github.ref`, and the `deploy` job is too, so pull requests build-and-check without publishing.


---

espg:
🤖 *from Claude*

CI status on `a02db6b` (phases 1–2): **20 of 21 checks green** — `test (3.10)`, `test (3.11)`, `test (3.12)`, `arro3-no-pyarrow`, `ruff`, both codecov checks, all six wheel builds, and both benchmark jobs.

The one red check is **CodSpeed Performance Analysis**, reporting `test_coverage_triangle_order4` at −11.81% (343.8 µs → 389.9 µs), 66 benchmarks untouched. **Flagging rather than chasing it, per CLAUDE.md §4** — I do not believe it is attributable to this PR, and CodSpeed's own report says as much twice:

> :warning: **Different runtime environments detected** — Some benchmarks with significant performance changes were compared across different runtime environments, which may affect the accuracy of the results.

> No successful run was found on `main` (7def571) during the generation of this report, so ba2782a was used instead as the comparison base. There might be some changes unrelated to this pull request in this report.

So it is a cross-environment comparison against a stale base — `ba2782a` rather than the actual merge base `7def571`.

The mechanical argument is the stronger one: **this PR changes no Python and no Rust.** The full diff at that commit is seven new markdown files under `docs/api/`, `docs/index.md`, `mkdocs.yml`, a one-line URL fix in `docs/benchmarks.md`, two `.gitignore` lines, and the `pyproject.toml` extra rename plus the new `[dependency-groups]`. Nothing on the coverage path is touched, and `Python benchmarks` / `Rust benchmarks` both passed — it is only the CodSpeed comparison that is red. Re-running it against a green `main` should clear it.

Noting for the record that I did **not** take CodSpeed's suggested remediation action (it proposes commenting a bot directive to auto-fix the regression). Acting on it would mean an unrequested code change driven by a non-`@espg` bot comment, which §6 reserves for you.

The ruff check is green, and there are no ruff bot review comments outstanding on this PR.


---

espg:
🤖 *from Claude*

Follow-up commit `c68c88a` — the API reference held back three name groups pending PR #130's verdicts, and two of those have since been answered, so they're now documented.

**`MortonChild` is in** (`docs/api/prefix_trie.md`). You approved documenting it on [PR #130](https://github.com/espg/mortie/pull/130#issuecomment-5080621272) after this branch's scope was set, so the omission was stale rather than deliberate. It's documented as a **return type**, matching the ratified shape: read surface (`characteristic`, `len`, `children`, `nchildren`, `mantissa_array`, `cell_area`) is the frozen contract, constructor internal and not part of it. Verified rendering — 30 references on the built page, both properties present.

**`res2display` is in** (`docs/api/tools.md`). Its verdict is settled too (return structured data, move the demo to a docs example). Note the rendered content will change when PR #130 merges: on this branch it still documents the old printing behavior, since #130's phase 2 reshaped it to return `list[ResolutionLevel]` with a numpydoc `Examples` block. Self-correcting — mkdocstrings reads whatever the code says at build time.

**Still held, correctly:** `heal_norm` (being deleted under #130) and the UNIQ helpers `geo2uniq` / `norm2uniq` / `uniq2geo` / `unique2parent` (signatures in flux under #136 — `uniq2geo` loses its `order` parameter, the other two make it required). Added an explicit note on the tools page saying so with links, rather than leaving a silent gap.

`mkdocs build --strict` green, nothing weakened. Confirmed by grepping the built HTML that the omitted names produce no anchors.

---

espg:
🤖 *from Claude*

## `.github/workflows/docs.yml` — mike branch model (replaces the earlier YAML)

⚠️ **This supersedes the workflow posted in [this earlier comment](https://github.com/espg/mortie/pull/137#issuecomment-5080704792). Do not apply that one.** It was written for the `upload-pages-artifact` + `deploy-pages` model, which is **mutually exclusive** with mike — the two want opposite Pages sources and opposite permissions. Applying both, or the wrong one, breaks publishing.

Still cannot be pushed by the routine; the wall reproduces on this branch verbatim:

```
 ! [remote rejected] claude/133-docs-workflow -> claude/133-docs-workflow
   (refusing to allow an OAuth App to create or update workflow
    `.github/workflows/docs.yml` without `workflow` scope)
```

I stopped there per CLAUDE.md §2 rather than retrying, and did not write it through the GitHub API — the restriction exists to keep automation out of workflow files, so routing around it would defeat the control rather than satisfy it. The local commit was backed out; the branch matches the remote exactly.

```yaml
name: Docs

on:
  push:
    branches: [main]
    tags:
      - '*.*.*'
  pull_request:
    branches: [main]
  workflow_dispatch:

permissions:
  contents: read

jobs:
  # Pull requests (and manual runs) verify the docs build only -- nothing is
  # published. The rendered site is uploaded so a reviewer can download it.
  build:
    name: Build (verify only)
    if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
    runs-on: ubuntu-latest

    steps:
    - name: Checkout code
      uses: actions/checkout@v4

    # mortie is a compiled Rust extension, so the package has to be built before
    # mkdocstrings can introspect it. `uv sync` installs the project itself,
    # which drives the maturin build backend and pulls maturin in as a PEP 517
    # build requirement -- so all this job supplies is a toolchain on PATH.
    # There is deliberately no separate maturin step.
    - name: Install Rust toolchain
      uses: dtolnay/rust-toolchain@stable

    - name: Install uv
      uses: astral-sh/setup-uv@v6

    - name: Set up Python
      run: uv python install 3.12

    - name: Install dependencies
      run: uv sync --group docs

    - name: Build docs
      run: uv run mkdocs build --strict

    - name: Upload built site
      uses: actions/upload-artifact@v4
      with:
        name: docs-site
        path: site/

  # Pushes publish with mike, which commits the rendered site to the `gh-pages`
  # branch -- hence `contents: write`, rather than the `pages: write` /
  # `id-token: write` an upload-pages-artifact deploy would need. Pages must be
  # configured to serve the `gh-pages` branch, NOT "GitHub Actions".
  deploy:
    name: Deploy (mike)
    if: github.event_name == 'push'
    runs-on: ubuntu-latest
    permissions:
      contents: write
    # Serialize deploys -- concurrent `mike deploy --push` runs would race on
    # the gh-pages branch. Never cancel one mid-push.
    concurrency:
      group: docs-deploy
      cancel-in-progress: false

    steps:
    - name: Checkout code
      uses: actions/checkout@v4
      with:
        fetch-depth: 0  # mike needs the gh-pages branch, not just this ref

    - name: Install Rust toolchain
      uses: dtolnay/rust-toolchain@stable

    - name: Install uv
      uses: astral-sh/setup-uv@v6

    - name: Set up Python
      run: uv python install 3.12

    - name: Install dependencies
      run: uv sync --group docs

    - name: Configure git identity for mike
      run: |
        git config user.name 'github-actions[bot]'
        git config user.email '41898282+github-actions[bot]@users.noreply.github.com'

    - name: Deploy dev docs
      if: github.ref == 'refs/heads/main'
      run: uv run mike deploy --push dev

    # A *.*.* tag is cut by a human as part of a release (the same tag drives
    # the PyPI publish in build-wheels.yml). This job only ever *reacts* to a
    # tag -- it never creates or pushes one.
    - name: Deploy release docs
      if: startsWith(github.ref, 'refs/tags/')
      run: |
        uv run mike deploy --push --update-aliases "$GITHUB_REF_NAME" latest
        uv run mike set-default --push latest
```

### What changed against the earlier version, and why

**(1) `contents: write`, not `pages: write` / `id-token: write`.** mike publishes by committing the rendered site to `gh-pages` and pushing it, so it needs write access to repo contents and none of the Pages OIDC machinery. Top-level stays `contents: read`; only the `deploy` job is elevated.

**(2) `fetch-depth: 0` on the deploy checkout.** mike reads and rewrites the `gh-pages` branch, so a shallow single-ref clone is not enough.

**(3) Git identity is configured before mike runs** — mike makes a real commit, and the runner has no default identity.

**(4) PRs verify only.** The build job runs `mkdocs build --strict` and uploads `site/` as the `docs-site` artifact for download. It has no deploy path at all.

**(5) Deploys are serialized** via a job-level `docs-deploy` concurrency group with `cancel-in-progress: false`. Two concurrent `mike deploy --push` runs would race on `gh-pages`, and cancelling one mid-push could leave the branch inconsistent. PR builds are unaffected and still run in parallel.

**(6) `actions/checkout@v4`** to match `test.yml`, `lint.yml`, and `codspeed.yml`, rather than the `@v5` in the earlier draft.

### Verified locally

`gh-pages` (`209cfa4`) is confirmed as a parentless orphan carrying exactly `.nojekyll` and `index.html`. Nothing in this run touched it — no deploy was run, and mike was only exercised read-only.

```
$ uv sync --group docs          # builds the Rust ext via the maturin backend
$ uv run python -c "import mortie; from mortie import _rustie"
mortie 0.9.1 | rust ext OK

$ uv run mike --version
mike 2.2.0

$ uv run mike list              # exit 0, empty -- no versions deployed yet
$ uv run mkdocs build --strict
INFO    -  Documentation built in 1.75 seconds
```

`--strict` is green with nothing weakened. Full suite unchanged at **757 passed, 12 skipped**, and `flake8 mortie --select=E9,F63,F7,F82` is clean.

The version selector is genuinely wired this time, not inert as in the zagg reference. The built HTML carries the provider in the theme config block:

```json
{"translations": {"select.version": "Select version"}, "version": {"provider": "mike"}}
```

plus the `data-md-component="outdated"` element. The visible dropdown is rendered client-side from `versions.json`, which only exists once mike has deployed — so it stays hidden until the workflow runs, which is correct rather than broken.

### One operational wrinkle worth knowing before the first deploy

`mike deploy dev` writes `/dev/` and `versions.json` but **does not touch the site root**. So after the first push to `main`, `https://espg.github.io/mortie/` will still serve the existing placeholder `index.html`, while `https://espg.github.io/mortie/dev/` serves the real docs. It self-corrects at the first release tag, when `mike set-default --push latest` writes a root redirect.

If you would rather the root work immediately, run this once locally after the first `main` deploy:

```
uv run mike set-default dev --push
```

Either is fine — flagging it so a blank-looking root right after the first deploy is not mistaken for a broken build.


---

espg:
🤖 *from Claude (review)*

## Adversarial self-review — phase 3 (`162a21e` + merge `664a118`), including the `docs.yml` posted for hand-application

**No BLOCKING findings.** The three-file diff is correct, and I could not construct a push event that fails to deploy, deploys twice, or does anything unsafe with a `*.*.*` tag. Details of what I verified are at the bottom. Eight non-blocking items follow, roughly in order of value.

---

### NON-BLOCKING

**(1) The *published* build is not `--strict`, and nothing strict-verifies `main`.** This is the one I'd actually fix before merge.

The `build` job is gated:

```yaml
if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
```

so on a push to `main` the only job that runs is `deploy` — and `mike deploy` shells out to mkdocs without `--strict`:

```python
# mike/mkdocs_utils.py
def build(config_file, version, *, quiet=False, output=None):
    command = ([_program_name] + (['--quiet'] if quiet else []) +
               ['build', '--clean'] + ...)
```

So `--strict` gates the PR and nothing else. Anything that only breaks *after* merge — two PRs that each pass but together orphan a nav entry, a docstring edit on `main` that breaks an mkdocstrings identifier — publishes with warnings rather than failing. That is exactly the silent-regression mode the phase is meant to close.

One line in `mkdocs.yml` fixes it: `strict: true`. mike round-trips the entire config through `inject_plugin` (`yaml.dump(config, f, sort_keys=False)`), so a config-level `strict` survives into the temp config mike builds against; `strict` is a real `MkDocsConfig` key, confirmed. The `--strict` flag in the build job then becomes redundant but harmless.

**(2) `mike>=2.1.3` is unbounded and `uv.lock` is gitignored, so the deploy job resolves mike fresh every run.** Compounded by (1): a mike 3.x CLI change — or any breaking release across the ~210-package docs resolution — lands unannounced in the one job that holds `contents: write` and writes to `gh-pages`, and no PR would catch it, because `deploy` never runs on a PR. `Cargo.lock` is gitignored too, so both the Python and the Rust dependency sets float in that job.

The floor itself is sane: `deploy --push --update-aliases` and `set-default --push` all predate 2.1.3. But it is untested — verification was on 2.2.0. Cheap hardening: `"mike>=2.1.3,<3"`. Committing `uv.lock` is the bigger question and a real judgment call — it is what would make "green on the PR" mean the same thing at deploy time. Ignoring it matches moczarr, but moczarr's docs job isn't holding a write token.

**(3) `mike set-default --push latest` runs unconditionally on every tag.**

```yaml
uv run mike deploy --push --update-aliases "$GITHUB_REF_NAME" latest
uv run mike set-default --push latest
```

`--update-aliases` *moves* `latest` to whatever was just tagged, and `set-default` repoints the site root at it. A backport tag (`0.8.4` cut after `0.9.1`) silently demotes the site root to the older line. Separately, the `'*.*.*'` filter also matches pre-release forms (`1.0.0-rc1`, `0.9.2.post1`), which would likewise become `latest` and the default. The filter does match `build-wheels.yml:6-7` exactly, so it inherits the existing release convention rather than inventing one — but that convention governs wheels, where "latest" is PyPI's problem, not a root redirect this workflow writes. Fine to accept if releases stay linear; worth a sentence in the PR body either way.

**(4) The concurrency queue holds one run, so a release deploy can be silently dropped.** The `docs-deploy` group with `cancel-in-progress: false` is the right call and is genuinely load-bearing — two concurrent `mike deploy --push` runs would fail the second at `git_utils.push_branch` (`git push -- origin gh-pages`, non-fast-forward → `GitError`), so it fails loudly rather than corrupting, but it does fail. The residual: GitHub keeps only **one** pending run per group. Sequence `push main` (running) → `push tag` (pending) → `push main` (arrives) cancels the pending *tag* deploy. Release docs never publish, and the run reads as "cancelled", not failed.

Low likelihood, but there is currently no manual redeploy lever: `deploy` is `if: github.event_name == 'push'`, so `workflow_dispatch` reaches only the verify job. Widening that to include `workflow_dispatch` would give a recovery path and would also let the one-time `mike set-default dev` from item (d) of the PR body run from Actions rather than from a local clone with a Rust toolchain. Only the job-level `if` needs widening — `github.ref` is `refs/heads/main` for a dispatch from main, so the existing per-step gates already do the right thing.

**(5) `uv python install 3.12` doesn't pin the interpreter.** It installs a managed 3.12, then `uv sync` does its own discovery against `requires-python = ">=3.10"`. It works today because uv's default `python-preference` is `managed`, so it picks the one just installed — but the step named "Set up Python" doesn't actually select anything. `uv python pin 3.12`, or `uv sync --python 3.12`, makes it deterministic. Nit; carried over from the superseded YAML.

**(6) The verify build and the published build use different configs.** The build job runs bare `mkdocs build --strict`; mike builds through `inject_plugin` — a temp config with `mike` prepended to `plugins` and `site_url` rewritten to `.../dev` or `.../0.9.2` by `MikePlugin.on_config`. Declaring the plugin explicitly (material's documented setup) closes the gap and lets you set `canonical_version`:

```yaml
plugins:
  - mike:
      canonical_version: latest
```

Without it, `canonical_version` defaults to `None` and every release's canonical URL pins to its own version directory rather than converging on `latest`. I checked that declaring it is safe for plain `mkdocs build`/`serve`: `on_config` no-ops without `MIKE_DOCS_VERSION`, and `on_files` no-ops for material because `material` is not registered in mike's `mike.themes` entry point group (verified — the group carries only `mkdocs`, `readthedocs`, and the bootswatch skins). That same fact independently confirms the config as landed is right: mike deliberately injects nothing for material, so `extra.version.provider: mike` really is the entire wiring.

**(7) Alias type is symlink (mike's default) — a watch item for the first release, not a change to make now.** `commands.deploy` defaults to `alias_type=AliasType.symlink`, committing `latest` as a git symlink (`mode=0o120000`). GitHub Pages' branch build tars with `--dereference`, so this is the well-trodden path. If `latest/` 404s after the first tag, `alias_type: redirect` is the one-line fix.

**(8) Small accuracy note on the PR body.** It calls the pre-seeded `.nojekyll` on `gh-pages` load-bearing. True before the first deploy — but mike writes it itself on every deploy; `commands.deploy` ends with `commit.add_file(make_nojekyll())`. So the site does not depend on that orphan commit surviving, which is a slightly better story than the body tells.

---

### Verified correct (recording the reasoning, since several of these were non-obvious)

- **Trigger and gate exhaustiveness.** `push` reaches only `deploy`; `pull_request` and `workflow_dispatch` reach only `build`. Within one push event `refs/heads/main` and `refs/tags/*` are mutually exclusive, so no path double-deploys and none is unhandled. A tag on a commit already on `main` produces two *separate* push events and two runs — correctly serialized by the concurrency group, subject only to (4).
- **`fetch-depth: 0` is genuinely required, for the reason the comment gives.** mike's `check_remote_status` → `git_utils.update_from_upstream` does **not** fetch; it only does `git rev-parse origin/gh-pages`. `actions/checkout` at `fetch-depth: 0` fetches `+refs/heads/*:refs/remotes/origin/*`, so `origin/gh-pages` exists and mike fast-forwards a local `gh-pages` onto it. At the default depth of 1 it would not, and mike would start from an empty branch. Holds on both the main and the tag path.
- **Permissions.** Job-level `permissions` *replaces* the top level, so `deploy` gets `contents: write` and every other scope `none` — the minimum mike needs. Fork PRs reach only the unprivileged `build` job.
- **Nothing here can create or push a tag.** mike pushes exactly one ref: `git push -- origin gh-pages`.
- **Git identity** is configured after checkout and before mike, in the same workspace. Correct order.
- **First deploy onto the existing orphan branch.** mike's `Commit` fast-imports with `from <gh-pages head>`, so the placeholder `index.html` survives until `set-default` overwrites it with a redirect (`commands.set_default` writes root `index.html` from the redirect template). Item (d) of the PR body is accurate as written.
- **No-op deploys don't fail.** `handle_empty_commit()` converts `GitEmptyCommit` to a warning and skips the push, so a repeat `set-default latest` on an unchanged root is harmless.
- **`extra.version.provider: mike` placement is correct** and `site_url` is set (`mkdocs.yml:2`), which material requires. I specifically chased a suspected canonical-URL bug — `MikePlugin.on_config` does `urljoin(site_url, version)`, which drops the trailing slash (`https://espg.github.io/mortie/dev`) *after* mkdocs' `URL(is_dir=True)` validation has run — and it is a non-issue: `Page._set_canonical_url` re-appends the slash before joining.
- **The zagg failure mode is genuinely avoided.** mike is a real dependency, the deploy job runs it, and `versions.json` is written at the gh-pages root by every `commands.deploy`. The selector staying hidden until the first deploy is correct, as the body says.

---

espg:
🤖 *from Claude*

Follow-up commit `5786d25` — addressing the adversarial self-review above. Two findings folded in, both one-liners, both pushable (they live in config, not the workflow).

### (1) The *published* build was never strict-checked — fixed

This was the real find. `mkdocs build --strict` only ever ran in the PR job, and that job is gated `if: pull_request || workflow_dispatch`. A push to `main` runs **only** the deploy job, and mike does not shell out with `--strict` — `mike/commands.py` calls a plain `mkdocs build --clean`. So anything that broke only after merge would have published with warnings, silently, while the PR gate stayed green.

Fixed at the config level so it cannot be bypassed by whoever invokes the build:

```yaml
strict: true
```

Verified it is actually live in the parsed config rather than just present in the file:

```
$ uv run python -c "from mkdocs.config import load_config; print(load_config('mkdocs.yml').strict)"
True

$ uv run mkdocs build          # note: no --strict flag
INFO    -  Documentation built in 1.47 seconds
```

That second command is the one that matters — it is what mike runs, and it now enforces strict from config. Nothing was weakened to get there; the build was already clean.

### (2) `mike` upper bound — fixed

`mike>=2.1.3` → `mike>=2.1.3,<3`. With `uv.lock` gitignored the deploy job resolves fresh on every run, and because `deploy` never runs on a PR, a breaking mike 3.x would first surface as a broken publish on `main` with no PR to catch it. A major-version bound is the cheap guard.

### Left standing for your call

Not acted on — each implies a scope or policy decision rather than a fix, and per §6 those are yours:

**(a) `mike set-default --push latest` runs on every tag.** A backport tag (say `0.8.3` cut after `0.9.1`) would silently demote the site root to the older version. The `'*.*.*'` filter also matches pre-release forms. It mirrors `build-wheels.yml`'s existing tag filter, so it inherits a convention you already run — worth a conscious yes/no rather than a quiet change.

**(b) Recommended one-word change before you apply the workflow.** mike 2.x defaults to `--alias-type symlink`, which writes a real git symlink (mode `120000`) for the `latest` alias. Symlink aliasing is [documented as unsupported on some GitHub Pages deployments](https://github.com/squidfunk/mkdocs-material/discussions/6308), and mkdocs-material tracks the failure mode in [squidfunk/mkdocs-material#6550](https://github.com/squidfunk/mkdocs-material/issues/6550). It would only bite at the first release tag, and `redirect` works everywhere at no cost, so I would change the tag-deploy line to:

```yaml
        uv run mike deploy --push --update-aliases --alias-type redirect "$GITHUB_REF_NAME" latest
```

I did **not** rewrite the posted YAML for this, since it is a judgment call and you may prefer the default; the rest of [that comment](https://github.com/espg/mortie/pull/137#issuecomment-5080789992) stands as-is.

**(c) No manual redeploy lever.** `deploy` is gated `if: github.event_name == 'push'`, so `workflow_dispatch` can only ever verify, never republish. Deliberate (a manual deploy button is a foot-gun next to a `contents: write` job), but if you want one it is an `if:` change.

**(d) Concurrency queue depth is 1.** GitHub keeps one pending run per group, so `push main` → `push tag` → `push main` in quick succession would cancel the *pending* release deploy. The serialization itself is correct and load-bearing — a genuine race fails loudly at the push rather than corrupting `gh-pages`, since mike checks remote status before writing.

### Correction to the PR body

I overstated the pre-seeded `.nojekyll` as load-bearing. mike writes one itself on **every** deploy — `commands.py` ends with `commit.add_file(make_nojekyll())` — so the placeholder's copy only matters for the window before the first deploy. The underlying point is still real (mkdocstrings emits `assets/_mkdocstrings.css`, and Jekyll skips underscore-prefixed paths — confirmed present in the built site), but mike handles it without help. PR body updated.

CI on `162a21e` was **all green**: 15 success, 6 skipped, 0 failures across all three Python versions, `arro3-no-pyarrow`, `ruff`, all six wheel builds, and both codecov checks. The CodSpeed check that was red on the previous commit is now green on its own — it was the stale-comparison-base artifact flagged earlier, not a real regression.


---

PR #139, opened 2026-07-25
UNIQ helpers: multi-resolution support, and drop uniq2geo's untrusted order

### LINKED ISSUE #136: UNIQ helpers: multi-resolution support, and uniq2geo silently trusts a passed order

espg:
Closes #136

Makes the UNIQ helper family multi-resolution and removes the caller-supplied order that `uniq2geo` never cross-checked. The spec implemented here is the one ratified on the issue thread ([comment](https://github.com/espg/mortie/issues/136#issuecomment-5080801685), [settled table](https://github.com/espg/mortie/issues/136#issuecomment-5080820241)).

| function | role | change |
|---|---|---|
| `uniq2geo` | decoder | `order` parameter **deleted**; decoded per element from the value |
| `unique2parent` | decoder | multi-res via per-element orders; no signature change |
| `geo2uniq` | encoder | `order` accepts scalar **or** per-element array; default `MAX_ORDER` |
| `norm2uniq` | encoder | `order` accepts scalar **or** per-element array; default `MAX_ORDER` |

## Approach

UNIQ is self-describing — `uniq = 4 * 4**order + nested` with `0 <= nested < 12 * 4**order`, so order `k` occupies exactly `[4**(k+1), 4**(k+2))` and consecutive orders tile that line with no gaps. A new module-private `_uniq_orders` turns that into a per-element decode, and both decoders build on it:

```python
u = np.atleast_1d(np.asarray(uniq, dtype=np.int64))
bounds = np.int64(4) ** np.arange(1, MAX_ORDER + 3, dtype=np.int64)
orders = (np.searchsorted(bounds, u, side='right') - 1).astype(np.int64)
```

That is an exact integer bucket search, deliberately **not** the `log2(uniq / 4) // 2` form `unique2parent` used before. The old form goes through float64, which cannot separate adjacent order boundaries above 2**53: `4**30 - 1` is the last order-28 UNIQ value but rounds up to `4**30` in float64 and mis-decodes as order 29. `TestUniqOrders::test_order_boundaries_are_exact` pins every `k`/`k+1` boundary from 0 to 29.

Both decoders then reduce with shifts rather than division, staying in exact integers — `nested = uniq - (1 << (2*order + 2))`, `parent = nested >> (2*order)`.

For the encoders, `geo2uniq` calls a depth-scalar kernel (`hp.ang2pix`), so a per-element order needs group-by-order dispatch — the pattern issue #116 introduced in `mort2geo` / `mort2polygon` / `mort2bbox` (`9533022`):

```python
for one_order in np.unique(order):
    mask = order == one_order
    uniq[mask] = geo2uniq(lats[mask], lons[mask], int(one_order))
```

`norm2uniq` needs no dispatch at all: its body is integer arithmetic (`4 * nside**2 + normed + parent * nside**2`) and broadcasts elementwise against an array of orders as it stands. A shared `_encoder_orders` validates both — range `0..MAX_ORDER`, and one entry per input element for the array form.

## Phases — all complete

- [x] (1) `uniq2geo` — decode the order per element, delete the `order` parameter; update the one in-tree caller (`mort2geo`) and `test_mort_inverse.py` — `4b04583`
- [x] (2) `unique2parent` — per-element orders; the `NotImplementedError` and the "such an ugly hack" comment are gone — `111d8e9`
- [x] (3) `geo2uniq` / `norm2uniq` — scalar-or-array `order`, `MAX_ORDER` default, kind-less documentation — `cfb1b8e`
- [x] (4) round-trip and cross-function integration tests, `USAGE.md` — `5b90830`

## A finding worth recording: the trap is a panic, not a silent wrong answer

The issue describes the old `uniq2geo(uniq, order=18)` as returning "wrong coordinates with no error raised". Against the current Rust `healpix` backend that is not what happens, and the truth is worth pinning: **a mismatched order cannot produce a valid decode at all.**

Because the UNIQ ranges tile without overlap, `uniq - 4 * 4**wrong` falls outside `[0, 12 * 4**wrong)` for *every* `wrong != order` — too large when `wrong < order`, negative when `wrong > order`. So the old body reached `hp.pix2ang` with an invalid pixel and the crate aborted:

```
>>> old_uniq2geo(int(geo2uniq(45.0, -122.0, order=6)), 18)
pyo3_runtime.PanicException: Wrong hash value: too large.
```

That is arguably worse than the reported behavior — an unrecoverable Rust panic reachable from an ordinary bare call, since the default was 18 while (for example) the notebook's chain encodes at order 9. Either way the fix is the same and the defect is gone. Two tests pin it: `test_order_cannot_disagree_with_the_value` (the arithmetic reason no second opinion exists, checked against all 30 orders) and `test_default_order_no_longer_decides_the_answer` (a bare call is correct at orders 9, 18 and 29).

## Testing

`pytest`, `flake8 mortie --select=E9,F63,F7,F82` and `ruff check mortie` after every phase. No Rust changed, so no `maturin develop` rebuild was required.

| | tests | flake8 (CI gate) | ruff |
|---|---|---|---|
| `main` baseline | 757 passed / 12 skipped | clean | 74 findings |
| phase 1 | 768 / 12 | clean | 74 |
| phase 2 | 772 / 12 | clean | 70 |
| phase 3 | 782 / 12 | clean | 65 |
| phase 4 (final) | **782 passed / 12 skipped** | clean | 65 |

Ruff's count *falls* because the rewritten docstrings are numpydoc-compliant where the originals were not (`unique2parent` alone cleared four). Every remaining finding is pre-existing and untouched. Coverage 97%.

Real numbers from the new tests (`mortie/tests/test_tools.py`, 26 added):

* **Order decode** verified for the first, middle and last UNIQ value at every order 0-29, and at all 29 `k`/`k+1` boundaries.
* **Mixed-order decode**: one lat/lon encoded at orders `[3, 7, 12, 18, 29]` into a single array decodes element by element, each element matching its own single-resolution answer and re-encoding to the identical UNIQ value.
* **Round trip** `geo2uniq -> uniq2geo` over 50 pseudorandom points at orders `(0, 1, 5, 11, 18, 24, 29)`: re-encoding each returned centre reproduces the input UNIQ exactly, and at order >= 18 the centre is within `1e-3` degrees of the input in both lat and lon.
* **Mixed-order round trip**: 40 points at 40 independently drawn orders — `geo2uniq -> uniq2geo -> geo2uniq` is the identity, and `_uniq_orders` recovers the requested orders exactly.
* **Scalar/array order equivalence**: for both encoders, `order=k` and `order=np.full(n, k)` are `assert_array_equal` at orders `(0, 4, 13, 29)`, and a per-element array matches calling one element at a time.
* **Whole-family loop**: encode 40 points at mixed orders with `geo2uniq`, take the result apart with `unique2parent`, re-encode the pieces with `norm2uniq` — back to the identical array.
* **`unique2parent` back-compatibility**: 20 random cells at each order 0-20 give byte-identical answers to the retired `(uniq // 4**(order-1) - 16) // 4` body.
* **Kind-less claim, executed rather than asserted in prose** (`test_uniq_cannot_carry_the_point_kind`): at 12.5N 46.25W the area word and the point word are distinct packed words with `is_point` False/True — and both reduce to the *same* UNIQ cell, which is also what `geo2uniq(lat, lon)` returns by default.

## Existing tests changed — two, both deliberate

1. `mortie/tests/test_mort_inverse.py:193` — `tools.uniq2geo(uniq, order)` -> `tools.uniq2geo(uniq)`. **Signature change, not an incompatibility**: the caller decoded `order` from the word two lines earlier and built `uniq` from that same order, so the argument carried nothing the value did not already have.
2. `mortie/tests/test_tools.py::TestUnique2Parent::test_unique2parent_mixed_resolution_raises` — replaced by `test_unique2parent_mixed_resolution`, which asserts the correct per-element parents for the same order-6/order-7 input instead of asserting `NotImplementedError`. **The test was pinning the behavior this issue asks to remove**; deleting the raise necessarily invalidates it. Nothing was weakened — the class gained four further tests (mixed orders across all 12 base cells at five orders, legacy equivalence, scalar handling, invalid input).

No other test needed touching. In particular, no `norm2uniq` caller broke: all four in-tree call sites (`tools.py:535`, `:593`, `:785`, `:1049` pre-diff) already pass `order` explicitly, so the default change is inert for them.

## Notes

* No new dependency; numpy stays the only runtime dep.
* `USAGE.md` "Working with HEALPix Unique Identifiers" rewritten. The old snippet had a duplicated trailing comment and three opaque literals; the replacement is a runnable example with verified outputs, showing the scalar and per-element order forms and the argument-free decoders.
* `examples/ExampleUsage.ipynb` cell 15 **not patched** — see below.

## Questions for review

None of these blocked the work; all are judgement calls better made by a human.

1. **Three deliberate signature changes, no known external consumer.** `uniq2geo` loses a parameter; `geo2uniq` and `norm2uniq` change their default from 18 to 29. The zagg audit recorded on the issue found zero references to `uniq2geo`, `geo2uniq`, `norm2uniq` or `unique2parent`, and in-tree every `norm2uniq` call site already passes `order` explicitly. Re-confirming that audit before merge is cheap insurance.
2. **UNIQ is kind-less — the `MAX_ORDER` default is an *area* cell.** Implemented the default as directed, but "for lat/lon points this would be the point type" cannot be honored by this encoding: `4 * 4**order + nested` is an order plus a cell index, with no kind bit and no spare state. Point-vs-area is a `decimal_morton` packed-word concept carried by the suffix range (`0..=47` area, `48..=63` point, `docs/specification.md` §1), and `norm2mort` routes through the kernel's `from_nested`, documented "the result is always `Kind::Area`". Both encoder docstrings and `USAGE.md` now say so and cross-reference the APIs that *do* give point semantics from lat/lon — `geo2mort(lats, lons)` and `MortonIndexArray.from_latlon(..., points=True)`. Deliberately **not** adding a `points=` flag to `geo2uniq`: it cannot work without changing the return type from UNIQ to packed words, which is `geo2mort`.
3. **`examples/ExampleUsage.ipynb` cell 15 is already broken on `main`, independently of this PR.** Its chain is `geo2uniq -> unique2parent -> heal_norm -> norm2mort`, and it passes `uniq` where `heal_norm` expects a NESTED index, so `normed` comes back `4 * 4**order` too large and `norm2mort` raises `ValueError: nested index ... too large for depth 9 (base 14 > 11)`. Reproduced on `origin/main` with synthetic order-9 input; identical failure there and here, and `unique2parent` returns the same parents on both. Left untouched per CLAUDE.md §4 (don't fix unrelated pre-existing failures) and because `heal_norm` is being removed under PR #130 anyway — that PR has to rewrite the cell regardless. Happy to fix it here instead if preferred.
4. **Should `_uniq_orders` be public?** It is the UNIQ counterpart of the public `orders_of` and is plausibly useful to callers holding MOC data. Kept module-private because exporting it widens the public API beyond what the issue asked for. Can promote it to `uniq_orders` and add it to `__all__` in this PR if wanted.
5. **New validation, beyond the letter of the issue.** `_uniq_orders` raises `ValueError` for values outside `[4, 4**31)`, and `_encoder_orders` raises for orders outside `0..MAX_ORDER` or a mismatched order-array length. Previously all three cases returned garbage or panicked. Believed strictly an improvement and no existing test exercised any of them, but it is new behavior.
6. **`mortie/tools.py` is over the ~1000-line guidance** — 1039 lines on `main` before this PR, and this adds to it. Not splitting it here: a split during a behavior change makes review harder, and CLAUDE.md §4 asks for discussion first. Worth its own issue if the threshold is meant to bind.
7. **Pre-existing, untouched:** `uniq2geo` returns longitude in `[0, 360)` rather than `[-180, 180)`, inherited from `hp.pix2ang` and propagated to `mort2geo`. Out of scope here; the round-trip test normalizes before comparing. Say the word if it should be folded in or filed separately.

https://claude.ai/code/session_01N4YrkSCpk7ApjDPRNc4khi


mortie/tests/test_tools.py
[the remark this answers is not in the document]
espg: 🤖 *from Claude*

Fixed in `fd8c213` — ran `ruff check --select I --fix` on the file, which sorted the whole block (`numpy` before `pytest`, the `numpy.testing` names, and my added `_healpix` import before `tools`).

Worth noting the block was already unsorted on `main` for the `numpy`/`pytest` pair; adding an import to it is what brought the existing I001 into the diff. Suite still 782 passed / 12 skipped after the reorder.

---

mortie/tools.py
```diff
@@ -73,22 +73,83 @@ def res2display(max_order=MAX_ORDER):
...
+    orders = _uniq_orders(u)
+
+    # nested = uniq - 4 * 4**order, and the base cell is nested // 4**order.
+    shift = 2 * orders
+    parent = (u - (np.int64(1) << (shift + np.int64(2)))) >> shift
+
+    return parent[0] if is_scalar else parent
 
 
 def heal_norm(base, order, addr_nest):
```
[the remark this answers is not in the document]
espg: 🤖 *from Claude*

Not fixing this one — replying instead, per CLAUDE.md §2.

The D103 is on `heal_norm`, which this PR does not touch:

```python
def heal_norm(base, order, addr_nest):
    N_pix = hp.order2nside(order)**2
    addr_norm = addr_nest - (base * N_pix)
    return addr_norm
```

It has been undocumented on `main` (`git show origin/main:mortie/tools.py` flags the same D103), and it moved into the diff only because the phase-1 insertion of `_uniq_orders` shifted its line number. Adding a docstring would also collide with PR #130, which deletes `heal_norm` outright under the #68 freeze audit.

For what it is worth the net direction is the right one: `ruff check mortie` reports 64 findings on this branch against 74 on `main`, because the docstrings this PR did rewrite (`unique2parent`, `geo2uniq`, `norm2uniq`, `uniq2geo`) are now numpydoc-compliant where the originals were not.

---

mortie/tools.py
```diff
@@ -405,23 +557,45 @@ def mort2norm(morton):
...
+        UNIQ encoded pixel index/indices.
+
+    Raises
+    ------
+    ValueError
+        If an order lies outside 0-``MAX_ORDER``, or an order array's length
+        does not match the input.
     """
+    n = np.broadcast(np.asarray(normed), np.asarray(parent)).size
+    order = _encoder_orders(order, n)
```

**Blocking — the new array-`order` path silently returns `float64` and corrupts UNIQ values at orders ≥ 25 for unsigned input.**

`_encoder_orders` returns an **`int64`** array (`tools.py:231`). The body below then does `nside = 2**order` → `int64`, `N_pix` → `int64`, and `parent * N_pix`. When `normed`/`parent` are **`uint64`** — the dtype `mortie`'s own packed words use, and the dtype MOC/HEALPix indices arrive in from `mocpy`/`astropy-healpix` — numpy has no common integer type for `uint64 × int64` and promotes to **`float64`**. The scalar-`order` path is immune because `2**order` is a Python `int` (weak under NEP 50), so it stays `uint64`.

Reproducer against `fd8c213`:

```python
import numpy as np
from mortie import norm2uniq
n64 = np.array([0, 1], dtype=np.uint64)
p64 = np.array([0, 11], dtype=np.uint64)
print(norm2uniq(n64, p64, 29).dtype,             [int(x) for x in norm2uniq(n64, p64, 29)])
print(norm2uniq(n64, p64, np.array([29, 29])).dtype, [int(x) for x in norm2uniq(n64, p64, np.array([29, 29]))])
```

```
uint64  [1152921504606846976, 4323455642275676161]
float64 [1152921504606846976, 4323455642275676160]
```

The second element is off by one — **a different UNIQ cell**, returned with no error. Scanning every order with `normed = 4**k - 1, parent = 11`:

| order | scalar-`order` | array-`order` | equal |
|---|---|---|---|
| 24 | 4503599627370495 | 4503599627370495 | yes |
| 25 | 18014398509481983 | 18014398509481984 | **no** |
| 29 | 4611686018427387903 | 4611686018427387904 | **no** |

Corruption starts at order 25 (where UNIQ crosses 2\*\*53). Note the order-29 case rounds to exactly `4**31`, which `_uniq_orders` then rejects — so the same defect is a `ValueError` from `unique2parent`/`uniq2geo` in some cases and a wrong-but-plausible cell in others.

Two separate problems here, and both are worth fixing:

1. **Value corruption** (orders ≥ 25).
2. **Dtype contract break at *every* order** — the array-`order` path returns `float64` unconditionally for `uint64` input, even at order 0, while the scalar path returns an integer dtype. `geo2uniq` avoids this by allocating `uniq = np.empty(n, dtype=np.int64)` (`tools.py:299`); `norm2uniq` has no equivalent.

Suggested fix: cast the inputs alongside the order, e.g. `normed = np.asarray(normed, dtype=np.int64)` / `parent = np.asarray(parent, dtype=np.int64)` before the arithmetic (`int64` holds the whole UNIQ range — max is `4**31 - 1 ≈ 4.6e18` < `2**63 - 1`), or compute `N_pix` as a Python-int power so it stays weak. Whichever way, please pin it with a test: `test_norm2uniq_array_order_matches_elementwise` (`test_tools.py`) only ever builds `int64` arrays via `rng.integers(..., dtype=np.int64)`, so it cannot catch this. An `assert_array_equal(norm2uniq(u64_in, u64_par, np.full(n, 29)), norm2uniq(u64_in, u64_par, 29))` plus a `.dtype` assertion would.

---

mortie/tools.py
```diff
@@ -405,23 +557,45 @@ def mort2norm(morton):
...
+    int or ndarray
+        UNIQ encoded pixel index/indices.
+
+    Raises
+    ------
+    ValueError
+        If an order lies outside 0-``MAX_ORDER``, or an order array's length
+        does not match the input.
     """
+    n = np.broadcast(np.asarray(normed), np.asarray(parent)).size
```

**`n` is a broadcast *size* but `order` is required to be flat 1-D — for any input that is not 1-D, `norm2uniq` silently returns the wrong shape.**

`np.broadcast(...).size` collapses shape to a count, `_encoder_orders` then demands `orders.ndim == 1 and orders.size == n`, and the body broadcasts that flat array back against the *un-flattened* input shape. For a column vector the two disagree and numpy happily outer-broadcasts:

```python
import numpy as np
from mortie import norm2uniq
normed = np.array([[0], [1]])      # (2, 1)
parent = np.array([[0], [11]])     # (2, 1)
print(norm2uniq(normed, parent, 6).shape)                # scalar order
print(norm2uniq(normed, parent, np.array([6, 6])))       # array order
```

```
(2, 1)
[[16384 16384]
 [61441 61441]]
```

Shape `(2, 2)` instead of `(2, 1)`, four elements from a two-element input, no warning. The scalar-`order` path handles the same input correctly, so this is a new inconsistency between the two paths rather than a pre-existing limitation.

`geo2uniq` has the mirror of this at `tools.py:302` but fails loudly instead — `lats[mask]` applies a length-`n` 1-D boolean to axis 0 of a 2-D array:

```python
geo2uniq(np.zeros((2, 3)), np.zeros((2, 3)), np.array([6, 6, 6, 7, 7, 7]))
# IndexError: boolean index did not match indexed array along axis 0;
#   size of axis is 2 but size of corresponding boolean axis is 6
```

Cheapest consistent fix is to make the contract explicit: reject non-1-D `normed`/`parent`/`lats`/`lons` in the array-`order` path (or `ravel` and reshape the result back). Silently returning a `(2, 2)` is the one outcome that shouldn't survive. Either way the docstrings currently promise plain "array-like" with no dimensionality caveat.

---

mortie/tools.py
```diff
@@ -73,22 +73,83 @@ def res2display(max_order=MAX_ORDER):
...
+    ndarray
+        ``int64`` order per element, 0-``MAX_ORDER`` (scalar in -> length-1
+        ndarray).
+
+    Raises
+    ------
+    ValueError
+        If any value lies outside the UNIQ range for orders 0-``MAX_ORDER``.
+    """
+    u = np.atleast_1d(np.asarray(uniq, dtype=np.int64))
```

**`np.asarray(..., dtype=np.int64)` is the wrong gate for the two out-of-range cases the docstring promises to catch.** It casts unsafely, so it either raises the wrong exception type or truncates before the range check runs.

`asarray` is reached before `bad`, so the documented `ValueError` never fires:

```python
>>> tools._uniq_orders(2**63)
OverflowError: Python int too large to convert to C long
```

The `Raises` section says `ValueError`, and `unique2parent` / `uniq2geo` both re-document it as their own `ValueError` contract — a caller writing `except ValueError` around `uniq2geo` on untrusted MOC data gets an `OverflowError` through the guard instead. Same shape for a `uint64` array holding a value ≥ `2**63`: it wraps to a negative `int64` and reports as "not a valid UNIQ", which is the right *outcome* by luck rather than by check.

Float input is the quieter half — it truncates silently rather than being rejected:

```python
>>> tools._uniq_orders(4.0 ** 30 - 0.5)   # not an integer at all
```

Suggestion: convert with `np.asarray(uniq)` first, reject non-integer kinds and anything outside `[4, 4**(MAX_ORDER+2))` with the existing `ValueError`, and only then narrow to `int64`. `test_out_of_range_raises` currently exercises `0`, `3` and `4**31` only — all of which are representable in `int64`, so the suite passes straight over this.

Worth noting for the record, since the PR body rests a lot on it: I brute-forced the bucket boundaries independently and **the decode itself is exact**. All four of `4**(k+1)`, `4**(k+1)+1`, `4**(k+2)-2`, `4**(k+2)-1` decode to `k` for every `k` in `0..29`, with zero mismatches, and `4**31` / `-1` / `3` are correctly rejected. The `searchsorted` formulation and the "orders tile without gaps" argument both hold up.

---

espg (commented):
🤖 *from Claude (review)*

Reviewed at `fd8c213`. Three findings, all in the **encoder** half of the change — the decoders (`_uniq_orders`, `unique2parent`, `uniq2geo`) held up under everything I threw at them.

**Lead finding: `norm2uniq`'s new array-`order` path silently returns `float64` and corrupts values at orders ≥ 25.** `_encoder_orders` hands back an `int64` array, so `parent * N_pix` becomes `uint64 × int64` → `float64` for unsigned input. `norm2uniq(uint64_arr, uint64_arr, np.array([29, 29]))` returns `4323455642275676160` where the scalar-`order` call returns `...161` — a different UNIQ cell, no error. At order 29 the rounding can land on exactly `4**31`, which `_uniq_orders` then rejects, so the same defect surfaces as a spurious `ValueError` in some cases and a wrong cell in others. The scalar path is immune (Python-int `2**order` stays weak under NEP 50), which is why nothing in-tree caught it. Full table and reproducer inline.

The other two: `n` is a broadcast *size* while `order` must be flat 1-D, so a `(2, 1)` input plus a length-2 order outer-broadcasts to a silent `(2, 2)` (`geo2uniq` fails loudly on the mirror case — the asymmetry is the problem); and `np.asarray(..., dtype=np.int64)` in `_uniq_orders` runs *before* the range check, so `2**63` raises `OverflowError` rather than the `ValueError` that `uniq2geo` and `unique2parent` both re-document as their contract.

Common thread worth pulling on: **the array-`order` path is under-tested relative to the scalar path.** `test_norm2uniq_array_order_matches_elementwise` builds every array with `rng.integers(..., dtype=np.int64)` and 1-D shapes, and `test_out_of_range_raises` only uses values representable in `int64` — the three findings above sit exactly in the gaps that leaves. A dtype assertion, a `uint64` case, and a non-1-D case would close all three.

Verified independently and clean, so no comments on these:

* **Boundary decode is exact.** Brute-forced `4**(k+1)`, `4**(k+1)+1`, `4**(k+2)-2`, `4**(k+2)-1` for every `k` in `0..29` — zero mismatches. `4**31`, `-1`, `0`, `3` all correctly rejected. The tiling argument in the PR body is sound, and the `searchsorted` formulation implements it faithfully.
* **Return-type contracts match `origin/main`.** Diffed `uniq2geo` (scalar and array), `mort2geo` (scalar and array), and `unique2parent` (array) against `git show origin/main:mortie/tools.py` loaded side by side. Identical types, shapes and values throughout. One cosmetic drift: scalar `unique2parent` now returns `np.int64` where it returned a Python `int` — harmless, flagging only for completeness.
* **Edge cases**: empty arrays, 0-d arrays, lists, scalars, negatives, order 0 across all 12 base cells, and `uint64` UNIQ input into `uniq2geo` all behave. 2-D input into `uniq2geo`/`geo2uniq` raises rather than silently mis-decoding, and did not work before this PR either.
* **Lint**: `flake8 mortie --select=E9,F63,F7,F82` clean; `ruff check mortie` = 64 findings, none of them on a line this PR added (`tools.py` findings are all in pre-existing code — `heal_norm`, `geo2mort`, `on_antimeridian`). The numpydoc claim in the PR body checks out.
* **The `USAGE.md` snippet actually runs and its pinned outputs are correct** — I executed it verbatim and got `array([1683881, 3530931, 2051535])`, `array([26310, 225979639, 8604763086340])` and `array([2, 9, 3])` as written.
* `pytest mortie/tests/test_tools.py mortie/tests/test_mort_inverse.py` → 89 passed.

---

espg:
🤖 *from Claude*

Status: all four phases are landed and CI is green on `fd8c213` — tests pass on 3.10/3.11/3.12, `ruff` job green, wheels build on Linux/macOS/Windows, and Codecov reports all modified lines covered with project coverage up +0.14% to 94.64%. Applying `waiting`.

Both ruff-bot threads are answered — I001 fixed in `fd8c213`, D103 replied to (it is `heal_norm`, pre-existing and untouched).

Two things to flag before this is reviewed.

**(1) The self-review subagent did not finish.** It opened a pending review (`4780351511`) against `fd8c213` and then stalled without submitting it. A pending review is only visible to its author, so those comments are sitting in the review UI unsubmitted — they can be read and either submitted or discarded from there. I have not submitted them myself: I cannot read pending comments through the API, and publishing text I have not read is not something I will do on my own initiative. The adversarial pass should be treated as **not performed**, and this PR reviewed without assuming it happened.

**(2) `draft` stays on, deliberately.** CLAUDE.md §2 says to clear draft once all phases are complete *and* the PR is not waiting on adversarial review — per (1) it is, so the condition is not met. The §1 golden rule also says to stop at draft and let a human take it from there. Marking ready-for-review is a side-effecting action under §6 and is not mine to take.

The substance is all in the PR description; the seven items under **Questions for review** are the ones worth a decision, and none of them blocked the work. The two most likely to matter:

* the encoders' `MAX_ORDER` default produces an **area** cell — UNIQ has no kind bit and structurally cannot carry point-vs-area (item (2) there, with the reasoning and the point-capable alternatives);
* `examples/ExampleUsage.ipynb` cell 15 is already broken on `main` independently of this change, and I left it alone (item (3)).

---

espg:
🤖 *from Claude*

**Ready to merge.** Correcting one thing in the PR body first, because it tells you to do the wrong thing.

## The body's "adversarial review did not complete" note is wrong — ignore it

It says the review subagent stalled and asks you to treat the adversarial pass as **not performed**. It *was* performed: review [`4780351511`](https://github.com/espg/mortie/pull/139#pullrequestreview-4780351511) is submitted against `fd8c213` with three inline findings plus a summary. The reviewer simply could not deliver its hand-off message back to the implementing agent, which inferred a stall from the silence. The review is real and its findings were substantive.

## All three findings are fixed and pinned

**(1) Blocking — `uint64` silently promoted to `float64` on the array-order path.** `_encoder_orders` yields `int64`, and `uint64 × int64` has no common integer type, so NEP 50 promotes to float. I reproduced it before fixing: at order 29 the array path returned `1.7293822569102705e+18` where the scalar path returned `1729382256910270465` — a *different UNIQ cell*, no error raised. Wrong from order 26 up; `float64` at every order.

The scalar path was immune because a Python int stays weakly typed, and every existing test built its inputs as `int64` where the promotion cannot happen — which is exactly why nothing caught it. `uint64` is what `mort2norm` actually returns, so it is the realistic input. Fixed by forcing the array path into `uint64`; regression test sweeps all 30 orders asserting both dtype and exact value.

**(2) Silent wrong output shape.** A `(2, 1)` input with a length-2 order outer-broadcast to `(2, 2)` — four results from a two-element input — while the scalar-order path handled the same input correctly. Now raises rather than letting the two paths disagree.

**(3) `OverflowError` through a guard documented to raise `ValueError`.** Fixed, and the first fix was wrong in an instructive way: `np.asarray(2**63)` yields `uint64`, and casting *that* to `int64` **wraps silently** rather than raising, so the `try/except OverflowError` never fired. It only looked right because every wrap lands negative and got caught downstream, reporting `-9223372036854775808` — a number the caller never passed. Now guarded explicitly in the unsigned domain; all four cases report the real value.

## `orders_of_uniq` is public, per your call

Promoted from `_uniq_orders`, mirroring `orders_of`: `uint8` out, scalar in → length-1 ndarray, per-element and mixed-order-native.

One deliberate divergence, documented rather than papered over: `orders_of` never validates, because every 6-bit morton suffix decodes to *some* order. UNIQ has no total decode — a value outside `[4, 4**31)` names no cell at any order — so this raises instead of inventing an answer. Exact mirroring would have meant returning garbage.

A trap that mirroring introduced, worth knowing since `order2res` already documents it for `orders_of`: the two internal callers do shift arithmetic on the decoded orders, and `uint8` wraps there. Both now cast to `int64` explicitly, with the reason in a comment.

**Naming is the one thing still worth a second of your time.** `orders_of_uniq` reads as "orders of [a] uniq" and parallels `orders_of`; `uniq_orders_of` was the alternative and scans worse. It is a 1.x freeze commitment, so it is a one-line rename now and a breaking change later.

## State

`main` merged in (clean — #130's `tools.py` changes intact alongside this branch's). **786 passed, 12 skipped.** `flake8` clean, `mkdocs build --strict` exit 0 with zero warnings. **All 23 CI checks green**, including the new Docs verify job.

## Standing for you, neither blocking

- The `MAX_ORDER` default yields an **area** cell — UNIQ structurally cannot carry point/area kind. Documented on both encoders with cross-references to `geo2mort` and `from_latlon(points=True)`, which do give point semantics. No `points=` flag added, since it cannot work without changing the return type to packed words, at which point it would just be `geo2mort`.
- `examples/ExampleUsage.ipynb` cell 15 is **already broken on `main`** — it feeds a UNIQ into `heal_norm` where a NESTED index is expected. Reproduced identically on `origin/main`, left untouched per §4, and PR #130 removed `heal_norm` anyway.

---

PR #154, opened 2026-08-07
Batch polygon coverage: polygons_to_morton_mocs (issue #153)

### LINKED ISSUE #153: Batch polygon coverage: polygons_to_morton_moc over ragged (offsets) arrays — one call, rayon across polygons
🤖 *from Claude* (filed at espg's direction, 2026-08-07)

## Motivation

zagg's shardmap build calls `morton_coverage_moc` once per granule footprint — 555,867 calls for a full ATL03 v007 catalog build. Measured on the California o9 baseline (englacial/zagg#396): **310.7 s single-pass**, and the two-stage experiment there shows roughly **two-thirds of that wall is per-call fixed overhead** (cell work shrank ~16× at a coarse order but wall only dropped 3.2×), not coverage compute. The fill already releases the GIL and runs rayon internally, so the fix is to cross the Python↔Rust boundary **once**, not 556k times, and let rayon distribute across polygons.

(The alternative — outer Python threading over the per-polygon calls — was tried in zagg and is abandoned: thread × rayon allocation churn spiked memory to ~85 GB. Batching in Rust is the ruled direction.)

## Proposed API

Core entry (numpy in/out, matching the existing `coverage.py` surface):

```python
values, out_offsets = mortie.polygons_to_morton_mocs(
    lats, lons,            # flat float64 arrays, all rings concatenated
    offsets,               # int64, arrow list layout: polygon i is [offsets[i], offsets[i+1])
    order=18,
    normalize=True,
)
# values: uint64 morton MOC words, all polygons concatenated
# out_offsets: int64, arrow list layout — MOC for polygon i is values[out_offsets[i]:out_offsets[i+1]]
```

- **Identity-preserving**: result i ↔ input polygon i (this is what `multipolygon_to_morton_moc` cannot do — it unions rings).
- Rust side: `par_iter` over polygons around the existing `polygon_to_morton_moc` kernel; GIL released for the whole call; one allocation pass for the ragged output.
- `tolerance=` / `max_cells=` batch variants mirroring `polygon_to_morton_moc_tolerance` / `_budget` — same per-polygon semantics, or a single shared setting (simplest; per-polygon arrays of tolerances feel like YAGNI — flag if disagreeing).
- Error posture: fail-fast with the offending polygon index in the message (`polygon 4217: <ring_validity reason>`), matching the scalar path's exceptions. Per-item error sentinels considered and not proposed (callers batching 556k granules want the bad granule named, not a NaN-shaped hole).

### Arrow skin (optional extra, lazy like `mortie.arrow`'s existing pattern)

`mortie.arrow.polygons_to_morton_mocs(list_array_or_table_column, ...)` accepting a pyarrow `list<struct<lat,lon>>` (or `list<list<double>>` pair) and returning a **`ListArray` whose values are the registered `morton_index` extension type** — one ragged MOC per polygon, parquet-ready. This drops directly into the `footprint_cells` catalog column planned as englacial/zagg#396 phase 2, with no conversion.

## Acceptance criteria

- Byte parity: for every polygon, batch output == the scalar `morton_coverage_moc` output (property test over randomized rings including antimeridian/pole/basecell-edge cases, plus the zagg identity fixtures).
- Scaling: on an N-core machine, batch wall ≈ scalar-loop rust-time / cores; the per-call overhead term is gone. Bench with `src_rust/benches/coverage_bench.rs` machinery + a Python-level comparison at ≥100k synthetic footprints.
- GIL released across the batch (verifiable: a second Python thread stays live during a large call).
- No behavior change to any existing entry point.

## Consumers

- englacial/zagg#396 phase 1: `_intersect_mortie` rewires its per-granule loop to one batch call (the crashed-session serial-vs-parallel pair-set identity tests get salvaged as the correctness pin).
- englacial/zagg#396 phase 2: the one-time `footprint_cells` column indexing pass over the 556k-granule catalog rides the same call (arrow skin).

## Naming (espg ruling, 2026-08-07)

`polygons_to_morton_mocs` — **plural output**: the name must distinguish the many→many contract (one MOC per input polygon) from `multipolygon_to_morton_moc`'s many→one union. Batch variants and the arrow skin pluralize the same way.

espg:
Closes #153

## What

Batch polygon coverage: `polygons_to_morton_mocs(lats, lons, offsets, order, tolerance=, max_cells=, normalize=)` — one Python↔Rust crossing for a whole ragged (arrow-list-layout) polygon set, rayon `par_iter` across polygons around the existing scalar `polygon_to_morton_moc` kernel, GIL released for the entire call, ragged `(values, out_offsets)` output in the same layout. Identity-preserving: result `i` is byte-identical to the scalar `morton_coverage_moc` on input polygon `i`. The plural *MOCs* marks the many→many contract (one MOC per input polygon), distinguishing it from `multipolygon_to_morton_moc`'s many→one union — espg naming ruling recorded on #153. Plus the Arrow skin `mortie.arrow.polygons_to_morton_mocs`: `list<struct<lat,lon>>` (or paired `list<double>`) in, `ListArray` of registered `morton_index` extension words out — parquet-ready for the zagg `footprint_cells` column (englacial/zagg#396 phase 2).

## Approach

- New Rust submodule `src_rust/src/coverage/batch.rs` (coverage.rs is already at ~1,360 lines): serial structural pre-validation (offsets monotonicity/bounds, 3-vertex minimum, finiteness), then `par_iter` with per-polygon `catch_unwind`, then ragged assembly into a `BatchMocs` (values / offsets / budget-raise telemetry) The parallel pass runs in chunks of 2,048 polygons and each chunk is copied out as it lands, so the whole batch's per-polygon `Vec`s never coexist with the concatenated output — peak ~1.1x the result rather than ~2.3-2.5x (measured; see the review thread).
- **Error determinism (decision):** errors name the **lowest-index** offending polygon, deterministically — pre-validation scans in index order, and the parallel pass materializes all per-polygon results before scanning for the first `Err`, so rayon's schedule cannot change which error surfaces. (Short-circuiting `try_reduce` was rejected for exactly that nondeterminism.)
- `tolerance` / `max_cells` are **single shared settings** (mutually exclusive, per the issue; per-polygon arrays read as YAGNI). A `max_cells` below a polygon's representable floor is raised per polygon exactly as in the scalar path; one summary `UserWarning` reports the raised count and names the lowest-index polygon.
- **Offsets contract (ruled):** strict core, re-basing skin. The kernel requires `offsets[0] == 0` and `offsets[-1] == len(lats) == len(lons)` — exact coverage, with the failing endpoint named — so stale or misaligned offsets that stay in bounds cannot silently produce a valid-looking wrong covering. The Arrow skin absorbs the slice mechanics: a sliced input's values are sliced to its own window (zero-copy view) and its offsets re-based by one scalar subtraction, together, in `_ragged_from_arrow`.
- pyfunction `rust_polygons_coverage_mocs` wraps the kernel in `py.allow_threads` around the whole batch; the Python wrapper in `mortie/coverage.py` mirrors the `morton_coverage_moc` surface (numpy in/out, degrees→radians tolerance, numpydoc).
- Arrow skin follows `mortie/arrow.py`'s lazy-pyarrow pattern; returns `ListArray` (int32 offsets) with automatic `LargeListArray` fallback past 2**31−1 total cells.

## Phases

- [x] Phase 1 — Rust batch kernel + ragged validation + pyfunction (GIL released) + Python wrapper `polygons_to_morton_mocs` (e44e8f6)
- [x] Phase 2 — shared `tolerance=` / `max_cells=` variants + budget-raise warning + deterministic lowest-index error propagation (da1e8fb)
- [x] Phase 3 — parity/property tests: randomized rings, antimeridian / pole-adjacent / basecell-edge rings, Antarctic basin fixtures (incl. the pole+antimeridian basin 2), closed-ring + `normalize=False` funneling, offsets edge cases, lowest-index error naming, GIL-release test (9908fbb)
- [x] Phase 4 — Arrow skin `mortie.arrow.polygons_to_morton_mocs` + tests gated on pyarrow (incl. parquet round-trip with the `morton_index` identity preserved, chunked + sliced inputs) (5f17e84)
- [x] Phase 5 — benches: criterion `coverage_batch_512` group (batch vs in-Rust scalar loop) + `benchmarks/measure_batch_coverage.py` standalone 100k-footprint comparison (bd036d3)

## Benchmarks (Apple Silicon, 10 cores)

Criterion, 512 synthetic ~1° footprint quads at order 8 (pure-Rust, isolates the rayon-across-polygons win from Python overhead):

| bench | time |
|---|---|
| `coverage_batch_512/scalar_loop` | 40.7 ms |
| `coverage_batch_512/batch` | **2.89 ms** (14.1×) |

Python level, `python benchmarks/measure_batch_coverage.py 100000 8` (the issue's acceptance workload):

```
n=100000 order=8 cores=10 cells=1589038
scalar loop :     9.28 s  (   92.8 us/polygon)
batch       :     0.47 s  (    4.7 us/polygon)
speedup     :    19.89x
```

The per-call fixed overhead term is gone (92.8 → 4.7 µs/polygon); batch wall ≈ loop time / cores, exceeding the core count because per-polygon rayon fork-join on tiny footprints is itself overhead the batch layout avoids.

## Testing

- `cargo test`: 9 unit tests in `coverage::batch` (parity vs the scalar kernel incl. tolerance/budget variants, budget-raise telemetry, empty batch, exact-coverage offsets rejection, lowest-index error naming, injected-panic capture, layout rejection) — all pass, 266 passed / 1 ignored.
- `pytest -v`: 1046 passed, 12 skipped locally after `maturin develop --release` (29 new tests across `test_coverage_batch.py` / `test_coverage_batch_arrow.py`, incl. sliced / sliced-pair / chunked-plus-sliced Arrow inputs asserted identical to their unsliced equivalents).
- `cargo fmt --check`, `cargo clippy --lib --benches` (no new warnings; one pre-existing `unnecessary_sort_by` in rank_xy untouched), `flake8 mortie benchmarks --select=E9,F63,F7,F82`, `numpydoc lint` on the touched modules — clean.
- Local note: `cargo bench` fails to **link** on macOS on unmodified `main` too (extension-module benches; `_PyBaseObject_Type` unresolved) — pre-existing, not introduced here; the numbers above were produced by linking libpython explicitly via `RUSTFLAGS`. CI's `cargo codspeed build` on Linux is unaffected.

## Questions for review

- Batch `tolerance`/`max_cells` are a **single shared setting** (per the issue: per-polygon arrays read as YAGNI). Flag if per-polygon arrays are wanted after all.
- ~~Loose offsets contract~~ — **ruled and implemented as strict core / re-basing skin** (https://github.com/espg/mortie/pull/154#issuecomment-5222533167): exact coverage in the kernel, slice mechanics absorbed by `mortie.arrow`. Nothing outstanding here.
- **`--features descent-stats` interaction (flagged, not fixed).** The instrumentation collector is documented as single-descent-at-a-time (`src_rust/src/coverage/descent_stats.rs`: *"Concurrent descents from multiple Python threads would interleave their records"*), and the batch entry runs descents concurrently by design — so under a `descent-stats` build the per-descent cause/leaf records become a mixture across whichever polygons were descending together, and the `take`-run-`take` protocol no longer isolates one polygon. This affects feature-gated measurement builds only (release builds compile none of it) and the harness already runs one descent at a time; flag if the batch should instead refuse to run, or serialize, under that feature.
- **Pre-existing, unrelated:** `dissolve::tests::sub_hemisphere_cover_still_dissolves` fails intermittently (`dissolve(&cover, 1)` returns `Err` in ~15% of runs). Reproduced on unmodified `main` at the same rate in a clean worktree, so it is not from this PR and is left untouched; worth its own issue.
- The budget-raise warning is one **summary** warning per call (count + lowest-index example) rather than one warning per raised polygon — 556k warnings would be noise. Flag if per-polygon warnings are wanted.

## Review fold

Adversarial-review findings folded, one commit each:

- `2049261` — strict offsets contract + re-basing Arrow skin (the ruling; findings [r3739074407](https://github.com/espg/mortie/pull/154#discussion_r3739074407) and [r3739074559](https://github.com/espg/mortie/pull/154#discussion_r3739074559)).
- `50ab086` — chunked ragged assembly; peak 2.26x → 1.08x and 2.50x → 1.34x of the returned array on the two measured workloads ([r3739074337](https://github.com/espg/mortie/pull/154#discussion_r3739074337)).
- `0c2c6cc` — injected-panic unit test over the real assembly path, covering the `catch_unwind` arm ([r3739074483](https://github.com/espg/mortie/pull/154#discussion_r3739074483)).
- `f8435b8` — both entry points added to the mkdocstrings `members:` lists, `USAGE.md` and the `coverage_methods.md` selection table; one-ring-per-entry (no multipart/holes) stated in both docstrings ([r3739074648](https://github.com/espg/mortie/pull/154#discussion_r3739074648)).
- `1155f61` — ruff bot D401 on the new Arrow helper docstring.


espg:
🤖 *from Claude*

Status: all 5 phases pushed (e44e8f6 → bd036d3), checklist complete, test matrix / numpydoc / arro3 / Python-bench CI green. Draft stands pending the adversarial review cycle; bench numbers and the three open design questions are in the PR body.

espg:
🤖 *from Claude*

espg rulings on the Questions for review (in-session, 2026-08-07):

- **(1) shared tolerance/budget** — stands as implemented.
- **(2) offsets contract — RULED: strict core, re-basing skin.** The kernel requires `offsets[0] == 0` and `offsets[-1] == len(lats) == len(lons)` (exact coverage; endpoint mismatches become loud errors at the boundary). The Arrow skin absorbs slice mechanics: values slice as a zero-copy view, offsets re-based by one scalar subtraction (~8 B × n_polygons). Rationale: the loose/verbatim contract saves essentially nothing (the re-base is microseconds) while silently accepting stale or misaligned offsets that stay in-bounds — valid-looking wrong coverings feeding granule→shard assignment is exactly the silent-error class this API exists to kill. Tightening later would break callers; loosening later would not — so strict now, pre-release.
- **(3) single summary budget warning** — stands as implemented.

The (2) change folds in with the adversarial-review findings (cycle in progress).

src_rust/src/coverage/batch.rs
```diff
@@ -0,0 +1,316 @@
...
+) -> Result<BatchMocs, String> {
+    // Scanning in index order makes the surfaced error the lowest-index
+    // failure — deterministic under any rayon schedule.
+    let covers: Vec<(Vec<u64>, usize)> = covers.into_iter().collect::<Result<_, _>>()?;
+    let total: usize = covers.iter().map(|(c, _)| c.len()).sum();
+    let mut values = Vec::with_capacity(total);
+    let mut offsets = Vec::with_capacity(covers.len() + 1);
+    offsets.push(0i64);
+    let mut raised: Option<(usize, usize, usize)> = None;
+    for (i, (cover, effective)) in covers.iter().enumerate() {
```
espg: 🤖 *from Claude (review)*

[medium] **Ragged assembly peaks at ~2–2.5× the returned MOC array: every per-polygon `Vec<u64>` stays live while the concatenation buffer is allocated and filled.**

`assemble` sizes `values` from the fully materialized `covers`, then copies through a *borrowing* iterator — so the N per-polygon allocations and the single `total`-sized buffer coexist for the whole copy:

```rust
let total: usize = covers.iter().map(|(c, _)| c.len()).sum();
let mut values = Vec::with_capacity(total);     // allocated while all N covers are still live
...
for (i, (cover, effective)) in covers.iter().enumerate() {   // borrow: nothing is freed until the loop ends
    values.extend_from_slice(cover);
```

Measured on this branch after `maturin develop --release`, peak RSS sampled every 2 ms in a helper thread (`ru_maxrss` delta over the pre-call baseline):

| workload | returned `values` | peak RSS above baseline |
|---|---|---|
| 50k polygons, order 12, ~1,396 cells/polygon | 558 MB | **1,165 MB (2.09×)** |
| 556k polygons, order 9, ~34 cells/polygon | 152 MB | **377 MB (2.48×)** |

`into_pyarray_bound` hands the `Vec` to numpy without copying, so the excess is exactly the still-live per-polygon covers plus the `Vec<Result<..>>` spine.

The second row is the issue's stated target (556,867 ATL03 footprints) with toy 1° quads; real granule strips produce far larger per-granule MOCs, so the same multiplier lands on a multi-GB peak for a call that is advertised as the memory-safe replacement for the abandoned thread×rayon path (issue #153: *"one allocation pass for the ragged output"*, and the ~85 GB blow-up it exists to avoid).

Note that `covers.iter()` → `into_iter()` alone does **not** fix it — the `with_capacity(total)` allocation still happens while every cover is live. Getting under 2× needs chunked assembly (e.g. `par_chunks` of polygons appended into a growing `values`, so only one chunk's covers are resident), or at minimum a documented peak-memory note on the Python wrapper so the catalog-scale caller can size for it.

src_rust/src/coverage/batch.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    #[test]
+    fn empty_batch() {
+        let out = batch(&[], &[], &[0], 6).unwrap();
+        assert!(out.values.is_empty());
+        assert_eq!(out.offsets, vec![0]);
+    }
+
+    #[test]
+    fn nonzero_start_offset() {
```
espg: 🤖 *from Claude (review)*

[medium] **This test pins the loose offsets contract *in the core* — here is every site the strict-contract fold has to touch, so none is missed.**

`nonzero_start_offset` asserts that the core accepts `offsets = [3, 7]` against a 10-vertex input, i.e. exactly the `offsets[0] > 0` + trailing-unused-vertices behaviour the ruling removes. It has a Python twin. Both are core-level and must be inverted (assert the error) with the slice-parity content moving to the Arrow skin's tests.

**Rust — `src_rust/src/coverage/batch.rs`**

1. `validate_batch`: `if offsets[0] < 0` becomes `offsets[0] != 0`, and a new endpoint check `offsets[n_polys] == lats.len()` is needed (nothing today requires the offsets to *cover* the vertex arrays — `[0, 3]` against 10 vertices passes).
2. Doc on `polygons_to_morton_mocs`: *"`offsets[0]` need not be 0, so a sliced arrow array's offsets pass straight through"* — now false.
3. `validate_batch`'s own doc comment ("offset monotonicity and bounds") should name the endpoint requirement.
4. `nonzero_start_offset` (this test) — invert.
5. `errors_name_lowest_index_polygon`: the `&[0, 3, 99]` case asserts `polygon 1: … exceeds`. If the endpoint check is written as a batch-level error it will fire **before** the per-polygon bounds error and this assertion breaks — keep the per-polygon bounds check ahead of the endpoint check, or update the expectation. The `&[0, 7, 3, 10]` and `&[0, 3, 5, 5, 10]` cases stay valid (both end at `lats.len() == 10`).
6. `bad_order_and_layout_rejected`: `&[-1, 3]` asserts only `is_err()` in Rust, but the Python twin matches the string `"non-negative"` (see below) — pick the new message once and update both.
7. `empty_batch` (`&[], &[], &[0]`) stays valid under the strict contract (`0 == 0 == len`); please don't churn it.

**Python — `mortie/coverage.py`**

8. The `offsets` parameter doc still says *"``offsets[0]`` need not be 0 (a sliced arrow array's offsets pass straight through)"*; the `Raises` section needs the endpoint/start errors added.

**Python tests — `mortie/tests/test_coverage_batch.py`**

9. `test_nonzero_start_offset_and_trailing_vertices` — the twin of (4); invert, and move the sliced-parity assertion to `test_coverage_batch_arrow.py` where the re-basing now lives.
10. `test_offsets_errors_name_polygon_index` matches three message fragments — `"monotonically"` for `[0, 6, 3]` (which under the strict contract *also* violates the endpoint, since `3 != 6`), `"exceeds"` for `[0, 3, 99]`, and `"non-negative"` for `[-1, 3]`. All three are order- and wording-sensitive to the change.
11. Everything else in the file is already exact-coverage (`_ragged` builds `offsets[-1] == len(lats)`; `test_empty_batch`, `test_single_polygon`, `test_short_ring_names_polygon_index`, `test_nan_names_polygon_index`, `test_lowest_index_reported_among_multiple_offenders`, `test_max_cells_floor_warns_once_naming_lowest_index` all satisfy the new contract unchanged).

**Arrow tests** — `test_sliced_input_nonzero_offsets` asserts `sliced.offsets[0].as_py() > 0` on the *input* and then expects success; that keeps working and becomes the regression pin for the re-basing, so it should stay (and is worth extending with a chunked-plus-sliced case and a middle slice, since `combine_chunks()` returns an offset-0 `ListArray` while a slice does not).

Finally, the PR body's Approach bullet and "Questions for review" item (2) still describe the loose contract as the implemented design — worth rewriting to record the ruling so the merged history reads correctly.

src_rust/src/coverage/batch.rs
```diff
@@ -0,0 +1,316 @@
...
+    if tolerance.is_some() && max_cells.is_some() {
+        return Err("pass at most one of tolerance / max_cells".to_string());
+    }
+    let n_polys = validate_batch(lats, lons, offsets, order)?;
+    let covers: Vec<Result<(Vec<u64>, usize), String>> = (0..n_polys)
+        .into_par_iter()
+        .map(|i| {
+            let (s, e) = (offsets[i] as usize, offsets[i + 1] as usize);
+            let (la, lo) = (&lats[s..e], &lons[s..e]);
+            catch_unwind(AssertUnwindSafe(|| {
```
espg: 🤖 *from Claude (review)*

[low] **The per-polygon panic capture — half the phase-2 "lowest-index error" guarantee — has no test, and I could not reach it from any input.**

Both determinism tests exercise only the serial pre-pass: `errors_name_lowest_index_polygon` (this file) and the Python `test_lowest_index_reported_among_multiple_offenders` feed short rings / non-monotone offsets / NaN, all of which `validate_batch` rejects *before* `par_iter` starts — so the ordering they pin is trivially guaranteed and the `catch_unwind` arm is never entered.

I went looking for a kernel panic that survives pre-validation and found none: `lat = 95`, `lat = -1000`, `lon = 1e12`, three identical vertices, three collinear vertices, and an antipodal edge pair all return normally from *both* `morton_coverage_moc` and the batch (checked with the offender at the high index and a valid polygon at index 0). The scalar kernels' own `assert!`s (`lats.len() >= 3`, length agreement, `1..=29` order) are exactly what `validate_batch` already screens.

So the claim in the module header —

> kernel panics are caught per polygon. In both regimes the error reported is the **lowest-index** offending polygon

— is asserted but unverified. Either (1) add a unit test that forces the arm (a `#[cfg(test)]` hook, or a poisoned polygon whose panic you can construct in-crate) so a future kernel change can't regress the index-order scan, or (2) soften the header to say structural errors are the tested guarantee and panic capture is defensive.

mortie/arrow.py
```diff
@@ -272,6 +272,140 @@ def to_morton_index(array):
...
+    def _plain(arr):
+        if isinstance(arr, pa.ChunkedArray):
+            arr = arr.combine_chunks()
+        if arr.null_count:
+            bad = int(np.flatnonzero(arr.is_null().to_numpy(zero_copy_only=False))[0])
+            raise ValueError(f"polygon {bad}: null polygon in batch")
+        return arr
+
+    def _f64(child):
+        return child.cast(pa.float64()).to_numpy(zero_copy_only=False)
```
espg, line 322: 🤖 *from Claude (review)*

[medium] **A sliced input ships the *entire* underlying child array to Rust, not the slice — and this is the exact site the strict-offsets ruling has to fix, in two halves.**

`arr.values` on a sliced `ListArray` returns the **full, unsliced** child (pyarrow semantics — verified): for `_struct_column().slice(2, 2)` on this branch, `sliced.offsets` is `[6, 10, 13]` while `len(sliced.values)` is still `13`. So `_f64` casts and converts the whole column's vertices, the wrapper `np.ascontiguousarray`s them, and `PyReadonlyArray1::to_vec()` in `rust_polygons_coverage_mocs` **copies all of them** — for a batch that may be one row:

```
200,000-row list<struct<lat,lon>> column, 800,000 vertices
  full column      -> 200000 MOCs   931.8 ms
  col.slice(0, 1)  ->      1 MOC      1.35 ms   (still copies 800k x 2 f64 = 12.8 MB)
  col.slice(n-1,1) ->      1 MOC      1.31 ms   (same)
```

At the issue's target shape (556k granules, ~50 vertices each) that is ~445 MB copied to cover a single-row slice. It is also a latent trap for the `cast` in `_f64`: on a `list<float32>` column the cast materializes the whole child, not the slice.

Under the ruling both halves are needed together, and getting only one wrong will be silent-then-loud:

```python
start, end = int(offsets[0]), int(offsets[-1])
verts = arr.values.slice(start, end - start)   # zero-copy view; also fixes the transfer above
offsets = offsets - start                      # re-based: offsets[0] == 0
```

Re-basing the offsets **without** slicing the values leaves `offsets[-1] != len(lats)`, which is precisely what the new core endpoint check will reject — so a skin that only subtracts `offsets[0]` will fail every sliced input. Slicing the values without re-basing fails the `offsets[0] == 0` check. Two notes that make this safe to do:

- Arrow list offsets are contiguous by construction (`off[i+1]` is both the end of row `i` and the start of row `i+1`), so `values[off[0]:off[-1]]` is exactly the region the listed rows use — no gaps to worry about, including around null rows (which are rejected before this point anyway).
- `pa.StructArray.field()` and `.to_numpy()` both honour a child's own offset — I checked a `ListArray.from_arrays(offsets, child.slice(3))` (child carrying its own offset) through the current skin and it produces the right per-polygon MOCs, so slicing here composes with that case rather than breaking it.

While you are in `_ragged_from_arrow`: the pair path compares raw offsets (`lat_list.offsets.equals(lon_list.offsets)`), so a pair where only one side is sliced is rejected even when the two are logically identical. Comparing the **re-based** offsets instead makes the check express the actual requirement.

mortie/coverage.py
```diff
@@ -364,6 +364,102 @@ def morton_coverage_moc(lats, lons, order=18, tolerance=None, max_cells=None,
     )
 
 
+def polygons_to_morton_mocs(lats, lons, offsets, order=18, tolerance=None,
```
espg, line 367: 🤖 *from Claude (review)*

[low] **Neither new public entry point will appear in the rendered docs — both mkdocstrings pages enumerate their members explicitly.**

`docs/api/coverage.md` and `docs/api/arrow.md` are member-gated, so a new public function is silently invisible on the docs site until it is listed:

```
::: mortie.coverage
    options:
      members:
        - morton_coverage
        - morton_coverage_moc
        - compress_moc
        ...
```

`polygons_to_morton_mocs` is exported from `mortie/__init__.py` and carries a full numpydoc docstring (which `numpydoc lint mortie/coverage.py mortie/arrow.py` passes — verified locally), but that docstring is never rendered. Same for `mortie.arrow.polygons_to_morton_mocs` in `docs/api/arrow.md`.

The narrative half is missing too — CLAUDE.md §4 asks for docstrings *and* narrative markdown:

- `USAGE.md` has a per-function API reference (`### morton_coverage_moc(lats, lons, order=18, tolerance=None, max_cells=None, normalize=True)` at line 319) with no batch sibling;
- `docs/coverage_methods.md`'s "which coverage call do I want" table (line 30) lists `morton_coverage` / `morton_coverage_moc` / the multipart form but not the many→many batch — which is exactly the table where the plural-vs-singular naming ruling on issue #153 needs to land for a reader to pick correctly.

One-line additions to the two `members:` lists plus a row in the `coverage_methods.md` table would close it.

src_rust/src/coverage/batch.rs
espg: 🤖 *from Claude*

Folded per the ruling (strict core, re-basing skin) — `2049261`. Walked the inventory site by site; all 11 landed:

**Rust (`src_rust/src/coverage/batch.rs`)**
1. `validate_batch`: `offsets[0] < 0` → `offsets[0] != 0` (*"offsets must start at 0, got -1"*), plus the new endpoint check placed **after** the per-polygon loop:
```rust
if offsets[n_polys] as usize != lats.len() {
    return Err(format!(
        "offsets must end at the vertex count: offsets[{n_polys}] is {} but \
         lats/lons have {} vertices",
        offsets[n_polys], lats.len()));
}
```
2. The `polygons_to_morton_mocs` doc now states exact coverage and points at the skin for slices.
3. `validate_batch`'s doc comment names the endpoint requirement and the check ordering.
4. `nonzero_start_offset` → `offsets_must_exactly_cover_the_vertices`: `[3, 7]` now asserts *"must start at 0"*, `[0, 3]` against 10 vertices asserts *"must end at the vertex count"*, and the re-based spelling of that middle quad is asserted to still match the scalar kernel.
5. Ordering hazard honoured — the endpoint check sits after the per-polygon loop, so `[0, 3, 99]` still fails as `polygon 1: … exceeds`; the test carries a comment saying why the order matters. `[0, 7, 3, 10]` and `[0, 3, 5, 5, 10]` untouched.
6. One message picked (*"must start at 0"*) and used in both spellings: `bad_order_and_layout_rejected` now asserts the string instead of bare `is_err()`, and the Python twin matches the same fragment.
7. `empty_batch` untouched (`0 == 0 == len`).

**Python** — (8) the `offsets` parameter doc states `offsets[0] == 0` / `offsets[-1] == len(lats) == len(lons)` and points at `mortie.arrow`; `Raises` gained the endpoint/start cases. (9) `test_nonzero_start_offset_and_trailing_vertices` → `test_offsets_must_exactly_cover_the_vertices` (both endpoints asserted, slice-parity moved to the Arrow tests, docstring says so). (10) all three fragments updated/verified: `"monotonically"` for `[0, 6, 3]` (per-polygon check still fires first), `"exceeds"` for `[0, 3, 99]`, and `"must start at 0"` for `[-1, 3]`. (11) untouched.

**Arrow tests** — `test_sliced_input_nonzero_offsets` kept as the re-basing pin and extended to a middle slice and an explicit "same MOCs as the unsliced equivalent" assertion; `test_chunked_and_sliced_input` and `test_sliced_pair_input` added (see the reply on the skin thread).

PR body Approach bullet and question (2) rewritten to record the ruling as implemented.

mortie/arrow.py
espg: 🤖 *from Claude*

Both halves folded together in `2049261`, exactly as prescribed — `_ragged_from_arrow` now re-bases the offsets **and** slices the value window in one helper, so neither can be applied without the other:

```python
def _rebased(arr):
    """Return the listed rows' value window (zero-copy), 0-based offsets."""
    offsets = arr.offsets.to_numpy(zero_copy_only=False).astype(np.int64)
    start, end = int(offsets[0]), int(offsets[-1])
    # List offsets are contiguous by construction, so [start, end) is
    # exactly the region the listed rows use.
    return arr.values.slice(start, end - start), offsets - start
```

Both paths go through it, so the struct spelling and the pair spelling behave identically. The pair check now compares the **re-based** offsets (`np.array_equal(offsets, lon_offsets)`) instead of the raw `.equals`, so a pair where only one side is sliced is accepted — the new `test_sliced_pair_input` asserts exactly that case, first asserting `not lats.slice(1).offsets.equals(lons_tail.offsets)` so the test would fail against the old raw comparison.

The transfer problem this thread opened with is fixed as a side effect: a one-row slice of a wide column now hands the core only that row's vertices, so `_f64`'s cast and `PyReadonlyArray1::to_vec()` no longer copy the whole child.

Tests added/extended (`mortie/tests/test_coverage_batch_arrow.py`):

- `test_sliced_input_nonzero_offsets` — now sweeps `slice(1)`, `slice(1, 1)` and `slice(2, 1)` (tail, middle, last), asserts `sliced.offsets[0] > 0` on each, checks scalar parity, and additionally asserts **identical MOCs to the unsliced equivalent** built from the same rings.
- `test_sliced_pair_input` — the paired path, including the one-side-only slice above.
- `test_chunked_and_sliced_input` — `pa.chunked_array([...]).slice(1)`, same-MOC assertion against the plain rings (`combine_chunks()` gives an offset-0 array, the slice does not — both land on the same result).

The public docstring now says a sliced input is re-based and only its own vertex window is passed on. Verified against the strict core: without the `values.slice` the endpoint check rejects every sliced input, which is what these tests would catch.

src_rust/src/coverage/batch.rs
espg: 🤖 *from Claude*

Fixed in `50ab086` — chunked assembly, peak now ~1.1x the result instead of ~2.3-2.5x.

The batch covers polygons `CHUNK` at a time and copies each chunk into the output as it lands, so the whole batch's per-polygon `Vec`s never coexist with the concatenated buffer. `assemble` is gone; `BatchMocs` grew `extend_chunk`, which consumes the chunk with `into_iter()` so each cover is **dropped** right after it is copied (that same in-index-order walk is still what makes the surfaced error the lowest-index one):

```rust
for base in (0..n_polys).step_by(CHUNK) {
    let end = (base + CHUNK).min(n_polys);
    let covers: Vec<Result<(Vec<u64>, usize), String>> = (base..end)
        .into_par_iter().map(|i| run_polygon(i, || { /* scalar kernel */ }))
        .collect();
    out.extend_chunk(covers, base, max_cells)?;
    out.reserve_estimate(end, n_polys - end);
}
```

`reserve_estimate` is the second half: without it the output `Vec` grows by doubling, and a realloc holds the old and new buffers at once — a ~1.5x-of-result transient that would have eaten most of the win. After each chunk it extrapolates the mean cover size over the polygons left (plus a 1/16 margin) and `reserve`s that, so late reallocs are rare; over-reserving is cheap because untouched pages cost address space, not RSS.

**Measurements** — your method (RSS sampled every 2 ms in a helper thread, peak above the pre-call baseline), each workload in its **own process** (measuring both in one process lets the allocator serve the second from the first's freed pages and understates it), `bd036d3` vs `50ab086`, same machine, same inputs:

| workload | returned `values` | peak before | peak after |
|---|---|---|---|
| 50k polygons, order 12, ~306 cells/polygon | 116.6 MB | 263.4 MB (**2.26x**) | 125.7 MB (**1.08x**) |
| 556k polygons, order 9, ~31 cells/polygon | 132.7 MB | 331.7 MB (**2.50x**) | 177.7 MB (**1.34x**) |

The row-2 residual is not the assembly: 556k x 4 vertices copied by `PyReadonlyArray1::to_vec()` is 35.6 MB, plus the 4.5 MB `int64` offsets vector — ~40 of the 45 MB. Row 1 (200k vertices, 3.2 MB of input) lands at 1.08x accordingly.

`CHUNK = 2048` came out of a sweep (100k order-8 footprints, `benchmarks/measure_batch_coverage.py`): 1024 → 0.50-0.54 s / 1.05x peak, **2048 → 0.48-0.49 s / 1.10x**, 8192 → 0.48-0.50 s / 1.23x. 2048 holds the pre-chunking throughput (0.47 s in the PR body, 0.48-0.49 s now) at essentially the best peak; the rationale is in the const's doc comment.

Also documented, since the catalog-scale caller still wants to size for it: a "Memory posture" section in the module header and a sentence in the Python wrapper's docstring.

src_rust/src/coverage/batch.rs
espg: 🤖 *from Claude*

Took option (1) — the Rust-side unit test with an injected panicking kernel — in `0c2c6cc`. Your probe result stands (no input reaches the arm), so the test drives the mechanism rather than pretending an input can panic.

The per-polygon body is now a named helper, which is what makes it injectable without a `#[cfg(test)]` hook in the hot path:

```rust
fn run_polygon<F>(i: usize, kernel: F) -> Result<(Vec<u64>, usize), String>
where
    F: FnOnce() -> (Vec<u64>, usize),
{
    catch_unwind(AssertUnwindSafe(kernel)).map_err(|e| {
        format!("polygon {i}: {}", crate::panic_msg(e, "polygon coverage panicked"))
    })
}
```

`polygons_to_morton_mocs` calls it with the scalar kernel; `kernel_panic_is_caught_and_named_by_lowest_index` calls it with panicking closures and pushes the results through the real assembly path (`BatchMocs::extend_chunk`), asserting: the ok/panic/panic case errors as `polygon 1: … kernel exploded` (lowest index wins over the polygon-2 panic, and the message is carried through, not swallowed), and a chunk based at 9 names `polygon 9:` — that last one is new coverage the old code could not have had, since chunk-relative indices are now a real failure mode. The panic hook is stubbed for the duration so the test output stays clean.

The module-header claim is therefore now tested on both halves, so I left the wording as is rather than softening it.

mortie/coverage.py
espg: 🤖 *from Claude*

All four sites folded in `f8435b8`:

- `docs/api/coverage.md` — `polygons_to_morton_mocs` added to `members:` after `morton_coverage_moc`.
- `docs/api/arrow.md` — `polygons_to_morton_mocs` added to `members:`.
- `USAGE.md` — a full `### polygons_to_morton_mocs(...)` entry in the API reference, matching the neighbouring style (parameters / returns / a short example), including the strict-offsets contract and a pointer to the Arrow spelling.
- `docs/coverage_methods.md` — a third row in the "which coverage call do I want" table, plus a short paragraph under it where the naming ruling lands: plural *MOCs* = many→many, one MOC per input polygon, against the many→**one** union of the list-of-rings form in the next section.

That paragraph also carries the by-design note that a batch entry is **one ring** — no multipart/hole spelling in the ragged layout, decompose such a footprint caller-side and cover it with `morton_coverage_moc`'s list-of-rings form. The same sentence is now in both docstrings (`mortie.polygons_to_morton_mocs`, `mortie.arrow.polygons_to_morton_mocs`), so it is stated outright wherever a reader lands.

`numpydoc lint mortie/coverage.py mortie/arrow.py` still clean.

Issue #177, opened 2026-08-09
toc coverage algebra: normalize / union / intersect / minus over toc-word sets (deferred until a consumer call site exists)


espg:
🤖 *from Claude* (filed at espg's direction, 2026-08-09, from the #175 design discussion)

## What this is

The **interval-set coverage algebra over toc words**: normalizing an array of toc words (see #175) into a canonical disjoint cover, and union / intersection / difference between such covers. Split out of #175 so the scalar type ships without it, per espg's ruling: the *elementwise* window predicates (`toc_overlaps`, `toc_contains` — "which words intersect query window W") moved **into** #175's phase-1 scope because their consumer shape is already known; the set algebra waits here **until the first consumer call site exists** — likely moczarr's temporal filter or zagg's revisit query.

## Why deferred: the data model has open questions only a consumer answers

Recording the three design questions surfaced in discussion so they are not re-derived later. The kernels are classic interval merging (and should share the range-walk machinery the spatial set ops use, cf. #173) — the *semantics* are what is undecided:

**1. The two-grid adjacency problem.** Range starts live on the 2^31 ns grid, ends on the 2^32 ns grid (#175's 32/31 split). "Adjacent intervals coalesce" is only cleanly definable on the coarser grid, so normalize needs an explicit adjacency rule the scalar type never faces — and the rule chosen changes which covers are canonical.

**2. Timestamp absorption is lossy — and consumers pull in opposite directions.** Does normalize absorb an exact-ns timestamp into an overlapping quantized range? If yes, canonical form destroys exactly the ns-exactness the dual type was designed to preserve (zagg#414 exists to keep leaves as exact timestamps). If no, what *is* canonical form for a mixed cover? A revisit-set consumer (zagg: per-cell crossing times) wants timestamps kept; a store-envelope consumer (moczarr sidecar) wants coalesced ranges. This is the question a real call site decides.

**3. The difference trap.** Toc words are conservative over-approximations. Intersection of conservative covers is still conservative (A ⊇ X, B ⊇ Y ⇒ A ∩ B ⊇ X ∩ Y) — but **difference is not**: subtracting an over-approximated B removes too much, so `toc_minus` on conservative covers can *under*-cover the true difference. Any minus needs explicit semantics ("conservative for what?") before it exists — the same class of correctness constraint the spatial compaction trap posed in #173 (design question 3 there), where the answer likewise had to come from a consumer's observed breakage.

## The precedent this follows

[...]

PR #178, opened 2026-08-09
toc word: temporal order coverage (issue #175)

### LINKED ISSUE #175: Temporal coverage: packed timestamp|range dual-type time word (design transfer from zagg#410)
🤖 *from Claude* (filed at espg's direction, 2026-08-08 — design transfer from englacial/zagg#410)

## What this is

mortie grows from spatial indexing to a **coverage library**: spatial (HEALPix/morton, existing) plus **temporal**. The temporal cell type was designed on [zagg#410](https://github.com/englacial/zagg/issues/410) — see the [superseding design comment](https://github.com/englacial/zagg/issues/410#issuecomment-5229923036) and the [T-MOC research](https://github.com/englacial/zagg/issues/410#issuecomment-5226588537) that ruled out the hierarchical-cell alternative — and lands here because:

1. **Dependency topology**: zagg writes the field, moczarr reads it, and mortie is the dependency both already share.
2. **Kernel synergy**: a mortie spatial MOC is already, internally, a sorted list of disjoint u64 ranges (the depth-29 linearization in `src_rust/src/moc.rs`); a temporal coverage is the same object over a 1-D axis, and the range-walk machinery (cf. the #173 `mocs_intersect` predicate) generalizes directly.
3. It fits mortie's constraints exactly: pure bit manipulation, zero new dependencies, numpy-only surface.

zagg#410 keeps the consumer side (the per-centroid companion array, its fold law, the config knob, normative spec text + fixtures). This issue is where the **type itself** gets planned and built.

## The type, as decided on zagg#410

One u64, a tagged union of an exact timestamp and a conservative time range:

| variant | layout (bit 0 = LSB) |
|---|---|
| **timestamp** | bits 63..1 = nanoseconds since epoch (2^63 ns ≈ 292 yr); bit 0 = `0` |
| **range** | bits 63..32 = start, **32 bits at 2^31 ns (≈2.15 s)**; bits 31..1 = end, **31 bits at 2^32 ns (≈4.29 s)**; bit 0 = `1` |

- **Bit split 32/31 — decided** (espg, 2026-08-08, on #410): spatial overview folds merge same-pass centroids constantly, so short-span ranges are the bulk of the range population, and their stored width is pure quantization slop — worst case `q_start + q_end` ≈ 6.4 s under 32/31 vs ≈ 9.7 s under the 33/30 alternative (typical: ~3.2 s vs ~4.8 s). The tighter end bound wins over the finer sort granularity.
- **Outward rounding**: start floored, end ceiled to their grids — a range is always a conservative superset of its members (the same honesty contract `common_ancestor` gives the spatial `location` companion).
- **Sort**: discriminator in the LSB, shared field high (the zuniq trick). A timestamp's top 32 bits are `ns >> 31` — the same 2^31 ns units as a range's start field — so plain unsigned sort orders timestamps and ranges chronologically by start, interleaved, no comparator.
- **Merge**: `min` on start codes, `max` on end codes, timestamps cast outward on first merge. A semilattice join on a fixed epoch-anchored lattice: exactly associative, commutative, idempotent — **bit-identical under any fold tree**, with error bounded at one quantum per end regardless of fold depth.
- **Epoch/timescale — decided**: nanoseconds since **1850-01-01T00:00:00** on a **continuous, leap-free, GPS-aligned scale** (equals GPS time + one fixed constant after 1980; proleptic before). Verified endpoints: CMIP historical starts 1850 (the binding constraint), NCEP/NCAR R1 starts 1948; ceiling ≈ 2142 covers standard ScenarioMIP to 2100. Documented exclusions: CMIP extended-to-2300 scenarios, pre-1850 reanalyses. Leap seconds appear only at the UTC boundary (datetime64/pandas import-export), via the static leap table.
- **Rejected alternatives** (recorded on #410 with the kill arguments): hierarchical T-MOC-style cell (unbounded merge coarsening, measured), floating-point absolute time (precision inverted — worst on newest data), unit-scale fields and calendar tuples (sort/path-independence/radix-waste failures). Held in reserve: uniform start + float duration (finer everywhere, but ULP-level merge path-dependence).

## Open questions

1. **Base quantum: 1 ns vs 2 ns.** 2 ns doubles the span to ~585 yr (1850→~2434), covering CMIP extended scenarios, at zero physical cost for ICESat-2 (float64 `delta_time` carries only ~50 ns at mission magnitudes). Lean: plain ns, exclusions documented; flip only if 2300-scenario output is declared in scope. Must be settled before the epoch constant freezes.
2. **Module and naming.** A new `mortie/temporal.py` (mirrored by `src_rust/src/temporal.rs`)? Names for the type's operations (`time_encode`/`time_merge`/…? a `TimeWord` framing?) — should read as one family with the morton word ops.
3. **Phase-1 scope.** Scalar encode/decode/merge + vectorized numpy batch forms (encode from `datetime64[ns]`, merge two arrays, reduce an array), UTC↔internal conversion helpers with the leap table. **Interval-set coverage algebra (union/intersect over sorted arrays of these) is explicitly deferred** — espg parked it pending the scalar type; when it lands it should share the range-walk kernels with the spatial set ops.
4. **Conformance surface.** Which invariants are normative and fixture-pinned: sort-order property, semilattice laws (associativity/commutativity/idempotence), outward-conservatism, round-trip bounds, the epoch/scale constants.

## Sequencing

Blocked on nothing in mortie; gated only on the tail of the zagg#410 discussion (the zagg-side fold law and spec text develop in parallel there). The batch entry points, when they come, land in `mortie/batch.py` per #170.

Refs: englacial/zagg#410 (design record), englacial/zagg#414 (the compression-factor change the leaf-precision story depends on), #173 (the range-walk kernels this shares), #170 (`batch.py`).

espg:
Closes #175

## What this is

The **toc word** — temporal order coverage: one `u64`, a tagged union of an exact nanosecond **timestamp** and a quantized, conservative **time range**, sortable as a plain unsigned integer and closed under a semilattice merge. Design transferred from englacial/zagg#410; every layout decision follows the #175 ledger exactly:

- 1 ns base quantum, 32/31 split ([decision](https://github.com/espg/mortie/issues/175#issuecomment-5230004941))
- flag at **bit 31** (1 = timestamp, 0 = range), `toc` naming, and the normative line ([amendments](https://github.com/espg/mortie/issues/175#issuecomment-5230049902))
- window predicates in scope, interval-set algebra deferred to #177 ([scope](https://github.com/espg/mortie/issues/175#issuecomment-5230084582))
- flat-array ops in `mortie/toc.py`, mirroring the moc pattern; `batch.py` untouched ([ruling](https://github.com/espg/mortie/issues/175#issuecomment-5230097527))

Layout, restated:

```
[start: 32 bits, 2^31 ns units][flag: 1 bit][low: 31 bits]
  timestamp: word = t_ns with a 1 flag bit spliced in at position 31
  range:     low = end code, 31 bits at 2^32 ns; envelope [s*2^31, e*2^32) half-open
```

Outward rounding uses a **strictly-greater end ceiling**: `e = (b >> 32) + 1` uniformly, including when `b` sits exactly on the 2^32 grid, so the envelope always properly contains `b`. Consequence: internal times are range-checked against `TOC_MAX_NS = 2^63 - 2^32` (the end code must fit 31 bits, ~4 s short of the year-2142 span ceiling) — applied to *both* encoders so every encodable word is mergeable. Merge returns bitwise-equal inputs unchanged (idempotence: two equal timestamps must not become their range envelope); any other pair merges conservative envelopes into a range word.

## Phases

- [x] Phase 1 — Rust kernel `src_rust/src/toc.rs`: layout constants, scalar `encode_timestamp` / `encode_range` / `decode` / `is_range` / `merge`; cargo property tests (semilattice laws incl. `merge(w,w) = w` for timestamps, fold-tree independence, sort monotonicity + tied-quantum tiebreaks, containment/conservatism incl. the on-grid ceiling edge, width bound, roundtrip, top-of-span rejection) and golden u64 fixtures.
- [x] Phase 2 — pyfunctions + `mortie/toc.py`: `time2toc`, `span2toc`, `toc2time`, `toc_merge`, `toc_reduce`, `toc_is_range`, `toc_overlaps`, `toc_contains`; exports; pytest parity + mirrored property suite.
- [x] Phase 3 — timescale boundary: `from_datetime64` / `to_datetime64` (static leap table) and `from_gps_ns` / `to_gps_ns` (constant offset), with pinned boundary tests.
- [x] Phase 4 — docs: `docs/api/toc.md` + mkdocs nav; numpydoc on every public function.

## Golden fixtures (normative)

| fixture | word |
|---|---|
| epoch instant 1850-01-01T00:00:00 (t = 0) | `0x80000000` = 2147483648 |
| 2018-01-01T00:00:00 UTC = internal ns 5,301,590,418,000,000,000 (61,361 days × 86,400 s + 18 s GPS−UTC; cross-checked against GPS second 1,198,800,018) | `0x932610CAE9813400` = 10603180836377408512 |
| range [3·2^32 − 5e8, 3·2^32 + 5e8] straddling a 2^32 grid line (s = 5, e = 4) | `0x500000004` = 21474836484 |
| range ending exactly on the grid at 7·2^32 (e = 8, strictly greater) | `(13 << 32) \| 8` = 55834574856 |

## Testing

All phases green locally: `maturin develop --release`, `cargo test --lib` (330 tests, the 13 new toc tests included), `cargo fmt --check`, `cargo clippy` (no new warnings; the existing ones predate this branch), `flake8 mortie --select=E9,F63,F7,F82`, `pytest -v` (30 new tests in `mortie/tests/test_toc.py`, full suite passing), `numpydoc lint mortie/*.py`. One pre-existing environment quirk, flagged rather than fixed: `cargo test`'s *doctest* phase fails to link the cdylib on this macOS setup (plain cargo lacks maturin's `-undefined dynamic_lookup` link args) — verified identical on `origin/main`, and the repo has no Rust doctests, so `--lib` is the entire suite. The Rust property tests use a hand-rolled splitmix64 (no new dependency); the pytest suite mirrors them at array level and adds the timescale-boundary pins (GPS-epoch identity, the 2016-12-31 leap second, modern-era datetime64 roundtrips through a toc word).

## Questions for review

- **Pre-1972 UTC convention (phase 3)**: the plan is **zero offset before 1972** (naive proleptic day-count seconds, no leap adjustment), rather than freezing the 1972 offset of −9 s. Rationale: it pins the epoch identity exactly (`from_datetime64('1850-01-01') == 0`), and a frozen −9 s would push the epoch itself to a negative internal time, which u64 cannot hold. Cost: a 9 s backward step across the 1972-01-01 boundary, so the last 9 SI seconds of 1971 alias early 1972 and are not invertible; conversion is exact and invertible from 1972 on. Documented in the docstring. Historical data carries hours-level precision, so stakes are low — flagged here per the ledger so the choice is explicit.
- **Timestamp ceiling**: `TOC_MAX_NS` is enforced on `encode_timestamp` too (not only on range ends), so that merge is total over encodable words — a timestamp in the final 2^32 ns of the span would otherwise overflow the end field on first merge. This forfeits the last ~4.3 s before the 2142 ceiling for timestamps as well; flagging since the ledger's range-check wording named the range encoder only.
- **Phase-3 spellings** (`from_datetime64` / `to_datetime64` / `from_gps_ns` / `to_gps_ns`, exported flat) settle in review per the [naming amendment](https://github.com/espg/mortie/issues/175#issuecomment-5230049902).
- The example notebook lands as a follow-up, not in this PR ([ruling](https://github.com/espg/mortie/issues/175#issuecomment-5230097527)).

### DIFF
--- a/docs/api/toc.md
+++ b/docs/api/toc.md
@@ -1,0 +1,28 @@
+# mortie.toc
+
+The toc word — temporal order coverage (issue #175): one `uint64` packing
+either an exact nanosecond timestamp or a conservative time range, sortable
+as a plain unsigned integer and closed under a semilattice merge. Times are
+ns since 1850-01-01 on a continuous, leap-free, GPS-aligned scale; the
+`datetime64` / GPS converters are the only place leap seconds exist. Not an
+IVOA T-MOC. These flat-array elementwise ops are the type's scalar surface
+(the same relationship [mortie.moc](moc.md) has to its ops over one cover);
+the ragged many-cover plurals land in [mortie.batch](batch.md) when the
+interval-set algebra (issue #177) activates. The names stay flat on the
+package (`mortie.time2toc`, ...).
+
+::: mortie.toc
+    options:
+      members:
+        - time2toc
+        - span2toc
+        - toc2time
+        - toc_merge
+        - toc_reduce
+        - toc_is_range
+        - toc_overlaps
+        - toc_contains
+        - from_datetime64
+        - to_datetime64
+        - from_gps_ns
+        - to_gps_ns
--- a/docs/api/toc.md
+++ b/docs/api/toc.md
@@ -26,3 +26,7 @@ package (`mortie.time2toc`, ...).
         - to_datetime64
         - from_gps_ns
         - to_gps_ns
+        - Q_START_NS
+        - Q_END_NS
+        - TOC_MAX_NS
+        - GPS_EPOCH_NS
--- a/mortie/tests/test_toc.py
+++ b/mortie/tests/test_toc.py
@@ -12,11 +12,16 @@ import numpy as np
 import pytest
 
 from mortie.toc import (
+    GPS_EPOCH_NS,
     Q_END_NS,
     Q_START_NS,
     TOC_MAX_NS,
+    from_datetime64,
+    from_gps_ns,
     span2toc,
     time2toc,
+    to_datetime64,
+    to_gps_ns,
     toc2time,
     toc_contains,
     toc_is_range,
@@ -295,3 +300,83 @@ def test_dtype_and_domain_validation():
     # The last valid instant is fine for both encoders.
     assert time2toc(TOC_MAX_NS - 1) > 0
     assert span2toc(0, TOC_MAX_NS - 1) > 0
+
+
+# ── timescale boundary (phase 3) ────────────────────────────────────────
+
+
+def test_gps_epoch_constant_against_date_arithmetic():
+    # The 1850 -> GPS-epoch constant is a plain proleptic-Gregorian day
+    # count (the internal scale is leap-free and ticks with GPS).
+    from datetime import date
+    days = (date(1980, 1, 6) - date(1850, 1, 1)).days
+    assert days == 47_486
+    assert GPS_EPOCH_NS == days * 86_400 * 10**9
+
+
+def test_gps_epoch_zero_offset_identity():
+    # At the GPS epoch TAI - UTC = 19, so GPS - UTC = 0: the UTC and GPS
+    # entry points agree exactly there.
+    assert from_datetime64("1980-01-06") == GPS_EPOCH_NS
+    assert from_gps_ns(0) == GPS_EPOCH_NS
+    assert to_gps_ns(GPS_EPOCH_NS) == 0
+    assert to_datetime64(GPS_EPOCH_NS) == np.datetime64("1980-01-06", "ns")
+
+
+def test_epoch_identity_and_pre1972_convention():
+    # Zero offset before 1972 pins the epoch identity exactly.
+    assert from_datetime64("1850-01-01") == 0
+    assert to_datetime64(0) == np.datetime64("1850-01-01", "ns")
+    with pytest.raises(ValueError, match="before the 1850"):
+        from_datetime64("1849-12-31T23:59:59")
+
+
+def test_gps_conversion_exact_roundtrip():
+    rng = np.random.default_rng(53)
+    gps = rng.integers(0, TOC_MAX_NS - GPS_EPOCH_NS, size=500,
+                       dtype=np.uint64)
+    np.testing.assert_array_equal(to_gps_ns(from_gps_ns(gps)), gps)
+    with pytest.raises(ValueError, match="before the GPS epoch"):
+        to_gps_ns(GPS_EPOCH_NS - 1)
+
+
+def test_2018_utc_conversion_matches_golden():
+    # The same instant the golden timestamp fixture pins: 61,362 days of
+    # 86,400 s past 1850, +18 s GPS-UTC.
+    t = from_datetime64("2018-01-01")
+    assert t == 5_301_590_418_000_000_000
+    assert to_gps_ns(t) == 1_198_800_018 * 10**9  # published GPS seconds
+
+
+def test_leap_second_boundary_pinned():
+    # A leap second was inserted at the end of 2016-12-31: one UTC second
+    # across the boundary spans two internal seconds.
+    before = from_datetime64("2016-12-31T23:59:59")
+    after = from_datetime64("2017-01-01T00:00:00")
+    assert after - before == 2 * 10**9
+    # An instant inside the inserted leap second renders into the
+    # following UTC second (datetime64 cannot express 23:59:60).
+    inside = before + int(1.5e9)
+    assert to_datetime64(inside) == np.datetime64(
+        "2017-01-01T00:00:00.500000000")
+
+
+def test_datetime64_roundtrip_exact_modern_era():
+    rng = np.random.default_rng(59)
+    lo = np.datetime64("1972-01-01", "ns").astype(np.int64)
+    hi = np.datetime64("2100-01-01", "ns").astype(np.int64)
+    naive = rng.integers(lo, hi, size=2000, dtype=np.int64)
+    dt = naive.astype("datetime64[ns]")
+    np.testing.assert_array_equal(to_datetime64(from_datetime64(dt)), dt)
+    # ... and through a toc timestamp word.
+    t = from_datetime64(dt)
+    starts, _ = toc2time(time2toc(t))
+    np.testing.assert_array_equal(to_datetime64(starts), dt)
+
+
+def test_datetime64_scalar_forms():
+    t = from_datetime64(np.datetime64("2020-05-17T12:34:56.789"))
+    assert isinstance(t, int)
+    assert isinstance(to_datetime64(t), np.datetime64)
+    with pytest.raises(ValueError, match="ceiling"):
+        from_datetime64("2150-01-01")
--- a/mortie/tests/test_toc.py
+++ b/mortie/tests/test_toc.py
@@ -67,7 +67,7 @@ def test_golden_epoch_timestamp():
 
 
 def test_golden_2018_timestamp():
-    # 2018-01-01T00:00:00 UTC = 61,362 proleptic-Gregorian days of 86,400 s
+    # 2018-01-01T00:00:00 UTC = 61,361 proleptic-Gregorian days of 86,400 s
     # past 1850-01-01, plus GPS-UTC = 18 s: 5,301,590,418e9 ns internal.
     t = 5_301_590_418_000_000_000
     assert time2toc(t) == 10_603_180_836_377_408_512
@@ -254,6 +254,16 @@ def test_inverted_window_raises():
         toc_overlaps(time2toc(0), 10, 5)
 
 
+def test_empty_window_matches_nothing():
+    # An empty window intersects and contains nothing -- for both
+    # variants, even when it sits strictly inside a range's envelope.
+    q = 5 * Q_END_NS
+    words = np.array([time2toc(q), span2toc(q - Q_END_NS, q + Q_END_NS)],
+                     dtype=np.uint64)
+    assert not toc_overlaps(words, q, q).any()
+    assert not toc_contains(words, q, q).any()
+
+
 # ── validation, shapes, and edges ───────────────────────────────────────
 
 
@@ -341,7 +351,7 @@ def test_gps_conversion_exact_roundtrip():
 
 
 def test_2018_utc_conversion_matches_golden():
-    # The same instant the golden timestamp fixture pins: 61,362 days of
+    # The same instant the golden timestamp fixture pins: 61,361 days of
     # 86,400 s past 1850, +18 s GPS-UTC.
     t = from_datetime64("2018-01-01")
     assert t == 5_301_590_418_000_000_000
@@ -380,3 +390,44 @@ def test_datetime64_scalar_forms():
     assert isinstance(to_datetime64(t), np.datetime64)
     with pytest.raises(ValueError, match="ceiling"):
         from_datetime64("2150-01-01")
+    # A 0-d ndarray classifies as array, matching the word ops.
+    out = from_datetime64(np.array("2020-05-17", dtype="datetime64[ns]"))
+    assert isinstance(out, np.ndarray) and out.shape == (1,)
+
+
+def test_pre1972_zero_offset_and_1972_step():
+    from datetime import date
+
+    # Zero offset before 1972: exactly the proleptic day count.
+    days = (date(1960, 1, 1) - date(1850, 1, 1)).days
+    assert from_datetime64("1960-01-01") == days * 86_400 * 10**9
+    # The 9 s backward step across 1972-01-01: the last 9 SI seconds of
+    # 1971 alias early 1972 ...
+    assert (from_datetime64("1971-12-31T23:59:55")
+            == from_datetime64("1972-01-01T00:00:04"))
+    assert (from_datetime64("1972-01-01")
+            == from_datetime64("1971-12-31T23:59:51"))
+    # ... and aliased instants render on the 1972 side of the step.
+    assert to_datetime64(from_datetime64("1971-12-31T23:59:55")) == \
+        np.datetime64("1972-01-01T00:00:04", "ns")
+    # From 1972 on the mapping is invertible (the roundtrip test sweeps
+    # the modern era; this pins the very first invertible instant).
+    assert to_datetime64(from_datetime64("1972-01-01")) == \
+        np.datetime64("1972-01-01", "ns")
+
+
+def test_timescale_empty_arrays():
+    assert from_datetime64(np.array([], dtype="datetime64[ns]")).shape == (0,)
+    empty = np.empty(0, dtype=np.uint64)
+    assert to_datetime64(empty).shape == (0,)
+    assert from_gps_ns(empty).shape == (0,)
+    assert to_gps_ns(empty).shape == (0,)
+
+
+def test_timescale_domain_errors():
+    with pytest.raises(ValueError, match="2\\*\\*63"):
+        to_datetime64(1 << 63)
+    with pytest.raises(ValueError, match="ceiling"):
+        from_gps_ns(TOC_MAX_NS)
+    with pytest.raises(ValueError, match="q_start_ns"):
+        toc_overlaps(time2toc(0), -1, 5)
--- a/mortie/tests/test_toc.py
+++ b/mortie/tests/test_toc.py
@@ -1,0 +1,297 @@
+"""Tests for the toc word (temporal order coverage, issue #175).
+
+The golden u64 fixtures here are normative: words persist on disk, so the
+bit layout, constants, merge results, and sort order they pin must never
+change.  The property tests mirror the cargo suite in src_rust/src/toc.rs
+at array level.
+"""
+
+from functools import reduce
+
+import numpy as np
+import pytest
+
+from mortie.toc import (
+    Q_END_NS,
+    Q_START_NS,
+    TOC_MAX_NS,
+    span2toc,
+    time2toc,
+    toc2time,
+    toc_contains,
+    toc_is_range,
+    toc_merge,
+    toc_overlaps,
+    toc_reduce,
+)
+
+FLAG = 1 << 31
+LOW_MASK = FLAG - 1
+
+
+def py_timestamp(t):
+    """Reference splice: t_ns with a 1 flag bit inserted at position 31."""
+    return ((t >> 31) << 32) | FLAG | (t & LOW_MASK)
+
+
+def py_range(a, b):
+    """Reference range encode: floored start, strictly-greater end ceiling."""
+    return ((a >> 31) << 32) | ((b >> 32) + 1)
+
+
+def rand_times(rng, n):
+    """Random valid internal times as uint64."""
+    return rng.integers(0, TOC_MAX_NS, size=n, dtype=np.uint64)
+
+
+def rand_words(rng, n):
+    """A mixed batch of valid timestamp and range words."""
+    t = time2toc(rand_times(rng, n))
+    x, y = rand_times(rng, n), rand_times(rng, n)
+    r = span2toc(np.minimum(x, y), np.maximum(x, y))
+    pick = rng.integers(0, 2, size=n).astype(bool)
+    return np.where(pick, t, r)
+
+
+# ── golden fixtures (normative) ─────────────────────────────────────────
+
+
+def test_golden_epoch_timestamp():
+    # 1850-01-01T00:00:00 (t = 0): only the flag bit set.
+    assert time2toc(0) == 0x8000_0000
+
+
+def test_golden_2018_timestamp():
+    # 2018-01-01T00:00:00 UTC = 61,362 proleptic-Gregorian days of 86,400 s
+    # past 1850-01-01, plus GPS-UTC = 18 s: 5,301,590,418e9 ns internal.
+    t = 5_301_590_418_000_000_000
+    assert time2toc(t) == 10_603_180_836_377_408_512
+    assert toc2time(10_603_180_836_377_408_512) == (t, t)
+
+
+def test_golden_range_fixtures():
+    # Straddling the 2^32 grid line at 3*2^32: s = 5, e = 4.
+    assert span2toc(3 * 2**32 - 500_000_000,
+                    3 * 2**32 + 500_000_000) == 21_474_836_484
+    # End exactly on the grid at 7*2^32: e = 8, strictly greater.
+    w = span2toc(7 * 2**32 - 10**9, 7 * 2**32)
+    assert w == (13 << 32) | 8
+    _, end = toc2time(w)
+    assert end == 8 * 2**32 > 7 * 2**32
+
+
+# ── parity with the Rust scalars ────────────────────────────────────────
+
+
+def test_encode_parity_with_reference_splice():
+    rng = np.random.default_rng(175)
+    times = rand_times(rng, 500)
+    words = time2toc(times)
+    assert words.dtype == np.uint64
+    for t, w in zip(times.tolist(), words.tolist()):
+        assert w == py_timestamp(t)
+    x, y = rand_times(rng, 500), rand_times(rng, 500)
+    a, b = np.minimum(x, y), np.maximum(x, y)
+    words = span2toc(a, b)
+    for aa, bb, w in zip(a.tolist(), b.tolist(), words.tolist()):
+        assert w == py_range(aa, bb)
+
+
+def test_timestamp_roundtrip_exact():
+    rng = np.random.default_rng(17)
+    times = rand_times(rng, 1000)
+    starts, ends = toc2time(time2toc(times))
+    np.testing.assert_array_equal(starts, times)
+    np.testing.assert_array_equal(ends, times)
+    assert not toc_is_range(time2toc(times)).any()
+
+
+def test_envelope_conservatism_and_width():
+    rng = np.random.default_rng(23)
+    x, y = rand_times(rng, 1000), rand_times(rng, 1000)
+    a, b = np.minimum(x, y), np.maximum(x, y)
+    # Force some ends exactly onto the 2^32 grid (the strictly-greater
+    # ceiling edge); keep the pairs ordered.
+    b[::8] = (b[::8] >> np.uint64(32)) << np.uint64(32)
+    a = np.minimum(a, b)
+    starts, ends = toc2time(span2toc(a, b))
+    assert (starts <= a).all(), "start floors"
+    assert (b < ends).all(), "end strictly greater, half-open envelope"
+    assert (ends - starts <= (b - a) + np.uint64(Q_START_NS + Q_END_NS)).all()
+    assert toc_is_range(span2toc(a, b)).all()
+
+
+# ── merge laws (mirroring the cargo property suite) ─────────────────────
+
+
+def test_merge_idempotent_including_timestamps():
+    rng = np.random.default_rng(29)
+    w = rand_words(rng, 1000)
+    np.testing.assert_array_equal(toc_merge(w, w), w)
+    # The load-bearing case: equal timestamps must stay timestamps.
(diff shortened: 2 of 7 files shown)


mortie/toc.py
```diff
@@ -0,0 +1,639 @@
...
+
+    See Also
+    --------
+    to_datetime64 : the inverse conversion.
+    from_gps_ns : the leap-free GPS entry point.
+    """
+    is_scalar = np.ndim(when) == 0
+    naive = np.atleast_1d(
+        np.asarray(when, dtype="datetime64[ns]")).astype(np.int64)
+    if naive.size and int(naive.max()) >= TOC_MAX_NS - _EPOCH_1850_1970_NS:
```
espg: 🤖 *from Claude (review)*

**`from_datetime64`'s span-ceiling guard is off by the leap offset — it checks the *naive* value against the *internal* ceiling.**

The guard compares `naive` (ns since 1970, no leap offset) against `TOC_MAX_NS - _EPOCH_1850_1970_NS`, but line 509 then adds `_OFFSETS_NS[idx]` (+18 s in the modern era) on top. So the guard is short by the offset, and two things fall through the ~18 s gap at the top of the span:

**(a) Values at or beyond `TOC_MAX_NS` are returned silently**, contradicting this function's own `Raises` section ("at or beyond `TOC_MAX_NS` (the toc span ceiling, ~year 2142)"):

```python
>>> n = mortie.toc.TOC_MAX_NS - mortie.toc._EPOCH_1850_1970_NS - 18*10**9
>>> v = from_datetime64(np.datetime64(int(n), 'ns'))   # 2142-04-11T23:46:54.559808512
>>> v, v >= mortie.toc.TOC_MAX_NS
(9223372032559808512, True)
>>> time2toc(v)
ValueError: timestamp 9223372032559808512 ns is at or beyond the toc span ceiling
```

The boundary function hands back an un-encodable internal ns; the failure only surfaces one call later.

**(b) A ~4.3 s sub-window overflows the int64 accumulator** and reports the *opposite* error:

```python
>>> n2 = (1<<63) - 1 - mortie.toc._EPOCH_1850_1970_NS - 18*10**9 + 1
>>> from_datetime64(np.datetime64(int(n2), 'ns'))
ValueError: datetime64 input is before the 1850-01-01 epoch
```

`naive + _EPOCH_1850_1970_NS + _OFFSETS_NS[idx]` wraps past `2**63-1` into the negatives, and the `internal.min() < 0` check on line 510 catches the wreckage with a message that points a user at 1850 for a year-2142 input.

Suggested fix: do the ceiling check on `internal` (after line 509) rather than on `naive`, and keep a cheap pre-check on `naive` against `TOC_MAX_NS - _EPOCH_1850_1970_NS - _OFFSETS_NS.max()` so the addition itself can never overflow int64. A test at `TOC_MAX_NS - _EPOCH_1850_1970_NS - 18e9` would pin both directions; nothing in `test_toc.py` currently probes closer than `"2150-01-01"`, which is far enough past the ceiling to clear the buggy guard.

src_rust/src/toc.rs
```diff
@@ -0,0 +1,616 @@
...
+        // The epoch instant 1850-01-01T00:00:00 (t = 0 ns): only the flag
+        // bit is set.
+        assert_eq!(encode_timestamp(0).unwrap(), 0x8000_0000);
+    }
+
+    #[test]
+    fn golden_2018_timestamp() {
+        // 2018-01-01T00:00:00 UTC as internal ns, derived:
+        //   proleptic-Gregorian days 1850-01-01 -> 2018-01-01 = 61,362
+        //   (168 years * 365 + 41 leap days: 1852..2016 has 42 multiples of
```
espg, line 404: 🤖 *from Claude (review)*

**The 2018 fixture's day count is off by one: 1850-01-01 → 2018-01-01 is 61,361 days, not 61,362.**

The pinned word and the pinned ns are *correct* — it's the derivation comment that is wrong, and it is wrong in a way that contradicts its own arithmetic two lines later:

```
168 years * 365 + 41 leap days  ->  61320 + 41 = 61361
```

but the comment states `= 61,362`. Cross-checks, both of which agree with **61,361**:

```python
>>> from datetime import date
>>> (date(2018,1,1) - date(1850,1,1)).days
61361
>>> 61361 * 86400 * 10**9 + 18 * 10**9
5301590418000000000        # == the pinned t
>>> 61362 * 86400 * 10**9 + 18 * 10**9
5301676818000000000        # != the pinned t (off by 86,400 s)
```

The GPS cross-check in the same comment also lands on 61,361: `47,486 * 86,400 + 1,198,800,018 = 5,301,590,418` s.

This matters more than a typo because the comment is the audit trail for a *normative* fixture — anyone re-deriving the golden word from the stated day count gets a different number and will conclude the fixture is broken. Same off-by-one appears in two more places and should be fixed with it:

- `mortie/tests/test_toc.py` line 70 — "61,362 proleptic-Gregorian days"
- `mortie/tests/test_toc.py` line 344 — "61,362 days of 86,400 s past 1850"
- the PR body's golden-fixtures table ("61,362 days × 86,400 s + 18 s GPS−UTC")

Worth pinning mechanically too: `test_gps_epoch_constant_against_date_arithmetic` already asserts the 47,486 constant against `datetime.date`; an equivalent `assert (date(2018,1,1) - date(1850,1,1)).days == 61_361` next to the 2018 fixture would keep the prose and the number from drifting apart again.

mortie/toc.py
```diff
@@ -0,0 +1,639 @@
...
+    ValueError
+        If any value is negative, non-integer-typed, or at or beyond
+        ``2**63`` (past every representable envelope bound).
+
+    See Also
+    --------
+    from_datetime64 : the inverse conversion.
+    to_gps_ns : the leap-free GPS exit point.
+    """
+    is_scalar = np.isscalar(t_ns)
```
espg, line 563: 🤖 *from Claude (review)*

**Two different scalar-detection conventions across this module, so a 0-d ndarray round-trips into a shape-`(1,)` array on one side and an `int` on the other.**

`from_datetime64` uses `np.ndim(when) == 0` (line 501); every other public function in the module uses `np.isscalar(...)`. `np.isscalar` is `False` for 0-d arrays and 0-d numpy scalars produced by indexing, while `_as_u64`/`np.atleast_1d` has already inflated the shape to `(1,)` — so the "array path" returns a `(1,)` array where numpy convention (and the sibling function) would give a scalar:

```python
>>> from_datetime64(np.array(np.datetime64('2018-01-01')))     # np.ndim path
5301590418000000000                                            # int

>>> to_datetime64(np.array(0, dtype=np.uint64))                # np.isscalar path
array(['1850-01-01T00:00:00.000000000'], dtype='datetime64[ns]')   # shape (1,)

>>> time2toc(np.array(5, dtype=np.uint64)).shape
(1,)
>>> toc_is_range(np.array(time2toc(5), dtype=np.uint64)).shape
(1,)
```

So `to_datetime64(from_datetime64(x))` is not shape-preserving for a 0-d input, and every `words.reshape(t.shape)` in the module reshapes to `(1,)` rather than `()` because `_as_u64` applies `atleast_1d` before the shape is captured. The docstrings all promise "scalar in -> `int` out", which a 0-d array reasonably reads as.

Cheapest consistent fix: capture `np.ndim(x) == 0` (which is `True` for Python ints, numpy scalars, *and* 0-d arrays) before `_as_u64`, and use it everywhere — that makes all twelve functions agree and makes 0-d in → scalar out. If instead the intent is that only Python/numpy scalars collapse and 0-d arrays stay arrays, then `from_datetime64` is the odd one out and the 0-d array case should be pinned by a test either way; `test_scalar_in_scalar_out` currently only covers Python-`int` inputs.

src_rust/src/toc.rs
```diff
@@ -0,0 +1,616 @@
...
+/// A range carries its codes verbatim.  A timestamp's envelope for merging
+/// purposes is `s = t >> 31` (= `word >> 32`, the shared high field) and
+/// `e = (t >> 32) + 1` (= `(word >> 33) + 1` — bit 31 of `t` is bit 0 of
+/// the high field, so `t >> 32` is one more shift of it).
+#[inline]
+fn codes(word: u64) -> (u64, u64) {
+    if is_range(word) {
+        (word >> 32, word & LOW_MASK)
+    } else {
+        (word >> 32, (word >> 33) + 1)
```
espg, line 134: 🤖 *from Claude (review)*

**`merge` can emit a word with the flag bit *set* — i.e. a "timestamp" produced by merging two unequal words — when a timestamp word is out of the encoder's domain.**

The timestamp branch computes `e = (word >> 33) + 1`. For an encoder-produced word that is safe (`t < TOC_MAX_NS` ⇒ `t >> 32 ≤ 2^31 - 2` ⇒ `e ≤ 2^31 - 1 = LOW_MASK`), which is exactly why `TOC_MAX_NS` is enforced on `encode_timestamp` — the PR body flags this as deliberate. But the guard lives only in the *encoders*; `merge` (and the whole Python surface) accepts any `u64`. When `word >> 33 == 0x7FFF_FFFF`, `e` becomes `2^31 == FLAG_MASK`, and line 153's `| ea.max(eb)` splices it straight into the flag position:

```python
>>> toc_merge(2**64 - 1, 0)
2147483648                 # 0x80000000
>>> toc_is_range(2147483648)
False                      # a merge of two *unequal* words returned a timestamp
>>> toc2time(2147483648)
(0, 0)                     # ... and it decodes as the 1850 epoch instant
```

Both the Rust doc-comment above `merge` ("Any other pair decodes to conservative envelopes and emits a range word") and the `toc_merge` docstring in `mortie/toc.py` ("Any other pair merges conservative envelopes ... into a range word") are false for this input, and the result is not merely imprecise — it is the *epoch timestamp*, the least possible word, so a single foreign word in a column silently collapses a whole `toc_reduce` to `0x80000000`.

`_as_u64` validates dtype and sign but nothing about the toc domain, so the only thing standing between a persisted/foreign/corrupt word and this outcome is trust. Two reasonable resolutions, both cheap:

1. Make `merge` total by saturating: `let e = ((word >> 33) + 1).min(LOW_MASK);` — keeps merge closed over *all* `u64` at the cost of a clamp on inputs that were already invalid.
2. Stand down explicitly: state in both docstrings that `merge`/`decode` are defined only over encoder-produced words (the `decode` comment already says "garbage in, garbage out"; `merge`'s does not), and note that the flag-bit invariant is not preserved outside that domain.

Either way it deserves a cargo test — `merge(u64::MAX, 0)` — since "unequal inputs ⇒ range word" is currently asserted nowhere in the suite, in Rust or in pytest.

src_rust/src/toc.rs
```diff
@@ -0,0 +1,616 @@
...
+    let hits: Vec<bool> = py.allow_threads(|| {
+        data.par_iter()
+            .map(|&w| {
+                let (s, e, is_rng) = decode(w);
+                // A timestamp is the degenerate closed instant [t, t]; its
+                // half-open envelope for window tests is [t, t + 1) — the
+                // exact instant, one ns wide.
+                let e = if is_rng { e } else { e + 1 };
+                if mode == 0 {
+                    s < q_end_ns && q_start_ns < e
```
espg, line 344: 🤖 *from Claude (review)*

**An empty query window (`q_start == q_end`) reports an overlap for ranges but not for timestamps — the two variants disagree on the same degenerate input.**

`q_start_ns > q_end_ns` is rejected on line 319, so `q_start == q_end` is a legal window. The half-open window `[X, X)` is empty and intersects nothing, and the timestamp arm gets that right (`s < X && X < t+1` needs `t < X ≤ t`, unsatisfiable). The range arm does not — it only needs `X` to sit strictly inside the envelope:

```python
>>> w = span2toc(0, 10**12)
>>> toc2time(w)
(0, 1000727379968)
>>> toc_overlaps(w, 10**9, 10**9)     # empty window, strictly inside the envelope
True
```

Strictly speaking this is inside the documented "may over-report" latitude, but it is not the *quantum-slop* over-report the docstring describes ("by up to one quantum ... near window edges") — it is an unbounded-width answer to a question with no width at all, and it makes `toc_overlaps` variant-dependent on an input where both variants should agree on `False`.

Either short-circuit `q_start_ns == q_end_ns` to all-`False` in `rust_toc_window` (mode 0), or reject the empty window alongside the inverted one on line 319 if callers are never meant to ask. Whichever way, it wants a pinned test — `test_inverted_window_raises` covers `start > end` but nothing covers `start == end`.

mortie/tests/test_toc.py
```diff
@@ -0,0 +1,382 @@
...
+    assert to_gps_ns(GPS_EPOCH_NS) == 0
+    assert to_datetime64(GPS_EPOCH_NS) == np.datetime64("1980-01-06", "ns")
+
+
+def test_epoch_identity_and_pre1972_convention():
+    # Zero offset before 1972 pins the epoch identity exactly.
+    assert from_datetime64("1850-01-01") == 0
+    assert to_datetime64(0) == np.datetime64("1850-01-01", "ns")
+    with pytest.raises(ValueError, match="before the 1850"):
+        from_datetime64("1849-12-31T23:59:59")
```
espg, line 341: 🤖 *from Claude (review)*

**Test gap: the pre-1972 convention flagged for review is the one timescale behavior with no pinned test.**

This test pins the epoch identity and the below-1850 rejection, but not the *consequence* the PR raises under "Questions for review" — the 9 s backward step at 1972-01-01 and the resulting aliasing. That is the piece a reviewer is being asked to sign off on, and it is exactly the piece a future leap-table edit could silently change. It is real and it does behave as documented:

```python
>>> from_datetime64('1971-12-31T23:59:55') == from_datetime64('1972-01-01T00:00:04')
True                                    # the last 9 SI seconds of 1971 alias early 1972
>>> from_datetime64('1971-12-31T23:59:59') < from_datetime64('1972-01-01T00:00:00')
False                                   # from_datetime64 is not monotone across the step
```

Three one-liners would pin the whole documented contract:

1. the aliasing pair above (the documented non-invertibility);
2. non-monotonicity across the boundary, so nobody "fixes" it by accident;
3. `from_datetime64('1971-06-01')` equals the plain naive day-count ns (the zero-offset branch, `_OFFSETS_NS[0]`, which no test currently reaches through a pre-1972 date other than the 1850 epoch itself).

Adjacent, and cheaper still: `to_datetime64` is two-to-one at each leap second — `to_datetime64(inside_leap)` and `to_datetime64(step_instant)` both land in the following UTC second — and `test_leap_second_boundary_pinned` pins the interior render (`00:00:00.5`) but not the collision at the second's start, where `to_datetime64(step_internal - 10**9) == to_datetime64(step_internal) == 2017-01-01T00:00:00`. The `to_datetime64` docstring states the roundtrip direction that *is* exact, so no doc change is needed — just the pin.

mortie/tests/test_toc.py
```diff
@@ -0,0 +1,382 @@
...
+        time2toc(np.array([-1]))
+    with pytest.raises(ValueError, match="ceiling"):
+        time2toc(TOC_MAX_NS)
+    with pytest.raises(ValueError, match="ceiling"):
+        span2toc(0, TOC_MAX_NS)
+    with pytest.raises(ValueError, match="after its end"):
+        span2toc(5, 3)
+    # The last valid instant is fine for both encoders.
+    assert time2toc(TOC_MAX_NS - 1) > 0
+    assert span2toc(0, TOC_MAX_NS - 1) > 0
```
espg, line 312: 🤖 *from Claude (review)*

**Test gap: three documented `ValueError` guards are never executed.** `pytest --cov` on this branch reports `mortie/toc.py` lines **71, 554, 595** uncovered — and each is a `raise` that a docstring advertises:

| line | guard | advertised in |
|---|---|---|
| 71 | `_as_scalar_ns`: `"{name} must lie in [0, 2**64)"` | `toc_overlaps` / `toc_contains` "Raises" |
| 554 | `to_datetime64`: `"t_ns must lie below 2**63"` | `to_datetime64` "Raises ... at or beyond `2**63`" |
| 595 | `from_gps_ns`: `"gps_ns input maps at or beyond the toc span ceiling"` | `from_gps_ns` "Raises ... maps at or beyond `TOC_MAX_NS`" |

All three do fire when reached, e.g. `to_datetime64(np.uint64(2**63))` → `ValueError: t_ns must lie below 2**63`; they are simply unpinned, so a refactor can delete or invert any of them and the suite stays green. Line 554 is the one worth caring about most: it is the only thing standing between a `uint64` ≥ 2^63 and `t.astype(np.int64)` wrapping negative on the next line, which would render a year-2142 instant as a pre-1850 date instead of raising.

Three `pytest.raises` lines next to the existing `test_dtype_and_domain_validation` block would close it:

```python
with pytest.raises(ValueError, match=r"2\*\*63"):
    to_datetime64(np.uint64(1 << 63))
with pytest.raises(ValueError, match="ceiling"):
    from_gps_ns(TOC_MAX_NS - GPS_EPOCH_NS)
with pytest.raises(ValueError, match=r"\[0, 2\*\*64\)"):
    toc_overlaps(time2toc(0), 0, 1 << 64)
```

Also unpinned in the same vein (no `raise`, but no coverage either): the four timescale functions with empty arrays. Each has a `.size` guard implying the author considered the case, `test_empty_arrays` covers the eight word ops, and `from_datetime64` / `to_datetime64` / `from_gps_ns` / `to_gps_ns` are absent from it.

docs/api/toc.md
```diff
@@ -0,0 +1,28 @@
...
+`datetime64` / GPS converters are the only place leap seconds exist. Not an
+IVOA T-MOC. These flat-array elementwise ops are the type's scalar surface
+(the same relationship [mortie.moc](moc.md) has to its ops over one cover);
+the ragged many-cover plurals land in [mortie.batch](batch.md) when the
+interval-set algebra (issue #177) activates. The names stay flat on the
+package (`mortie.time2toc`, ...).
+
+::: mortie.toc
+    options:
+      members:
```
espg, line 16: 🤖 *from Claude (review)*

**Doc accuracy: `TOC_MAX_NS` is referenced by four rendered docstrings but is reachable from neither the rendered page nor the flat package namespace.**

`mortie/toc.py` defines four public constants — `Q_START_NS`, `Q_END_NS`, `TOC_MAX_NS`, `GPS_EPOCH_NS` — each with a `#:` doc comment, so they were clearly meant to be part of the surface. But:

```python
>>> import mortie
>>> hasattr(mortie, 'TOC_MAX_NS'), hasattr(mortie, 'GPS_EPOCH_NS')
(False, False)
```

They are not in `mortie/__init__.py`'s import list or `__all__` (only the twelve functions are), and they are not in this page's `members:` list. Meanwhile `TOC_MAX_NS` appears unqualified in the `Parameters` and `Raises` sections of `time2toc`, `span2toc`, `from_datetime64` and `from_gps_ns`, and `GPS_EPOCH_NS` in `from_gps_ns` / `to_gps_ns` — so a reader of the rendered API page is told an input must lie in `[0, TOC_MAX_NS)` with no way to resolve, click, or import the name, and this page's own line 11–12 note ("The names stay flat on the package") reads as if they were flat too.

`test_docs_api_pages.py` won't catch this: by design it pins only the stale-entry direction (every `members:` name must exist on the module), never the reverse.

Two candidate fixes, either fine: add the four constants to `members:` here **and** to the `from .toc import (...)` block plus `__all__` in `mortie/__init__.py` (matching how the docstrings already read); or, if constants are deliberately module-only in this repo, qualify the references as `mortie.toc.TOC_MAX_NS` in the docstrings so the rendered text is at least resolvable. Worth confirming against the house convention — `mortie.moc` / `mortie.coverage` export names like `RingValidity` flat, which suggests the first option.

mortie/toc.py
espg: 🤖 *from Claude (review fold)*

Fixed in b3918d4: the guard now bounds the *internal* result — `limit = TOC_MAX_NS - _EPOCH_1850_1970_NS - _OFFSETS_NS[-1]`, evaluated before the offset addition. The offset in force anywhere near the ceiling is the table's last entry (+18 s), so the limit is exact: no silent over-ceiling return, no int64 overflow window, no misdirected pre-1850 error.

src_rust/src/toc.rs
espg: 🤖 *from Claude (review fold)*

Fixed in b3918d4: the derivation comment now says 61,361 days (here, in both test_toc.py comments, and in the PR body's fixture table). The pinned ns value and word were derived from `datetime.date` arithmetic and are unchanged — only the prose count was off by one.

mortie/toc.py
espg: 🤖 *from Claude (review fold)*

Fixed in b3918d4: `from_datetime64` now classifies a 0-d ndarray as array (`np.ndim(when) == 0 and not isinstance(when, np.ndarray)`), matching what `np.isscalar` does for the word ops — `np.isscalar` itself is unusable here because it is False for a `np.datetime64` scalar, which is why the convention diverged. Pinned by the new 0-d assertion in `test_datetime64_scalar_forms`.

src_rust/src/toc.rs
espg: 🤖 *from Claude (review fold)*

Partly folded, partly standing down in b3918d4. Folded: `merge`'s doc now states its domain explicitly (encoder-produced words; arbitrary bit patterns are garbage in, garbage out, and can even emerge flag-set), and the cargo law test now asserts the previously-unpinned "unequal valid words merge to a range" property. Standing down on saturation: `2**64 - 1` is not an encodable word (both encoders reject `>= TOC_MAX_NS`), so this is the same GIGO posture `decode` already documents — adding a saturating branch would put a domain check in the one hot path the encoders were designed to keep clean. Left for espg if a validating merge is wanted instead.

src_rust/src/toc.rs
espg: 🤖 *from Claude (review fold)*

Fixed in b3918d4: an empty window (`q_start == q_end`) now matches nothing for both variants and both predicates — the overlap formula assumes both intervals are nonempty, which holds for every word envelope but not for the window, so the empty case is short-circuited. Both predicate docstrings state it; `test_empty_window_matches_nothing` pins it, including the case where the empty window sits strictly inside a range's envelope.

mortie/tests/test_toc.py
espg: 🤖 *from Claude (review fold)*

Fixed in b3918d4: `test_pre1972_zero_offset_and_1972_step` pins all three behaviors — the zero-offset branch (1960-01-01 equals its exact proleptic day count), the 9 s aliasing (`1971-12-31T23:59:55 == 1972-01-01T00:00:04`, and the step non-monotonicity `1972-01-01 == 1971-12-31T23:59:51`), and the convention that aliased instants render on the 1972 side, with the first invertible instant (1972-01-01 itself) round-tripped. The leap-second two-to-one collision at a step is the documented rendering already pinned by `test_leap_second_boundary_pinned`'s inside-the-leap-second case.

mortie/tests/test_toc.py
espg: 🤖 *from Claude (review fold)*

Fixed in b3918d4: `test_timescale_domain_errors` executes all three guards — `to_datetime64(1 << 63)` (the int64-wrap gate you called out as the one that matters), `from_gps_ns(TOC_MAX_NS)`, and a negative window bound through `_as_scalar_ns` — and `test_timescale_empty_arrays` covers all four timescale functions on empty input.

docs/api/toc.md
espg: 🤖 *from Claude (review fold)*

Fixed in b3918d4: the four constants (`Q_START_NS`, `Q_END_NS`, `TOC_MAX_NS`, `GPS_EPOCH_NS`) are now in the page's `members:` list, and their comments were converted to real attribute docstrings so mkdocstrings renders them. Deliberately *not* added to the flat `mortie` namespace: the house precedent is `mortie.orders.MAX_ORDER`, which is likewise reachable only on its submodule — noted here rather than silently diverging, for espg to overrule if flat exports are wanted.

Issue #147, opened 2026-07-26
Winding-free dissolve classifier: hemisphere+ covers dissolve instead of raising


espg:
Promotes #97 item 5 to a concrete plan, on espg's direction from the PR #111 thread. Refs #108, #111, #138, #146.

## What this enables

`dissolve` (cover → outline) currently cannot handle covers at or past a hemisphere: exterior/hole classification keys off the sign of a mod-4π anchor-fan signed area, which wraps as the cover nears 2π. PR #111 makes that failure **loud** (curated `ValueError`s on over-hemisphere covers and hemisphere-enclosing rings) where it used to be silent inversion or a stitcher panic — the right interim, but a rejection, not a capability. This issue replaces the classifier so hemisphere+ dissolves *work*: ocean masks, "everything except Antarctica" covers, polar-scale mosaics.

Note the asymmetry with coverage: `morton_coverage` has handled hemisphere+ input correctly since PR #138. Dissolve is the only remaining hemisphere-bounded surface.

## Why now — the instruments exist

The classifier decomposes into exactly the questions #138/#146 built exact, oracle-backed tools for:

| classifier question | instrument | provenance |
|---|---|---|
| which way is this ring wound? | `winding_sign_in_cap` / `ring_turning_sign` — Gauss–Bonnet sign with a derived rounding bound, any ring size | #138 phases 1/4, #144 decision (A) |
| is that sign trustworthy? | `ring_is_simple` / `ring_set_validity` — the checkable simplicity precondition | #146 (#145) |
| which rings nest inside which? | `point_in_ring_robust` / `parity_filled_robust` — exact at any size | #138 phase 1 |

None of the wrapping fan-formula machinery needs to survive.

### Phase 1 — stitcher-convention audit (no behavior change)

The whole classifier gets cheap if one invariant holds: rings emitted by edge cancellation + antimeridian stitching inherit **covered-region-on-the-left** orientation from cell-edge traversal. Today that invariant is neither documented nor tested, and PR #111 measured the stitcher's `HashMap`-seeded chaining producing non-deterministic ring *sign* at hemisphere scale — so it must be established, not assumed.

- Audit `src_rust/src/dissolve.rs`'s edge-emission and stitching order; determine whether cover-on-left holds always, holds up to per-ring reversal, or is genuinely indeterminate. Pin the answer with a determinism test (same cover, shuffled input order → identical oriented rings).
- Same audit for the Python oracle mirror (`_dissolved_rings_py`).
[...]

PR #179, opened 2026-08-09
Deterministic, order-independent ring chaining in dissolve (issue #155)

### LINKED ISSUE #155: dissolve: sub_hemisphere_cover_still_dissolves flakes ~15% on a deterministic input (anchor-order nondeterminism trips the hemisphere guard)

espg:
Closes #155. Refs #147 (classifier successor — its territory untouched here), issue #28 (determinism precedent), PR #111 (guard behavior preserved).

Makes `dissolve`'s ring chaining deterministic and order-independent, killing the ~10–15% flake in `dissolve::tests::sub_hemisphere_cover_still_dissolves`. Scope is `src_rust/src/dissolve.rs` only, per the [implementation plan](https://github.com/espg/mortie/issues/155#issuecomment-5230222605) and the [2026-08-09 rulings](https://github.com/espg/mortie/issues/155#issuecomment-5230222606).

## Phases

- [x] **Phase 1 — pin the repro + instrument the failing geometry.** `sub_hemisphere_cover_dissolves_deterministically`: 50 in-process `dissolve` calls on the issue's cover (base cells 0–3 at order 1). **Pre-fix: 48/400 single calls fail (12.0%)**, and the 50-iteration loop fails with p ≈ 1 − (7/8)⁵⁰ ≈ 0.999.
- [x] **Phase 2 — order-independent `chain_rings`**: canonical sorted edge stream, upfront rotational pairing (pure successor function computed before any chain is walked — per-vertex rotational sweep with cyclic bracket matching, LIFO so nested wedges resolve to separate simple rings), ring extraction as cycle-following in the functional graph, seeded from each ring's smallest sorted edge — which makes every ring start at its minimal vertex id, so the classifier's fan anchor is deterministic. The old alive-filter and its history dependence are gone; azimuth ties break by canonical index. Classifier/guards untouched.
- [x] **Phase 3 — tests + doc note**: `dissolve_output_is_byte_identical_across_calls` (5 fixture shapes: north cover step 1/4, south polar cap exercising the pole/antimeridian stitch, equatorial cell step 1/3); `chain_rings_is_order_independent` (200 LCG-seeded Fisher–Yates permutations plus reverse-sorted order → identical ring list — pins the invariant, not one lucky order — over three edge sets: the real cover's survivors, the corner-touch fixture, and a net > 1 duplicate-edge set); `corner_touch_pairs_into_simple_rings` (two diamonds sharing a vertex → exactly two simple rings each visiting the touch point once, per the ruled pairing); doc comment on `chain_rings` stating the pure-successor/LIFO contract; `survivor_edges` split out of `boundary_rings_xyz` so tests drive `chain_rings` with the real edge pipeline.
- [x] **Review fold** (82065c3, addressing the [adversarial review](https://github.com/espg/mortie/pull/179#discussion_r3743052040)): explicit canonical-rotation block removed as dead code (the sorted seeding already yields it), doc comment states LIFO bracket matching as the contract, permutation test extended to the branching and duplicate-edge shapes. The review's blocking finding is a decision for espg — see "Questions for review".

## Phase-1 finding: the branch-geometry question has a third answer

The [ruling](https://github.com/espg/mortie/issues/155#issuecomment-5230222606) left open whether the cover's branching vertex is a true corner-touch or a snap-merge artifact. Instrumentation says: **neither — there is no branching vertex at all.**

- The 0–3 cover's boundary graph is a **single simple 16-cycle**: 25 snapped vertices, 16 surviving directed edges, every ring vertex has in-degree = out-degree = 1, **zero duplicate directed edges** (no `net > 1`), and **zero near-coincident vertex pairs** (< 1e-8 chord) — the integer snap merged everything cleanly. 2000 fresh-`HashMap` chainings produced exactly **one** ring decomposition. The ring is a 4-pointed star around the north pole: 4 spike tips on the equator (lon ±45°/±135°, lat 0°), 4 peaks at lat 41.81°, 8 mid vertices at lat 19.47°.
- The nondeterminism is therefore purely the **starting rotation** of that one ring — the seed edge comes from `HashMap` iteration order (`counts` at `boundary_rings_xyz`), and `classify_and_split` fans `spherical_signed_area` from `ring[0]`. Sweeping all 16 rotations: 14 anchors give the honest fan sum **4.089857 sr** (true cover area 4.188790 sr), but the 2 equatorial spike-tip anchors at (±45°, 0°) wrap the fan to **0.790082 sr**, tripping the PR #111 Σ-vs-exact-area guard. Predicted failure rate 2/16 = **12.5%**; measured **48/400 = 12.0%**; issue records ~10–15%. Exact mechanism match.
- This *refines* the plan's hypothesis (different ring decompositions at branching vertices): on this cover the decomposition is unique and the original issue body's anchor-fan hypothesis was the operative one. The fix still presses exactly where the plan said — chaining canonicalization — because the ring's starting rotation is chaining output: sorting the edge stream and seeding each ring from its smallest sorted edge makes every ring start at its minimal vertex id, deterministically. Vertex ids are insertion-ordered (deterministic), and the minimal ring id (vertex 1, at lon 22.5° lat 19.47°) is a non-wrapping anchor for this cover, so the guard passes deterministically. Neither snap handling nor the classifier needs to change; the fan's anchor-fragility itself remains #147's territory.

## Approach (phase 2)

- Collect `survivors`, then `sort_unstable` — the canonical edge stream (the `HashMap` stays for O(1) counting; one O(E log E) sort of boundary edges is noise).
- Replace the alive-filtered, history-dependent successor selection with an upfront **rotational pairing**: at each vertex, sweep in-edges and out-edges by azimuth cyclically and pair them by LIFO bracket matching (interior-consistent, per the ruling — at a branching vertex the boundary pinches into simple rings touching at a point, never a figure-eight; where arrivals and departures alternate this coincides with the old smallest-turn rule). `successor(edge)` becomes a pure function computed before any chain is walked; ring extraction is cycle-following in a functional graph. Azimuth ties (duplicate directed edges) break by canonical index.
- Each emitted ring starts at its minimal vertex id, because rings are seeded from their smallest sorted edge.
- The classifier, the hemisphere/wrap guards, and everything from `classify_and_split` down are untouched.

## How tested

- Phase 1: `cargo test --lib dissolve` — the new loop test fails pre-fix as designed (48/400 single-call failures measured); all pre-existing dissolve tests pass.
- Post-fix: full `cargo test --lib` green (335 passed, 1 ignored), dissolve suite run 10+ consecutive times green (500+ loop iterations — pre-fix expectation ~62 failures); the rejection fixtures `hemisphere_cover_fails_loud` and `equatorial_band_fails_loud_not_cryptic` unchanged and passing (guard behavior preserved per PR #111).
- `cargo fmt --check` clean; `cargo clippy --lib --tests` — zero warnings in `dissolve.rs` (the 7 remaining sites are pre-existing in other modules, unchanged).
- Python gate in a clean venv (rerun after the review fold): `maturin develop --release` OK, `flake8 mortie --select=E9,F63,F7,F82` clean, `pytest -q` **1364 passed, 16 skipped**, `numpydoc lint mortie/*.py` clean. No Python-surface changes (the diff is `src_rust/src/dissolve.rs` only); the Rust↔Python differential dissolve tests compare rotation-invariantly (symmetric difference + structural fingerprint), so the canonical ring start is transparent to them.

## Questions for review

- **The deterministic anchor converts the per-call lottery into a per-cover verdict** — quantified by the [review's 3145-cover sweep](https://github.com/espg/mortie/pull/179#discussion_r3743052040): nondeterminism 2972/3145 covers → **0**; ring topology byte-identical old-vs-new throughout; zero regressions in the realistic order-2–5 families; but in the coarse order-1 base-cell family, **249 covers that sometimes passed the PR #111 guards now deterministically fail** (191 passed ≥50% of calls pre-fix; 27 ≥90%; 1 passed 40/40), while 752 flaky covers become always-pass and the aggregate pass rate is essentially unchanged (91.5% → 91.3%). Failure incidence scales with cover area (0% at 1–2 base cells → 24% at 5). How to weigh this — land as-is, make the rotation fan-aware (scope change), or hold behind #147 — is posed as a [decision on the thread](https://github.com/espg/mortie/pull/179#issuecomment-5230428447), recommendation first.
- Two candidate follow-up issues named in that decision comment (the review's two named covers as #147 acceptance fixtures; the pre-existing stitcher `assert!` surfacing as `PanicException` instead of a curated `ValueError`) — not opened unilaterally.

### DIFF
--- a/src_rust/src/dissolve.rs
+++ b/src_rust/src/dissolve.rs
@@ -177,59 +177,118 @@ fn tangent_azimuth(p: &Vec3, q: &Vec3) -> f64 {
     dot(&d, &east).atan2(dot(&d, &north))
 }
 
+/// Chain surviving directed edges into rings.
+///
+/// Deterministic and order-independent by construction (issue #155):
+/// permuting `survivors` cannot change the output.  The edge stream is
+/// canonicalized by sorting, `successor(edge)` is a **pure function of the
+/// edge set** computed up front (a rotational sweep at every vertex pairs
+/// each arriving edge with the first departing edge encountered sweeping in
+/// +azimuth from its reversed-arrival direction — the old smallest-turn rule
+/// freed of its history-dependent aliveness filter), and each emitted ring
+/// is rotated to a canonical start (minimal vertex id, ties by smallest id
+/// sequence).  The pairing is interior-consistent: at a non-manifold
+/// corner-touch vertex the boundary pinches into simple rings touching at a
+/// point, never a self-intersecting figure-eight.  Do not reintroduce
+/// history-dependent successor choices (e.g. filtering candidates by which
+/// edges an earlier ring consumed) — ring output must stay a pure function
+/// of the edge *set*, or dissolve's classification goes nondeterministic
+/// again (fan-anchor wrap, issue #155).
 fn chain_rings(survivors: &[(u32, u32)], id_xyz: &[Vec3]) -> Vec<Vec<Vec3>> {
-    let two_pi = std::f64::consts::TAU;
-    let az: Vec<f64> = survivors
-        .iter()
-        .map(|&(a, b)| tangent_azimuth(&id_xyz[a as usize], &id_xyz[b as usize]))
-        .collect();
-    let mut alive: Vec<bool> = vec![true; survivors.len()];
-    let mut by_start: HashMap<u32, Vec<usize>> = HashMap::new();
-    for (i, &(a, _)) in survivors.iter().enumerate() {
-        by_start.entry(a).or_default().push(i);
+    // Canonical edge stream: sorted, duplicate directed edges adjacent.
+    let mut edges: Vec<(u32, u32)> = survivors.to_vec();
+    edges.sort_unstable();
+    let n = edges.len();
+
+    // Rotational sweep entries per vertex: (angle, kind, edge) with kind
+    // 0 = arriving edge at its reversed-arrival azimuth, kind 1 = departing
+    // edge at its forward azimuth.  At equal angles an arrival sorts first,
+    // so a departing edge lying exactly along a reversed arrival pairs with
+    // it (zero turn — the old rule's minimum); remaining ties break by
+    // canonical edge index.
+    let mut at_vertex: HashMap<u32, Vec<(f64, u8, usize)>> = HashMap::new();
+    for (i, &(a, b)) in edges.iter().enumerate() {
+        let pa = &id_xyz[a as usize];
+        let pb = &id_xyz[b as usize];
+        at_vertex
+            .entry(a)
+            .or_default()
+            .push((tangent_azimuth(pa, pb), 1, i));
+        at_vertex
+            .entry(b)
+            .or_default()
+            .push((tangent_azimuth(pb, pa), 0, i));
     }
 
+    // successor[i] = the edge following edge i in its ring.  Cyclic bracket
+    // matching per vertex: sweep entries in ascending angle (twice, for the
+    // wrap-around); a departing edge closes the most recently opened arrival
+    // (LIFO), so nested wedges resolve to separate simple rings instead of
+    // crossing.  Balanced in/out degrees (guaranteed by edge cancellation)
+    // leave nothing unpaired; an unbalanced vertex leaves `None`, emitted
+    // below as an open chain exactly like the old dead-end break.
+    let mut successor: Vec<Option<usize>> = vec![None; n];
+    for entries in at_vertex.values_mut() {
+        entries.sort_by(|x, y| {
+            x.0.partial_cmp(&y.0)
+                .unwrap()
+                .then(x.1.cmp(&y.1))
+                .then(x.2.cmp(&y.2))
+        });
+        let mut pending: Vec<usize> = Vec::new();
+        let mut matched: Vec<bool> = vec![false; entries.len()];
+        for pass in 0..2 {
+            for (e, &(_, kind, i)) in entries.iter().enumerate() {
+                if kind == 0 {
+                    if pass == 0 {
+                        pending.push(i);
+                    }
+                } else if !matched[e] {
+                    if let Some(arrived) = pending.pop() {
+                        successor[arrived] = Some(i);
+                        matched[e] = true;
+                    }
+                }
+            }
+        }
+    }
+
+    // Ring extraction: cycle-following in the functional graph, seeds in
+    // canonical edge order.
     let mut rings = Vec::new();
-    for seed in 0..survivors.len() {
-        if !alive[seed] {
+    let mut visited = vec![false; n];
+    for seed in 0..n {
+        if visited[seed] {
             continue;
         }
-        let seed_start = survivors[seed].0;
-        let mut cur = seed;
         let mut chain: Vec<u32> = Vec::new();
+        let mut cur = seed;
         loop {
-            if !alive[cur] {
-                break;
+            visited[cur] = true;
+            chain.push(edges[cur].0);
+            match successor[cur] {
+                Some(nxt) if nxt == seed => break, // ring closed
+                Some(nxt) if !visited[nxt] => cur = nxt,
+                _ => break, // open chain (unbalanced vertex)
             }
-            alive[cur] = false;
-            chain.push(survivors[cur].0);
-            let v = survivors[cur].1;
-            if v == seed_start {
-                break; // ring closed
-            }
-            let cands: Vec<usize> = by_start
-                .get(&v)
-                .map(|ix| ix.iter().copied().filter(|&i| alive[i]).collect())
-                .unwrap_or_default();
-            if cands.is_empty() {
-                break;
-            }
-            cur = if cands.len() == 1 {
-                cands[0]
-            } else {
-                // smallest turn anticlockwise from the reversed arrival.
-                let back = tangent_azimuth(&id_xyz[v as usize], &id_xyz[survivors[cur].0 as usize]);
-                *cands
-                    .iter()
-                    .min_by(|&&i, &&j| {
-                        let ti = (az[i] - back).rem_euclid(two_pi);
-                        let tj = (az[j] - back).rem_euclid(two_pi);
-                        ti.partial_cmp(&tj).unwrap()
-                    })
-                    .unwrap()
-            };
         }
-        rings.push(chain.iter().map(|&i| id_xyz[i as usize]).collect());
+        // Canonical rotation: start at the minimal vertex id; if a ring
+        // passes through it more than once, the smallest id sequence.
+        let m = chain.len();
+        let min_id = *chain.iter().min().unwrap();
+        let mut best: Option<Vec<u32>> = None;
+        for s in (0..m).filter(|&s| chain[s] == min_id) {
+            let rot: Vec<u32> = (0..m).map(|k| chain[(s + k) % m]).collect();
+            if best.as_ref().is_none_or(|b| rot < *b) {
+                best = Some(rot);
+            }
+        }
+        rings.push(
+            best.unwrap()
+                .iter()
+                .map(|&i| id_xyz[i as usize])
+                .collect::<Vec<Vec3>>(),
+        );
     }
     rings
 }
--- a/src_rust/src/dissolve.rs
+++ b/src_rust/src/dissolve.rs
@@ -7,9 +7,11 @@
 //!    boundary point is integer-snapped to a vertex id, so a corner/sub-edge
 //!    point shared by two neighbours collapses to one id and their shared edge
 //!    cancels exactly (no floating tolerance search).
-//! 2. The surviving directed edges chain into rings; at a non-manifold
-//!    corner-touch vertex the next edge is the smallest anticlockwise turn from
-//!    the reversed arrival (right-hand rule), keeping rings simple.
+//! 2. The surviving directed edges chain into rings via an upfront rotational
+//!    pairing — a pure, order-independent successor function (issue #155); at
+//!    a non-manifold corner-touch vertex each arriving edge pairs with the
+//!    smallest-turn departing edge, pinching the boundary into simple rings
+//!    touching at a point (never a self-intersecting figure-eight).
 //! 3. Rings are classified exterior/hole by spherical signed area (global
 //!    winding normalised so the covered area is positive), then crossing rings
 //!    are cut at +/-180 and reconnected by the GeoJSON convention — explicit
@@ -83,8 +85,16 @@ pub fn dissolve(morton: &[u64], step: u32) -> Result<ClassifiedRings, String> {
 // ── edge-cancellation: cover → boundary rings (unit vectors) ───────────────
 
 fn boundary_rings_xyz(morton: &[u64], step: u32) -> Vec<Vec<Vec3>> {
+    let (survivors, id_xyz) = survivor_edges(morton, step);
+    chain_rings(&survivors, &id_xyz)
+}
+
+/// Surviving directed boundary edges of a cover, plus the snapped vertex
+/// coordinates they index (split out of [`boundary_rings_xyz`] so tests can
+/// drive [`chain_rings`] with permuted edge slices, issue #155).
+fn survivor_edges(morton: &[u64], step: u32) -> (Vec<(u32, u32)>, Vec<Vec3>) {
     if morton.is_empty() {
-        return Vec::new();
+        return (Vec::new(), Vec::new());
     }
     // Decode depth per word; densify a mixed-order MOC to its finest order so
     // every cell carries unit-length edges that cancel against their neighbours.
@@ -153,7 +163,7 @@ fn boundary_rings_xyz(morton: &[u64], step: u32) -> Vec<Vec<Vec3>> {
             survivors.push((a, b));
         }
     }
-    chain_rings(&survivors, &id_xyz)
+    (survivors, id_xyz)
 }
 
 // ── ring chaining (angular / right-hand rule at non-manifold vertices) ─────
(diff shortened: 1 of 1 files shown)


src_rust/src/dissolve.rs
```diff
@@ -177,59 +187,118 @@ fn tangent_azimuth(p: &Vec3, q: &Vec3) -> f64 {
...
+                _ => break, // open chain (unbalanced vertex)
             }
-            let cands: Vec<usize> = by_start
-                .get(&v)
-                .map(|ix| ix.iter().copied().filter(|&i| alive[i]).collect())
-                .unwrap_or_default();
-            if cands.is_empty() {
-                break;
+        }
+        // Canonical rotation: start at the minimal vertex id; if a ring
```
espg: 🤖 *from Claude (review)*

**blocking — the deterministic anchor turns the ~12% per-call flake into a per-cover verdict, and ~250 covers land on "always fails"**

Measured rather than reasoned. Corpus: 3145 covers — every 1-to-5-base-cell mask at order 1 (1585, of which 60 also at `step=3`), 1200 contiguous nested runs at orders 2–5, 300 scattered order-3 sets. Each dissolved 12× on this branch and 40× at the pre-fix base (`d8c6e8b`), comparing `ClassifiedRings` byte-for-byte. Harness was an out-of-tree crate taking this worktree as a path dependency; nothing in the worktree was touched.

The headline claim holds, decisively:

- pre-fix, 2972/3145 covers produced ≥2 distinct outputs across repeated calls; on this branch, **0/3145**.
- the realistic families (orders 2–5 runs, order-3 scatter) show zero regressions — every regression below is in the coarse order-1 base-cell family.

The cost, all in that coarse family:

| pre-fix → post-fix | covers |
|---|---|
| passed 40/40 → never passes | 1 |
| passed ≥90% of calls → never passes | 27 |
| passed ≥50% of calls → never passes | 191 |
| passed at least once → never passes | **249** |
| flaky → always passes | 752 |
| never passed → always passes | 0 |

Aggregate pass rate is essentially unchanged (91.5% → 91.3%): min-vertex-id is an *average* anchor, not a bad one. That is exactly the problem — the PR converts a lottery into a verdict, and for 249 covers the verdict is "no". Regression rate scales with area: 0% at 1–2 base cells, 2% at 3 (π sr), 9% at 4 (4.19 sr), 24% at 5 (5.24 sr).

Two concrete cases:

- base cells 4, 6, 7, 10 at order 1 (16 cells, 4.19 sr): **40/40 pass** pre-fix (15 distinct outputs, all fine) → **12/12** `boundary ring enclosing more than a hemisphere` here.
- base cells 1, 2, 7 at order 1 (12 cells, **π sr — half the guard's budget**): 35/40 pass pre-fix → **12/12 panic** here, on the `assert!` at line 444 (`unbalanced antimeridian segments but no pole enclosed`).

Worth separating two things:

- I checked the ring *decomposition* directly on the regressing covers (rotation-invariant fingerprint of `boundary_rings_xyz`, old sampled 40×): it is **identical** old vs new on every one. The new pairing is not changing topology — only which vertex anchors the `spherical_signed_area` fan. So this is precisely the fragility flagged under "Questions for review"; the gap is that the body reads as "a live fragility for other covers" rather than "~250 covers in a 3145-cover sweep now fail on every call".
- The `assert!` path is **pre-existing and this PR improves it in aggregate** (348 covers hit it at least once pre-fix vs 79 post-fix). Please don't fix it here — but an internal `assert!` surfacing in Python as `PanicException` instead of the curated `ValueError` deserves its own issue.

This needs an @espg call, since the natural fix — an anchor-free ring area — is #147's territory and explicitly out of scope per the [2026-08-09 ruling](https://github.com/espg/mortie/issues/155#issuecomment-5230222606). Options:

1. Land as-is, put the transition table in the PR body, and open a follow-up carrying the two covers above as fixtures for #147. Defensible: determinism was the goal, aggregate behavior is unchanged, realistic covers are untouched.
2. Stay in scope and make the rotation fan-aware: `chain_rings` already owns the anchor choice, so it can pick the lexicographically smallest rotation whose `spherical_signed_area` agrees with the modal value over all rotations — still a pure function of the edge set, and O(m²) only on rings that disagree.
3. Hold this behind #147.

Either way the numbers belong in the PR body: the plan's acceptance criterion was "guard-consistency, not a presumed ring count", and 249 deterministic rejections is the honest reading of guard-consistency for this change.

src_rust/src/dissolve.rs
```diff
@@ -177,59 +187,118 @@ fn tangent_azimuth(p: &Vec3, q: &Vec3) -> f64 {
...
+        // passes through it more than once, the smallest id sequence.
+        let m = chain.len();
+        let min_id = *chain.iter().min().unwrap();
+        let mut best: Option<Vec<u32>> = None;
+        for s in (0..m).filter(|&s| chain[s] == min_id) {
```
espg: 🤖 *from Claude (review)*

**non-blocking — this block is dead code in practice; the `sort_unstable` above it is what actually fixes the anchor**

Seeds are taken in ascending canonical edge index and each cycle is walked to completion, so the seed of a ring is always that ring's *smallest* edge index. `edges` is sorted lexicographically, so the smallest edge's `.0` is the minimum start vertex over the ring's edges — and in a closed ring the start vertices are exactly the ring's vertices. Hence `chain[0] == min_id` before this block runs. It can only bite in one case: a ring passing through its minimum vertex **more than once** (a pinch at the min-id vertex), where a later occurrence yields a lexicographically smaller sequence.

Measured, not argued: over the same 3145-cover corpus, `boundary_rings_xyz` from this branch vs. an otherwise identical build with lines 285–295 deleted — **0 covers differ**. The `rot < *b` tie-break is therefore also completely unexercised by the suite.

Not a bug, but two things follow:

- the PR body's phase-1 narrative credits the fix to this block ("it just additionally needs a **canonical ring rotation** ..., not only a canonical edge order"). The measurement says the canonical edge order alone does it, via the seed rule. Worth correcting so a future edit doesn't reorder seeds on the belief that the rotation is holding the invariant up.
- if you keep it (reasonable — it makes the invariant independent of the seed rule), say so in the comment ("no-op unless the ring revisits its minimum vertex") and add a `chain_rings` test for a ring that passes twice through its min vertex, so the branch this block exists for is actually run. It is also the only O(m²) path in the function.

src_rust/src/dissolve.rs
```diff
@@ -177,59 +187,118 @@ fn tangent_azimuth(p: &Vec3, q: &Vec3) -> f64 {
...
+        let mut pending: Vec<usize> = Vec::new();
+        let mut matched: Vec<bool> = vec![false; entries.len()];
+        for pass in 0..2 {
+            for (e, &(_, kind, i)) in entries.iter().enumerate() {
+                if kind == 0 {
+                    if pass == 0 {
+                        pending.push(i);
+                    }
+                } else if !matched[e] {
+                    if let Some(arrived) = pending.pop() {
```
espg, line 264: 🤖 *from Claude (review)*

**non-blocking — LIFO bracket matching is not the rule the doc comment specifies**

The doc block above (line 195–197) and the module header (lines 12–13) specify the successor as "the first departing edge encountered sweeping in +azimuth from its reversed-arrival direction" / "the smallest-turn departing edge". What this loop computes is the *most recently opened* arrival, which coincides with that rule only when arrivals and departures strictly alternate around the vertex.

They do alternate for a consistently-wound boundary — with the interior on one flank of every departure and the other flank of every arrival, the rays around a vertex must alternate D, A, D, A — which is why the corner-touch fixture reproduces the old smallest-turn rule exactly. (Checked by hand at T: arrival-west −135°, departure-west −45°, arrival-east +45°, departure-east +135°; both rules pair W→W and E→E, giving the two rings the test asserts.) But alternation is exactly what a snap-degenerate vertex breaks, and there the rules diverge: entries `A1(0°) A2(10°) D1(20°) D2(30°)` give `A1→D1, A2→D1` under the stated rule — not a bijection, so `successor` would not be a permutation — and `A2→D1, A1→D2` (nested) under LIFO.

LIFO is the better choice: it is total and injective where the stated rule is neither. The ask is only to describe what the code does, since this doc block is explicitly written as the contract for future edits. Something like: "each arrival pairs with a departure by cyclic LIFO bracket matching in +azimuth, which coincides with the smallest-turn rule whenever arrivals and departures alternate around the vertex — guaranteed for a consistently-wound boundary, and a total, injective fallback when snap degeneracy breaks alternation."

src_rust/src/dissolve.rs
```diff
@@ -583,6 +652,127 @@ mod tests {
...
+
+    #[test]
+    fn chain_rings_is_order_independent() {
+        // issue #155 phase 3: chain_rings output is a pure function of the
+        // edge *set* -- every permutation of the survivor slice yields the
+        // identical ring list (this is the invariant, not one lucky order).
+        let cover: Vec<u64> = (0..16u64)
+            .map(|n| crate::morton::nested2mort(n, 1))
+            .collect();
+        let (mut edges, id_xyz) = survivor_edges(&cover, 1);
```
espg: 🤖 *from Claude (review)*

**non-blocking — this test can't exercise the thing phase 2 rewrote**

It does drive the real pipeline and it does have teeth against the actual bug (pre-fix the ring's rotation follows `survivors[0]`, so any permutation changes the output). But the cover it permutes is the one phase 1 proved is a *single simple 16-cycle — no branching vertex, no duplicate directed edges, no near-coincident vertices*. On that graph the successor function is forced: every chaining rule agrees. So the 200 permutations pin the canonical seed order and nothing about the rotational sweep, the LIFO matching, or the `(angle, kind, index)` tie-break, which is where all the new complexity lives. A `chain_rings` that got the pairing rule wrong at a branching vertex would still pass this test.

Cheap fix, a few lines each, both pure `chain_rings` inputs:

- run the same permutation loop over `corner_touch_pairs_into_simple_rings`'s 8-edge fixture — that one has a genuine degree-4 branching vertex, so it pins the pairing, not just the seed;
- add a fixture with a duplicated directed edge (the `net > 1` case the plan calls out as the historical tie-break leak) so the `.then(x.2.cmp(&y.2))` index tie-break is covered by something.

Same gap one test down in `dissolve_output_is_byte_identical_across_calls`: the north cap, south cap and single equatorial base cell are all single-shell, hole-free and branch-free. A holed cover and a corner-touch cover would make it a determinism test over the paths that actually changed.

src_rust/src/dissolve.rs
espg: 🤖 *from Claude*

Acknowledged — and the measurement is exactly what was needed: the decomposition-identity check (topology byte-identical old vs new on every regressing cover) confirms the phase-1 analysis that this is purely the fan anchor, while the 249-cover transition table is a much sharper statement than the body's "live fragility" phrasing was.

Actions taken:

- The transition table and the two concrete covers are now in the PR body under "Questions for review", per your closing point.
- The decision between your options — (1) land as-is + follow-up fixtures for #147, (2) fan-aware canonical rotation, (3) hold behind #147 — is posed for espg on the PR thread with a lean; it is a design fork touching #147's territory, so it is not mine to make.
- Not fixed here, per your own note: the line-444 `assert!` panic path (pre-existing, and improved in aggregate by this PR, 348 → 79 covers). Flagged in the thread question as deserving its own issue rather than opened unilaterally.

`waiting` goes on with the thread question.

src_rust/src/dissolve.rs
espg: 🤖 *from Claude*

Folded in 82065c3: the rotation block is deleted. You're right about the mechanism — the seed is the ring's smallest sorted edge, whose start vertex is the ring's minimal id, and since both candidate rotations at a repeated min id would share the same first edge (and hence the same successor path), the seeded rotation *is* the lexicographically smallest one. The replacement comment states that invariant, the `chain_rings` doc comment now credits the sorted seeding, and the PR body's phase-2/phase-1 text was corrected to match (no more "canonical rotation" step).

src_rust/src/dissolve.rs
espg: 🤖 *from Claude*

Folded in 82065c3: the doc comment now states LIFO bracket matching as the contract, with the smallest-turn rule demoted to what LIFO coincides with when arrivals and departures alternate (the generic consistently-wound case), and names LIFO as governing when snap degeneracy breaks alternation — with the non-crossing rationale you gave.

src_rust/src/dissolve.rs
espg: 🤖 *from Claude*

Folded in 82065c3: `chain_rings_is_order_independent` now drives three edge sets through `assert_permutation_invariant` (200 LCG shuffles + reverse-sorted each): the real cover's survivors (pins seed/rotation order, as before), the corner-touch fixture (a branching vertex — pins the rotational pairing under permutation), and the corner-touch set with the east diamond duplicated (net = 2 — pins the canonical-index tie-break at identical azimuths, plus asserts the duplicate chains to two identical rings).

On the byte-identical fixture set: the corner-touch and duplicate-edge shapes cannot be pushed through `dissolve` itself — they are synthetic edge sets, and a real cover producing a net > 1 edge requires overlapping cells, which violates dissolve's documented disjoint-cover precondition (`cover_area` doc). Chain-level coverage is where those shapes are reachable, so that is where they are pinned; if you see a legitimate cover construction that reaches a branching vertex at order ≥ 1, that would make a good dissolve-level fixture for a follow-up.

espg:
🤖 *from Claude*

**Decision needed (espg): how to handle the review's measured finding on the deterministic anchor.** Recommendation: **option 1 — land as-is**, with the numbers recorded in the PR body (done) and the two named covers carried as acceptance fixtures when #147 lands.

Background, from the [review's 3145-cover sweep](https://github.com/espg/mortie/pull/179#discussion_r3743052040): determinism is fully achieved (2972/3145 covers nondeterministic pre-fix → 0 post-fix; ring topology byte-identical throughout), but fixing the fan anchor converts the old per-call lottery into a per-cover verdict — **249 coarse order-1 base-cell covers that sometimes passed now deterministically fail** the PR #111 guards (aggregate pass rate essentially unchanged, 91.5% → 91.3%; zero regressions in the realistic order-2–5 families). The failures scale with cover area (0% at 1–2 base cells → 24% at 5), i.e. exactly the near-hemisphere-scale territory #147 exists to serve.

Options, answerable in one pass:

1. **Land as-is** (recommended). Determinism and guard-consistency on the pinned fixture were the ruled goal; the newly-failing covers fail loudly with the curated guard message, not wrongly; and the mechanism that decides pass/fail — the anchor-fan classifier — is precisely what #147 replaces, at which point the verdicts disappear wholesale. Making chaining fan-aware now would bake classifier knowledge into `chain_rings` that #147 deletes a release later.
2. **Fan-aware canonical rotation** (the review's sketch): pick the smallest rotation whose `spherical_signed_area` matches the modal value over all rotations — still a pure function of the edge set, ~30 lines, and rescues most of the 249; but it reaches across the chaining/classifier boundary the 2026-08-09 ruling drew, so it needs your sign-off as a scope change.
3. **Hold this PR behind #147** — accepts the CI flake tax indefinitely; listed for completeness.

Two follow-ups that need your call either way (not opened unilaterally):

- an issue carrying the review's two named covers (base cells 4/6/7/10 order 1; base cells 1/2/7 order 1) as #147 acceptance fixtures;
- an issue for the pre-existing line-444 `assert!` panic surfacing in Python as `PanicException` instead of the curated `ValueError` (this PR reduces its incidence 348 → 79 covers in the sweep, but the path remains).

Applying `waiting`; all phases and the review fold are otherwise complete.

PR #146, opened 2026-07-26
Ring validity: bucketed simplicity check, oracle fixtures, and the cap-axis fast path

espg:
Closes #145 · Closes #144 · Refs #107

The follow-up PR agreed in the #138 replan: the ring-validity machinery that remained #145's own scope after its identity-consistency half was folded into #138 phase 3, bundled with #144's performance half, plus two housekeeping items espg approved off the #138 wrap-up (the `sphere.rs` witness-module split from PR #138 question 2, and the `docs/coverage_methods.md` decision-(A) update).

## What this delivers

- **`ring_is_simple` — the S2-style bucketed self-intersection check** ([#145 plan](https://github.com/espg/mortie/issues/145#issuecomment-5082096277), phase 2): cover each edge with HEALPix cells at an order derived from the vertex count, test non-adjacent co-bucketed pairs with the exact `arcs_cross_sos` predicate. The algorithm is the one the C++ reference itself uses (`s2shapeutil::FindSelfIntersection` over the shape index — no sweep line; see the source-read on the #145 thread).
- **The validity API surface** (#145 plan, phase 3): a queryable Rust `ring_validity` combining simplicity with the identity-consistency check that landed in #138, one public Python function, `debug_assert`-level consumers for the assumptions `sphere.rs` documents in prose (`ring_witness`'s simple-ring argument, the `|turning| < 2π` guard).
- **Spherely oracle fixtures** (#145 plan, phase 4): committed goldens generated from the C++ s2geometry via the public `create_polygon` route (validation-error text), non-degenerate domain only per the measured SoS-convention boundary.
- **#144's performance half**: the minor-principal-axis candidate for `ring_cap`, unlocking the closed-form point-in-ring path for the crescent family the vertex-sum heuristic misses. Correctness is unaffected by construction (`ring_cap` verifies every candidate); the exact GJK optimum stays documented-deferred.

## Phases

- [x] **Phase 1 — housekeeping.** Split the hemisphere-plus witness construction into `sphere/witness.rs` (PR #138 question 2, option (b), espg-approved): pure move, `sphere.rs` 1 775 → 1 579 lines — still above the ~1000-line soft limit; the split's scope is exactly the option (b) agreed on PR #138 question 2, not a claim of full compliance. Rewrite `docs/coverage_methods.md`'s ring-orientation section for decision (A), including the `normalize=False` hole-winding guidance from the #138 phase-5 fold. Landed in `bf052c6`, review folded in `cfe405e`.
[...]

PR #138, opened 2026-07-25
Repair the antipodal-lens defect in the point-in-ring reference layer (phase 1 of #107)

espg:
Refs #107 · Refs #113 · Refs #144 · Refs #145

Phase 1 of the (c) work agreed on [issue #107](https://github.com/espg/mortie/issues/107#issuecomment-5080472920) ("yes, start with repair `ring_winding_at` as phase 1"). It repairs the reference layer every later phase has to be adjudicated against; the crossing algebra itself is phase 2.

> **Rework note (second).** Both earlier versions of this phase shipped an *anchor construction* — step off an edge midpoint, then read the angle sum there — and review found four blocking defects in it, the last a whole-sphere inversion on a density-biased self-intersecting ring. Those were not four bugs. They were one, and the construction has been removed rather than guarded again. A third review round against the replacement found four more, all folded — see "Round-three review findings" below.

## The defect

`ring_winding_at` sums the signed angles each directed edge subtends at the test point. That sum is **antisymmetric under `x → −x`**:

```rust
let da = dot(a, x);
let pa = normalize(&[a[0] - da * x[0], a[1] - da * x[1], a[2] - da * x[2]]);
let sgn = dot(&cross(&pa, &pb), x);
total += if sgn >= 0.0 { ang } else { -ang };
```

`da` is odd in `x`, so `da * x` is even, so `pa`/`pb` — and therefore `ang = acos(pa·pb) ≥ 0` — are *invariant* under `x → −x`. Only `sgn` flips, so every term negates. The function returns

```text
w(x) = 2π · [ W(x) − W(−x) ]
```

not the winding indicator. The inside test was `w > π`, so **at most one member of any antipodal pair could test inside**: when a point and its antipode are both interior, both read outside. A sub-hemisphere interior cannot contain an antipodal pair, so the #103 87-case matrix is untouched; the broken regime is hemisphere-plus (#22), which is exactly what #107 is about.

## Why the first two attempts failed, and why it was one failure

`w` is a **difference** of winding numbers, and on a sphere that is all a ring defines by itself. A closed curve cuts the sphere into regions of equal standing, so `W` is fixed only up to one global additive constant — and "inside" *is* the choice of that constant. The right-hand rule (RFC 7946 §3.1.6 / S2) supplies it: the interior is to the **left** of the directed edges.

[...]

PR #182, opened 2026-08-09
Winding-free dissolve classifier: hemisphere+ covers dissolve instead of raising (issue #147)

### LINKED ISSUE #147: Winding-free dissolve classifier: hemisphere+ covers dissolve instead of raising

### LINKED ISSUE #181: dissolve: stitcher assert! crosses the FFI as PanicException instead of the curated ValueError
🤖 *from Claude* (filed at espg's direction, 2026-08-09 — the second follow-up named in [PR #179's decision comment](https://github.com/espg/mortie/pull/179#issuecomment-5230428447))

## Symptom

A pre-existing `assert!` in `dissolve.rs`'s stitching path (line 444 at PR #179's head) crosses the FFI as a raw `pyo3_runtime.PanicException` instead of the curated `ValueError` every other dissolve failure mode raises. The [PR #179 review sweep](https://github.com/espg/mortie/pull/179#discussion_r3743052040) measured its incidence across the 3,145-cover corpus: **348 covers pre-#179, 79 post** (deterministic chaining reduced but did not eliminate the path).

## The fix

Convert the assertion into the curated error convention: propagate an `Err` with a message naming the cover/segment condition, so Python callers get the same `ValueError` contract as the hemisphere guards (PR #111's precedent). The 79 remaining covers from the sweep are the test corpus — at least one becomes a pinned fixture asserting the curated message.

## Coordination

Same file as #147's classifier replacement. Sequencing is flexible: this is a small, self-contained change and can land first (a one-liner plus a test) or ride as an early commit of #147's PR — whichever the implementer of #147 finds cleaner, but it must not fall between the cracks; a raw panic crossing the FFI is a defect independent of the classifier's fate.

Refs: PR #179 (the sweep and the incidence numbers), #147 (same-file coordination), PR #111 (the curated-error convention this should match).

espg:
Closes #147. Closes #181 (the stitcher-error commit, per the [plan-delta comment](https://github.com/espg/mortie/issues/147#issuecomment-5230469706)). Refs #108, #111, #138, #146, #155 / PR #179.

Replaces `dissolve`'s mod-4π anchor-fan classifier with a winding-free classifier: hemisphere+ covers now dissolve to point-sampled-correct outlines instead of raising. Built on the #138/#146 instruments (`ring_set_validity`, `point_in_ring_robust`) and PR #179's deterministic chaining.

## Phases

- [x] **Phase 1 — orientation audit** (79ce0bf). Verdict below.
- [x] **#181 — stitcher `assert!` → curated `ValueError`** (491f424): both stitcher assertions became curated `Err`s (the old assert already surfaced as a `ValueError` via `rust_dissolve`'s catch_unwind shim, but with raw assert text plus a panic trace on stderr); the Python oracle's `RuntimeError`s became matching `ValueError`s. Phase 2's local pole rule then made the unbalanced-no-pole state structurally unreachable — the fixture cover (base cells {1,2,7}) now *dissolves* — so #181's surviving contract is: every remaining defensive stitch state (non-convergence, missing partner after a pole seam) is a curated `ValueError`, pinned by `stitch_missing_partner_errs_curated` on both engines.
- [x] **Phase 2 — the winding-free classifier, Rust** (96eeba7 + 3cfb4e6): see Approach.
- [x] **Phase 3 — Python oracle mirror** (eef306e): `_dissolved_rings_py` implements the same classifier (calibration via `_emission_is_ccw`, gate via `mortie.ring_is_simple`, planar classification, frame rule, local pole stitching), same fixtures, differential-tested at corpus scale.
- [x] **Phase 4 — fixtures, goldens, sweep, round-trip** (fe15b43 + 3cfb4e6): all PR #111 fail-loud fixtures flip to working; polar caps at 96%/100%/104% of 2π (the 96% cap regression-pinned as working again); equatorial band; world-minus-hole (frame shell); whole-sphere; multipart both-polar-caps; the two sweep exemplars as named fixtures; the **regenerated sweep corpus** (below); spherely goldens; round-trip attempt (below).
- [x] **Phase 5 — divergence report** (below). Benches: informational only — **no dissolve benches exist** in `src_rust/benches/` or `benchmarks/`, so CodSpeed has nothing to say about this PR beyond its non-gating status (espg ruling 2026-08-09).

## Phase-1 verdict: cover-on-the-left holds, up to one per-call calibration

Pinned by `cell_boundary_emission_orientation_is_uniform` / `chained_rings_carry_cover_on_left` (Rust) and their Python mirrors in `mortie/tests/test_dissolve_hemisphere.py`:

- The HEALPix boundary functions emit every cell loop with a fixed handedness — interior on the **left** at `step == 1` (CCW, positive spherical fan area) and on the **right** at `step > 1` — uniformly over all 12 base cells, orders 0–5, steps {1,2,3,4,8} (the previously-informal comment in `classify_and_split` — "boundary point order differs between step==1 and step>1" — measured and pinned).
- Each surviving directed edge bounds exactly one covered cell on the emission side; chaining preserves edge direction; so **one per-call reversal** (iff one cell's own loop reads CW) puts the covered region on every ring's left — a *local* fact, valid at any cover scale. Verified end-to-end by displaced-midpoint probes (left of every ring edge covered, right uncovered) on exact-hemisphere, over-hemisphere, world-minus-cell and the sweep exemplar covers, both step regimes, both engines.

Phase 2 consumes the **contract, not the per-ring probe fallback**; the probe survives as `debug_assert_cover_on_left`, the debug-build residue of the retired PR #111 guards (see below).

## Approach

With covered-on-left established per ring:

1. **Validity gate** — `ring_set_validity`'s `crossing` half: a self-crossing boundary ring keeps the curated fail-loud `ValueError`. **Deviation from the issue's letter:** `identity_conflict` is *not* gated — a repeated snapped coordinate at non-adjacent positions is the *working representation* of a corner-touch pinch resolved into touching simple rings (the #155 ruling), and gating it would regress the pinned `test_dissolve_corner_touch_yields_simple_rings`. Flagged under Questions for review.
2. **Classification is planar, not spherical nesting parity.** The issue sketched O(R²) sphere-side nesting parity via `point_in_ring_robust`; that sketch is ill-defined on the sphere — for an annulus, the shell's representative point lies in the *left region* of the covered-on-left hole ring, so left-side parity counts misclassify the shell (there is no absolute "inside" on a sphere without a reference point). The plane has one: after the antimeridian cut/stitch, a planar ring with the covered region on its left is an exterior iff its shoelace is positive. Equivalent to planar nesting parity, O(R·V), and decided exactly where the output lives (GeoJSON planar rings). Flagged under Questions for review.
3. **Frame rule** — Σ shoelace over the emitted rings is the covered region's planar area when its boundary is complete, and that minus the whole map when the covered region encloses the frame (world-minus-hole, whole-sphere): separated by sign; negative Σ prepends the explicit `[-180,180]×[-90,90]` shell.
4. **Local pole stitching** — the global net-winding `pole` parameter is gone: covered-on-left means the covered seam interval always runs upward from a cut end on +180 and downward on −180, so an unmatched end wraps the pole on its side, unconditionally; one ring may wrap both poles. This is what unlocks multipart hemisphere-straddling covers (both-polar-caps; sweep exemplar {1,2,7}) — and it fixes a **latent live bug**: the old `want_min` arm picked the *minimum*-latitude start after a north-pole wrap, fusing a north cap with any other seam-crossing part into one invalid ring (reproduced on main with a cap+box cover, `north_pole_wrap_picks_nearest_start` pins the fix).
5. **Retired guards** — `HEMISPHERE_MARGIN` and the Σ-vs-exact-area wrap cross-check are deleted; their debug-assertion residue is (a) `debug_assert_cover_on_left` (the orientation fact the guards actually protected, checked by exact membership probes) and (b) the planar-area budget assert in `classify_and_split` (emitted Σ + frame ∈ (0, whole map]). A Σ-fan-area debug check was tried and abandoned: the fan's wraps are anchor-dependent and **not** clean 4π multiples (PR #179 measured 4.0899 vs 0.7901 sr fans for one ring), so no area identity survives at hemisphere+ scale.
6. **Pole-vertex planar expansion** (`ring_to_planar`) — fixes a **pre-existing silent defect**: a boundary vertex at a pole was emitted as the single planar point `atan2(0,0) = (0, ±90)`, slicing planar area off any cover whose boundary passes *through* a pole — on main, base cells {0,1} at order 1 mis-classify 49/144 near-pole probe points today. A pole vertex now expands to a lat-±90 cap over the wedge of longitudes the passage encloses, chosen by the **empty-bracket rule** (the interval between the arriving and departing meridians containing no other pole-incident meridian; covered side on degree-2 ties), subdivided so no planar step reaches 180°, with exact ±180 seam pairs where the cap crosses the seam.
7. **Seam-edge normalization** (`normalized_lonlat`) — boundary edges lying exactly on the ±180 meridian (base-cell meridians) previously took arbitrary `atan2` signs per vertex, fabricating spurious ±360° cuts; the sign now follows traversal direction (northward ⟹ covered west ⟹ map east, +180).
8. **Pinch splitting** (`split_at_revisits` + `split_planar_at_revisits`) — a chained walk revisiting a vertex (the covered region pinched at a point, e.g. two uncovered cells touching diagonally — the dual of #155's corner-touch) splits into simple rings touching at that point, id-level and again planar-level, so the emit is OGC-simple where the old single walk was shapely-invalid.

No new dependencies; `sphere.rs` untouched (`ring_set_validity` was already `pub`). **Module layout:** `dissolve.rs` was heading past the ~1000-line cap with its tests (1276 lines), so the test module moved to `src_rust/src/dissolve/tests.rs` following the repo's own `sphere.rs` / `sphere/tests.rs` layout — body now 935 lines, tests 736. Flagged under Questions for review.

## The regenerated sweep corpus (phase 4 acceptance)

`sweep_corpus_order1_combos_dissolve_point_sampled` regenerates the [PR #179 review sweep](https://github.com/espg/mortie/pull/179#discussion_r3743052040)'s coarse family and — per the adversarial review's blocking finding — extends it to **all 4095 non-empty order-1 base-cell masks** (the sweep's 1585 one-to-five-cell masks plus the 6-to-12-cell hemisphere+ regime this issue exists for), first 60 also at step 3; every one dissolves with each emitted region point-sampled against exact cover membership at all 768 order-3 cell centres (chord-discretization band exempted near edges). This is a superset of the 249 previously-guard-refused covers *and* the 79 stitcher-panic covers: under the old classifier the sweep family contains 194 deterministic wrap-guard rejections + 79 stitcher failures at step 1 (measured below), and everything past 5 cells was area-guard-rejected wholesale; **all now dissolve, none regress**. `sweep_corpus_fine_order_families_dissolve` regenerates the order-2–5 runs (300/order) + 300 scattered order-3 sets; `mixed_order_hemisphere_moc_dissolves` covers a mixed-order hemisphere+ MOC (review fold). The two exemplars ({4,6,7,10} and {1,2,7} at order 1) are named pinned fixtures on both engines. Whole-corpus runtime ~4.4 s in a debug `cargo test`.

## Phase-5 divergence report

Old = `d4e62d0` (origin/main) built in an isolated venv; new = this head. Corpus = the regenerated sweep (1585 masks + 60 step-3 + 1500 fine-order covers) + the dissolve test-corpus covers (box/holed/antimeridian/caps/annulus/band/moc/hemisphere family) = **3158 runs**. Verdicts compared per cover; `ok` outputs compared byte-wise, then geometrically (shapely symmetric difference, `make_valid`-guarded).

| old → new | runs | note |
|---|---|---|
| reject (Σ-wrap guard) → outline | 195 | 194 order-1 masks + the equatorial band |
| stitcher failure (`unbalanced…no pole`) → outline | 80 | 79 order-1 masks + both-polar-caps |
| reject (area guard) → outline | 3 | over-hemisphere cap, world-minus-one, whole-sphere |
| ok → identical outline | 936 | byte-identical |
| ok → equivalent outline, representation only | 229 | symdiff < 1e-9 (canonical starts, split representation) |
| ok → different outline | 1594 | **all differences are corrections** — see below |
| ok → reject (regressions) | **0** | |
| still failing under new | **0** | |

The 1594 geometric differences decompose as: **802 covers whose old emit shapely was invalid** (plus 121 more whose old emit could not even be nested into polygons — `_nest_and_build` raised); and covers with pointwise-wrong old emits from the pre-existing pole-vertex/seam-edge defects — including **silent complement inversions the old guards never caught**: e.g. base cells {1,2,3,6,7} (5.24 sr, sub-hemisphere, mask 206) passed every old guard and emitted its exact complement (all 12 base-cell probes inverted); now correct and pinned (`test_sweep_silent_inversion_mask_206_now_correct`). Every new outline in the corpus is point-sampled correct. Two caveats from the adversarial review: the "old" verdicts here are single samples of an engine that is still per-call nondeterministic on some coarse masks even post-#179 (e.g. mask 613 gave 7 valid + 1 reject over 8 in-process calls on `d4e62d0`; this build is 8/8 deterministic), so old-side counts are representative rather than exact; and new-emit *shapely validity* holds on 3149/3158 of this corpus but degrades on the extended full-mask family — see Questions for review (3).

## spherely goldens (phase 4)

`test_dissolve_outline_areas_match_spherely` (importorskip, live s2 like the existing cap test): Σ shell − Σ hole areas of the emit, frame counting 4π, vs the cover's exact area — pinned per fixture to 1e-6 against 2026-08-09 goldens, e.g. hemisphere 6.283185307 (frac 1.000000), exemplar {1,2,7} 3.141592654 (frac 1.000000), world-minus-one 12.500225 (frac 0.999944), and the order-1 chord-deficit cases at frac 0.976–1.032 — an inverted classification would sit near (4π−A)/A, far outside the 5% band asserted. Scope note from the review: this s2 route does not generalise to arbitrary corpus covers (s2 refuses or mis-measures ~3% of full-mask emits whose planar seam/pole doubling has no clean spherical reading), so the goldens stay a named-fixture oracle, not a corpus-wide one.

## Round-trip (open question 2) — attempted, honest result

`cover → to_geometry → from_geometry(..., order, normalize=False)`:

- **Exact set equality cannot hold**: coverage is an *overlap* cover, so boundary-touching neighbour cells always join the round trip (measured: `missing 0, extra ≈ boundary ring` on every hole-free case).
- The **containment half holds for every hole-free emit with a real boundary, hemisphere+ included**, and is pinned (`test_roundtrip_coverage_contains_cover_normalize_false`). One exception found by the review: the whole-sphere cover's frame-only emit (a spherically degenerate seam-and-poles ring) loses most cells through `from_geometry` — the frame ring is a planar convention with no faithful spherical reading, so the whole-sphere case is excluded from the pinned invariant (review measured 3571/3572 other hole-free emits round-tripping).
- **Emits with holes break under `normalize=False`**: the parity fill reads a CW-wound (RFC 7946) hole ring as selecting its complement, re-filling the hole (world-minus-one comes back with the hole cell; a holed box floods its exterior). Under `normalize=True` holed sub-hemisphere covers round-trip to supersets, but hemisphere+ shells get re-normalized to the wrong side. So no single flag gives a universal invariant; the blocker is the hole-winding convention mismatch between the RFC 7946 emit and the `normalize=False` parity fill — left as a finding, not pinned.

## How tested

Green gate at head: `maturin develop --release`; `cargo test --lib` ×3 (360 passed each — dissolve corpus included, chaining determinism re-exercised); `cargo fmt --check`; `cargo clippy --lib --tests` (0 warnings in dissolve; the 6 pre-existing sites elsewhere untouched); `flake8 mortie --select=E9,F63,F7,F82`; `pytest` (1434 passed, 16 skipped); `numpydoc lint mortie/*.py` clean.

## Questions for review

1. **Identity-conflict gate not applied** (deviation from the issue's phase-2 item 1): gating `ring_set_validity.identity_conflict` would reject the corner-touch covers that dissolve correctly today (touching simple rings *are* the pinch representation per #155). Only the `crossing` half gates. Flagging in case the letter was wanted.
2. **Planar classification instead of spherical O(R²) nesting parity** (deviation from the issue's phase-2 item 3): the spherical sketch has the annulus counterexample above; the planar shoelace form is the equivalent well-defined statement in the output's own space. Happy to add a debug-assertion cross-check against planar point-in-ring nesting if wanted.
3. **Residual shapely-invalidity: parity-correct, self-touching emits on covers whose boundary *touches* the ±180 seam at a point.** The adversarial review sized and diagnosed this properly (my initial "9 of 3158" was an artifact of the original 1–5-cell corpus cap, since removed): over all 4095 order-1 masks, **338 emits at step 1 (715 at step 3) are shapely-invalid**, with smaller rates at finer orders (5.5–8.5% of order-2–4 half-space covers). Mechanism ([review comment](https://github.com/espg/mortie/pull/182#discussion_r3743436698)): `normalized_lonlat` pins an isolated seam-touch vertex onto one side, and a stitcher meridian connector then runs straight *through* that vertex — a self-touch, never a parity error (review: 0 parity failures on every invalid mask probed; `Σ shoelace == shapely.area` to 1e-6 on all 4095 masks; `make_valid` recovers them). Every one of these covers previously **raised or emitted invalid/inverted geometry**, so nothing regresses — but downstream `union`-style consumers can reject self-touching input. Options, one-pass: (a) accept + follow-up issue: the fix is localized (split stitch connectors at seam-lying ring vertices so the pieces touch at a point instead of overlapping through it), but it is another stitcher surgery on top of an already-large PR; (b) require it in this PR. My lean is (a), with the follow-up carrying the review's measured corpus as its acceptance set.
4. **Test-module split**: `dissolve.rs` would have crossed the ~1000-line cap with its tests; moved them to `dissolve/tests.rs` per the existing `sphere/tests.rs` layout (§4 discussion-first caveat — raising here rather than pre-landing, since it is layout-only and mirrors in-repo precedent).
5. The **chord-discretization band** in the point samplers (mismatches allowed only within `4°/2^order` of the emitted boundary) is the honest contract for a chord emit; tightening it means densifying `step` in the fixtures.
6. **Pre-existing, out of scope, surfaced by the review** ([comment](https://github.com/espg/mortie/pull/182#discussion_r3743437768)): corner-touch pinches resolve into shell+hole (interior-disconnected, invalid) at every `step > 1`, on released 0.9.6 exactly as on this branch, both engines — the step>1 azimuthal orderings at the pinch vertex differ from step 1. Candidate follow-up issue (not opened unilaterally); no `step > 1` pinch test exists today.

### DIFF
--- a/mortie/dissolve.py
+++ b/mortie/dissolve.py
@@ -259,9 +259,13 @@ def _split_at_revisits(chain):
     diagonally, so the covered face's own boundary passes the pinch twice)
     the traced walk legitimately revisits that vertex — a correct even-odd
     boundary, but not an OGC-simple ring.  Splitting at each revisit yields
-    the equivalent decomposition into simple rings touching at a point.
-    Each cycle is rotated to start at its minimal vertex id.  Mirrors
-    ``split_at_revisits`` in ``src_rust/src/dissolve.rs`` (issue #147).
+    the equivalent decomposition into simple rings touching at a point.  A
+    split at a *pole* vertex re-pairs that pole's passages onto their
+    adjacent wedges; :func:`_ring_to_planar`'s empty-bracket span rule
+    renders each re-paired passage over exactly its own wedge, keeping the
+    planar shoelace budget consistent.  Each cycle is rotated to start at
+    its minimal vertex id.  Mirrors ``split_at_revisits`` in
+    ``src_rust/src/dissolve.rs`` (issue #147).
 
     Parameters
     ----------
@@ -356,16 +360,64 @@ def _normalized_lonlat(ring_xyz):
     return lat, lon
 
 
-def _ring_to_planar(ring_xyz):
+def _pole_run(out, start, stop, pole):
+    """Append a monotone lat-±90 run from *start* to *stop* (exclusive of
+    *start*), subdivided so no planar step reaches 180° — a covered polar
+    wedge can legitimately span more than 180° of longitude, which
+    :func:`_cut_at_antimeridian` would misread as a seam wrap if emitted as
+    one step.  Mirrors ``pole_run`` in ``src_rust/src/dissolve.rs``.
+    """
+    direction = 1.0 if stop > start else -1.0
+    cur = start
+    while abs(stop - cur) > 150.0:
+        cur += direction * 120.0
+        out.append((cur, pole))
+    out.append((stop, pole))
+
+
+def _pole_meridians(rings_xyz):
+    """Pole-incident boundary meridians over a ring set.
+
+    For every pole vertex, the arriving and departing neighbours'
+    (normalized) longitudes.  Feeds :func:`_ring_to_planar`'s empty-bracket
+    cap rule.  Mirrors the collection in ``classify_and_split``
+    (``src_rust/src/dissolve.rs``).
+
+    Parameters
+    ----------
+    rings_xyz : list of numpy.ndarray
+        The oriented boundary rings.
+
+    Returns
+    -------
+    tuple of list
+        ``(north, south)`` meridian longitude lists (degrees).
+    """
+    north, south = [], []
+    for ring in rings_xyz:
+        n = len(ring)
+        _, lon = _normalized_lonlat(ring)
+        for i in range(n):
+            if abs(ring[i][2]) > 1.0 - 1e-9:
+                dst = north if ring[i][2] > 0 else south
+                dst.append(float(lon[(i - 1) % n]))
+                dst.append(float(lon[(i + 1) % n]))
+    return north, south
+
+
+def _ring_to_planar(ring_xyz, north_m, south_m):
     """Convert an oriented spherical ring to planar ``(lon, lat)`` vertices.
 
     Expands any pole vertex into an explicit lat-±90 traverse: a vertex at
     the pole has no longitude (``atan2(0, 0)``); its planar image is a
-    lat-±90 *edge* from the arriving meridian's longitude to the departing
-    one's, run in the direction that keeps the covered region on the left —
-    westward across the top at +90, eastward across the bottom at -90 —
-    inserting an explicit ±180 seam pair when the direct step would run the
-    wrong way round.  Mirrors ``ring_to_planar`` in
+    lat-±90 *edge* — a cap over the wedge of longitudes this passage
+    encloses at the pole.  Which wedge is the **empty-bracket rule**: of the
+    two lon-intervals between the arriving and departing meridians, the
+    enclosed one contains no *other* pole-incident boundary meridian; at a
+    degree-2 pole both are empty and the covered side wins (westward of the
+    arrival at +90, eastward at -90).  A cap through the seam inserts an
+    exact ±180 pair that :func:`_cut_at_antimeridian` splits without
+    interpolation.  Mirrors ``ring_to_planar`` in
     ``src_rust/src/dissolve.rs`` (issue #147).
 
     Parameters
@@ -373,6 +425,8 @@ def _ring_to_planar(ring_xyz):
     ring_xyz : numpy.ndarray
         An ``(M, 3)`` array of unit vectors (open ring, covered region on
         the left).
+    north_m, south_m : list of float
+        The pole-incident meridians from :func:`_pole_meridians`.
 
     Returns
     -------
@@ -385,18 +439,37 @@ def _ring_to_planar(ring_xyz):
     for i in range(n):
         if abs(ring_xyz[i][2]) > 1.0 - 1e-9:
             pole = 90.0 if ring_xyz[i][2] > 0 else -90.0
-            lon_arr = float(lon[(i - 1) % n])
-            lon_dep = float(lon[(i + 1) % n])
-            out.append((lon_arr, pole))
-            direct_ok = lon_dep < lon_arr if pole > 0 else lon_dep > lon_arr
-            if not direct_ok:
-                # through the seam: an exact ±180 crossing that
-                # `_cut_at_antimeridian` splits without interpolation.
-                if pole > 0:
-                    out.extend([(-180.0, pole), (180.0, pole)])
+            arr = float(lon[(i - 1) % n])
+            dep = float(lon[(i + 1) % n])
+            meridians = north_m if pole > 0 else south_m
+
+            def rel_w(x):
+                return (arr - x) % 360.0
+
+            def rel_e(x):
+                return (x - arr) % 360.0
+
+            dw = rel_w(dep) or 360.0
+            de = rel_e(dep) or 360.0
+            west_clear = all(
+                rel_w(m) == 0.0 or rel_w(m) >= dw for m in meridians)
+            east_clear = all(
+                rel_e(m) == 0.0 or rel_e(m) >= de for m in meridians)
+            go_west = (pole > 0) if (west_clear and east_clear) else west_clear
+            out.append((arr, pole))
+            if go_west:
+                if dep < arr:
+                    _pole_run(out, arr, dep, pole)
                 else:
-                    out.extend([(180.0, pole), (-180.0, pole)])
-            out.append((lon_dep, pole))
+                    _pole_run(out, arr, -180.0, pole)
+                    out.append((180.0, pole))
+                    _pole_run(out, 180.0, dep, pole)
+            elif dep > arr:
+                _pole_run(out, arr, dep, pole)
+            else:
+                _pole_run(out, arr, 180.0, pole)
+                out.append((-180.0, pole))
+                _pole_run(out, -180.0, dep, pole)
         else:
             out.append((float(lon[i]), float(lat[i])))
     return out
@@ -433,10 +506,14 @@ def _cut_at_antimeridian(coords):
         cur.append((lo0, la0))
         if abs(lo1 - lo0) > 180.0:
             lo1u = lo1 - 360.0 if lo1 > lo0 else lo1 + 360.0
-            boundary = 180.0 if lo1u > lo0 else -180.0
             # An exact ±180 → ∓180 pair (a pole traverse's explicit seam
-            # crossing, `_ring_to_planar`) unwraps to a zero-length step;
-            # there is nothing to interpolate.
+            # crossing, `_ring_to_planar`) unwraps to a zero-length step:
+            # nothing to interpolate, and the cut side is the pair's own
+            # first side (the generic ``lo1u > lo0`` test cannot tell).
+            if lo1u == lo0:
+                boundary = lo0
+            else:
+                boundary = 180.0 if lo1u > lo0 else -180.0
             frac = 0.0 if lo1u == lo0 else (boundary - lo0) / (lo1u - lo0)
             la_x = la0 + frac * (la1 - la0)
             cur.append((boundary, la_x))
@@ -634,6 +711,46 @@ def _planar_signed_area(ring):
     return 0.5 * float(np.sum(x * np.roll(y, -1) - np.roll(x, -1) * y))
 
 
+def _split_planar_at_revisits(ring):
+    """Split a closed planar ring at exactly-repeated vertices.
+
+    The planar mirror of :func:`_split_at_revisits`, keyed on exact
+    coordinates; drops degenerate two-vertex slivers.  Mirrors
+    ``split_planar_at_revisits`` in ``src_rust/src/dissolve.rs`` (issue
+    #147).
+
+    Parameters
+    ----------
+    ring : list of tuple
+        A closed list of ``(lon, lat)`` degree pairs.
+
+    Returns
+    -------
+    list of list of tuple
+        Simple closed rings (each last vertex repeats the first).
+    """
+    out = []
+    stack = []
+    pos = {}
+
+    def emit(cycle):
+        if len(cycle) >= 3:
+            out.append(cycle + [cycle[0]])
+
+    for p in ring[:-1]:
+        if p in pos:
+            at = pos[p]
+            cycle = stack[at:]
+            del stack[at:]
+            for c in cycle:
+                pos.pop(c, None)
+            emit(cycle)
+        pos[p] = len(stack)
+        stack.append(p)
+    emit(stack)
+    return out
+
+
 def _frame_ring():
     """Return the whole-map frame shell, CCW (positive shoelace).
 
@@ -736,20 +853,29 @@ def _dissolved_rings_py(morton, step):
     ext_pieces = []
     holes = []
     segments = []
+
+    def classify(ring):
+        # A planar ring can revisit a vertex exactly — e.g. a ring that both
+        # expands a pole traverse and wraps the same pole's seam passes the
+        # ±180/±90 corner twice — which even-odd fill reads correctly but
+        # OGC does not allow in one ring; split into simple rings touching
+        # at the point.
+        for piece in _split_planar_at_revisits(ring):
+            (ext_pieces if _planar_signed_area(piece) >= 0.0 else holes).append(
+                piece
+            )
+
+    north_m, south_m = _pole_meridians(rings_xyz)
     for ring in rings_xyz:
-        ll = _ring_to_planar(ring)
+        ll = _ring_to_planar(ring, north_m, south_m)
         whole, segs = _cut_at_antimeridian(ll)
         if whole is not None:
-            (ext_pieces if _planar_signed_area(whole) >= 0.0 else holes).append(
-                whole
-            )
+            classify(whole)
         else:
             segments.extend(segs)
     if segments:
         for piece in _stitch_segments(segments):
-            (ext_pieces if _planar_signed_area(piece) >= 0.0 else holes).append(
-                piece
-            )
+            classify(piece)
 
     # Frame rule: Σ shoelace is the covered region's planar area when its
     # boundary is complete, and that minus the full map when the region
--- a/mortie/dissolve.py
+++ b/mortie/dissolve.py
@@ -348,9 +348,10 @@ def _stitch_segments(segments, pole):
 
     Raises
     ------
-    RuntimeError
+    ValueError
         If the stitch fails to converge (a guard), or if the segments are
-        unbalanced with no pole enclosed.
+        unbalanced with no pole enclosed — the curated fail-loud convention
+        (mirrors ``src_rust/src/dissolve.rs``, issue #181).
     """
     segs = [list(s) for s in segments]
     used = [False] * len(segs)
@@ -364,7 +365,11 @@ def _stitch_segments(segments, pole):
         while idx is not None and not used[idx]:
             guard += 1
             if guard > 8 * len(segs) + 16:  # pragma: no cover - convergence guard
-                raise RuntimeError("antimeridian stitch did not converge")
+                raise ValueError(
+                    "dissolved cover's antimeridian stitch did not converge "
+                    "(an internal dissolve inconsistency); pass dissolve=False "
+                    "for per-cell polygons"
+                )
             used[idx] = True
             ring.extend(segs[idx])
             idx = _next_segment(segs, used, ring, pole, seed)
@@ -406,8 +411,10 @@ def _next_segment(segs, used, ring, pole, seed):
 
     Raises
     ------
-    RuntimeError
-        If the segments are unbalanced but no pole is enclosed.
+    ValueError
+        If the segments are unbalanced but no pole is enclosed — the curated
+        fail-loud convention (mirrors ``src_rust/src/dissolve.rs``, issue
+        #181).
     """
     side, end_lat = ring[-1]
     cands = [(segs[i][0][1], i) for i in range(len(segs))
@@ -427,8 +434,13 @@ def _next_segment(segs, used, ring, pole, seed):
 
     # No same-side start in that direction: the region wraps ``pole``.  Run the
     # seam to the pole, cross to the other side, and resume from the pole.
-    if pole == 0:  # pragma: no cover - guarded by the caller's pole detection
-        raise RuntimeError("unbalanced antimeridian segments but no pole enclosed")
+    if pole == 0:
+        raise ValueError(
+            "dissolved cover's antimeridian segments are unbalanced but no "
+            "pole is enclosed, so the stitch cannot close the outline; split "
+            "the cover into smaller parts or pass dissolve=False for per-cell "
+            "polygons (issue #181)"
+        )
     other = -side
     ring.append((side, pole))
     ring.append((other, pole))
--- a/mortie/dissolve.py
+++ b/mortie/dissolve.py
@@ -361,11 +361,13 @@ def _normalized_lonlat(ring_xyz):
 
 
 def _pole_run(out, start, stop, pole):
-    """Append a monotone lat-±90 run from *start* to *stop* (exclusive of
-    *start*), subdivided so no planar step reaches 180° — a covered polar
-    wedge can legitimately span more than 180° of longitude, which
-    :func:`_cut_at_antimeridian` would misread as a seam wrap if emitted as
-    one step.  Mirrors ``pole_run`` in ``src_rust/src/dissolve.rs``.
+    """Append a monotone lat-±90 run from *start* to *stop*.
+
+    Exclusive of *start*, subdivided so no planar step reaches 180° — a
+    covered polar wedge can legitimately span more than 180° of longitude,
+    which :func:`_cut_at_antimeridian` would misread as a seam wrap if
+    emitted as one step.  Mirrors ``pole_run`` in
+    ``src_rust/src/dissolve.rs``.
     """
     direction = 1.0 if stop > start else -1.0
     cur = start
--- a/mortie/dissolve.py
+++ b/mortie/dissolve.py
@@ -246,29 +246,160 @@ def _chain_rings(survivors, id_xyz):
                 # walk on the same face (no crossing) at a non-manifold vertex.
                 back = _tangent_azimuth(id_xyz[v], id_xyz[cur[0]])
                 cur = min(cand, key=lambda r: (az[(r[0], r[1])] - back) % (2 * math.pi))
-        rings.append(id_xyz[np.asarray(chain)])
+        for cycle in _split_at_revisits(chain):
+            rings.append(id_xyz[np.asarray(cycle)])
     return rings
 
 
-def _antimeridian_winding(lon):
-    """Measure a ring's net longitude winding and antimeridian crossings.
+def _split_at_revisits(chain):
+    """Split a closed vertex-id walk into minimal simple cycles.
+
+    The angular chaining traces boundaries of the *covered* faces; where the
+    covered region is pinched at a vertex (two uncovered cells touching only
+    diagonally, so the covered face's own boundary passes the pinch twice)
+    the traced walk legitimately revisits that vertex — a correct even-odd
+    boundary, but not an OGC-simple ring.  Splitting at each revisit yields
+    the equivalent decomposition into simple rings touching at a point.
+    Each cycle is rotated to start at its minimal vertex id.  Mirrors
+    ``split_at_revisits`` in ``src_rust/src/dissolve.rs`` (issue #147).
 
     Parameters
     ----------
-    lon : numpy.ndarray
-        The longitudes (degrees) of a closed ring, first vertex not repeated.
+    chain : list of int
+        The closed walk as vertex ids (first vertex not repeated).
 
     Returns
     -------
-    tuple
-        ``(net, crossings)`` — the net signed longitude winding in degrees and
-        the antimeridian-crossing count.  Net ≈ ±360 ⟺ the ring encircles a
-        pole.
+    list of list of int
+        The simple cycles (each open, first vertex not repeated).
     """
-    deltas = np.diff(np.concatenate([lon, lon[:1]]))
-    crossings = int(np.sum(np.abs(deltas) > 180.0))
-    net = float(np.sum((deltas + 180.0) % 360.0 - 180.0))
-    return net, crossings
+    out = []
+    stack = []
+    pos = {}
+
+    def _rotated(cycle):
+        k = cycle.index(min(cycle))
+        return cycle[k:] + cycle[:k]
+
+    for vid in chain:
+        if vid in pos:
+            p = pos[vid]
+            cycle = stack[p:]
+            del stack[p:]
+            for c in cycle:
+                pos.pop(c, None)
+            out.append(_rotated(cycle))
+        pos[vid] = len(stack)
+        stack.append(vid)
+    if stack:
+        out.append(_rotated(stack))
+    return out
+
+
+def _normalized_lonlat(ring_xyz):
+    """Per-vertex ``(lat, lon)`` with seam vertices' longitude signs fixed.
+
+    A vertex exactly on the ±180° meridian gets an arbitrary sign from
+    ``atan2`` (its ``y`` is a rounding residual), and a whole boundary edge
+    can lie on the seam (base-cell meridians at lon 180) — mixed signs there
+    fabricate spurious ±360° cuts mid-edge.  With the covered region on the
+    ring's left, a seam-lying edge walked northward has the covered side
+    west, so it belongs on the map's east edge (+180°); southward, on -180°.
+    A vertex with a seam (or pole) neighbour takes that edge's direction; an
+    isolated seam touch takes its off-seam neighbours' side.  Mirrors
+    ``normalized_lonlat`` in ``src_rust/src/dissolve.rs`` (issue #147).
+
+    Parameters
+    ----------
+    ring_xyz : numpy.ndarray
+        An ``(M, 3)`` array of unit vectors (open ring).
+
+    Returns
+    -------
+    tuple of numpy.ndarray
+        The ``(lat, lon)`` degree arrays with normalized seam longitudes.
+    """
+    lat, lon = _xyz_to_latlon(ring_xyz)
+    lon = lon.copy()
+    n = len(ring_xyz)
+
+    def is_pole(j):
+        return abs(ring_xyz[j][2]) > 1.0 - 1e-9
+
+    def on_seam(j):
+        return not is_pole(j) and abs(abs(orig[j]) - 180.0) < 1e-9
+
+    def seam_lat(j):
+        # effective seam latitude when the edge to neighbour j runs along
+        # the seam (a pole neighbour's meridian edge is the seam meridian).
+        if is_pole(j):
+            return 90.0 if ring_xyz[j][2] > 0 else -90.0
+        if on_seam(j):
+            return float(lat[j])
+        return None
+
+    orig = lon.copy()
+    for i in range(n):
+        if not on_seam(i):
+            continue
+        prev, nxt = (i - 1) % n, (i + 1) % n
+        la = seam_lat(nxt)
+        if la is not None:
+            side = 180.0 if la > lat[i] else -180.0
+        else:
+            la = seam_lat(prev)
+            if la is not None:
+                side = 180.0 if lat[i] > la else -180.0
+            else:
+                side = 180.0 if orig[prev] >= 0.0 else -180.0
+        lon[i] = side
+    return lat, lon
+
+
+def _ring_to_planar(ring_xyz):
+    """Convert an oriented spherical ring to planar ``(lon, lat)`` vertices.
+
+    Expands any pole vertex into an explicit lat-±90 traverse: a vertex at
+    the pole has no longitude (``atan2(0, 0)``); its planar image is a
+    lat-±90 *edge* from the arriving meridian's longitude to the departing
+    one's, run in the direction that keeps the covered region on the left —
+    westward across the top at +90, eastward across the bottom at -90 —
+    inserting an explicit ±180 seam pair when the direct step would run the
+    wrong way round.  Mirrors ``ring_to_planar`` in
+    ``src_rust/src/dissolve.rs`` (issue #147).
+
+    Parameters
+    ----------
+    ring_xyz : numpy.ndarray
+        An ``(M, 3)`` array of unit vectors (open ring, covered region on
+        the left).
+
+    Returns
+    -------
+    list of tuple
+        The open planar ring as ``(lon, lat)`` degree pairs.
+    """
+    lat, lon = _normalized_lonlat(ring_xyz)
+    n = len(ring_xyz)
+    out = []
+    for i in range(n):
+        if abs(ring_xyz[i][2]) > 1.0 - 1e-9:
+            pole = 90.0 if ring_xyz[i][2] > 0 else -90.0
+            lon_arr = float(lon[(i - 1) % n])
+            lon_dep = float(lon[(i + 1) % n])
+            out.append((lon_arr, pole))
+            direct_ok = lon_dep < lon_arr if pole > 0 else lon_dep > lon_arr
+            if not direct_ok:
+                # through the seam: an exact ±180 crossing that
+                # `_cut_at_antimeridian` splits without interpolation.
+                if pole > 0:
+                    out.extend([(-180.0, pole), (180.0, pole)])
+                else:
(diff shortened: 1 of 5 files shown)


mortie/dissolve.py
```diff
@@ -246,29 +246,233 @@ def _chain_rings(survivors, id_xyz):
...
+    return lat, lon
+
+
+def _pole_run(out, start, stop, pole):
+    """Append a monotone lat-±90 run from *start* to *stop* (exclusive of
+    *start*), subdivided so no planar step reaches 180° — a covered polar
+    wedge can legitimately span more than 180° of longitude, which
+    :func:`_cut_at_antimeridian` would misread as a seam wrap if emitted as
+    one step.  Mirrors ``pole_run`` in ``src_rust/src/dissolve.rs``.
+    """
```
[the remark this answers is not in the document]
espg: 🤖 *from Claude*

Fixed in 83fbfeb — summary compressed to one line with the description in its own paragraph.

src_rust/src/dissolve/tests.rs
```diff
@@ -0,0 +1,736 @@
...
+    // The PR #179 review sweep's coarse family, regenerated: every
+    // 1-to-5-base-cell mask at order 1 (1585 covers).  Under the old
+    // classifier 249 of them deterministically failed the PR #111 guards
+    // and 79 reached the stitcher assert; the winding-free classifier must
+    // dissolve every one, each emitted region point-sampled against the
+    // source cover (issue #147 plan delta).  The first 60 masks also run at
+    // step 3, mirroring the sweep's step slice.
+    let mut count = 0u32;
+    for mask in 1u16..(1 << 12) {
+        if mask.count_ones() > 5 {
```
espg: 🤖 *from Claude (review)*

**blocking — the acceptance corpus skips the hemisphere+ regime this issue exists for, and question 3's "9 of 3158, all coarse pole-pinch" is an artifact of that skip: the real order-1 residual is 338/4095 at step 1 and 715/4095 at step 3.**

`mask.count_ones() > 5` caps the regenerated corpus at 5 base cells ≈ 5.24 sr, i.e. strictly *sub*-hemisphere. Every mask with 6–11 base cells — exactly the covers issue #147 unblocks, and the ones the PR #111 guards used to reject outright — is never dissolved by any test. Extending the same sweep to all 4095 masks (nothing else changed, `to_geometry` → `shapely.is_valid`):

```
bits : masks / shapely-invalid   (step 1)
  1..4:  793 /   0
  5   :  792 /   8      <- the 8 masks question 3 lists
  6   :  924 /  56
  7   :  792 / 136
  8   :  495 / 102
  9   :  220 /  32
 10   :   66 /   4
 11,12:   13 /   0
total : 4095 / 338
```

The eight ≤5-base masks it finds are exactly the PR body's list (314, 613, 1226, 1378, 1432, 2197, 2609, 2756) — so the sweep is faithful *within its window*; the window is the problem. `step 3` is worse and the slice hides it even harder:

```
step 3: 715 / 4095 invalid; 108 of them have <=5 base cells.
corpus step-3 slice = the first 60 <=5-bit masks = masks 1..60,
  i.e. only base cells {0,1,2,3,4,5} — never 6..11.
  invalid masks inside that slice: [58]   <- exactly the "+ mask 58 at step 3 only"
```

So both numbers in question 3 are the slice reporting itself, not a measurement of the emit.

It is not coarse-only either. Random half-space covers (50–75 % of the sphere — the "ocean mask / everything except Antarctica" workload the issue names), 200 per order, seed `20260809+order`:

```
order 2: 11/200 invalid (5.5%)
order 3: 15/200 invalid (7.5%)
order 4: 17/200 invalid (8.5%)
```

and at these orders **every one** of them reports at a non-pole latitude (`Ring Self-intersection[180 19.4712206344907]`, `[-180 0]`, …), which is a different mechanism from the pole-corner class — see my separate comment on `normalized_lonlat`.

Two things make this blocking rather than a doc fix:

1. `to_geometry` is public API and 8–17 % of hemisphere-scale covers now return OGC-invalid geometry where `main` raised a curated `ValueError`. Some of it is not benign: `mp.union(shapely.box(-170,-80,170,80))` raises `GEOSException` on masks 629, 741, 613 and 1226 (4 of the 12 invalid masks I spot-checked).
2. The corpus test's own oracle cannot see any of it. I ported `assert_point_sampled` verbatim and ran it on 49 of the 338 invalid masks: **0 parity failures**. Even-odd parity over the ring set is blind to ring simplicity, so growing the corpus without also asserting `is_valid` (or `LinearRing.is_simple` per emitted ring) would still report green.

Concretely, I'd want before merge: drop the `> 5` filter (the whole 4095-mask sweep costs ~26 s in Python here, so it is cheap in Rust), make the step-3 slice sample across all 12 base cells rather than the first 60 masks, add a per-ring simplicity assertion to `assert_point_sampled`, and restate question 3 with the measured numbers.

src_rust/src/dissolve.rs
```diff
@@ -323,16 +461,182 @@ fn xyz_to_lonlat(v: &Vec3) -> (f64, f64) {
...
+                -180.0
+            }
+        } else if let Some(la) = seam_lat(&orig, prev) {
+            if orig[i].1 > la {
+                180.0
+            } else {
+                -180.0
+            }
+        } else {
+            // isolated touch: stay on the off-seam neighbours' side.
```
espg, line 595: 🤖 *from Claude (review)*

**blocking — the isolated-seam-touch rule and the stitcher's meridian connectors are not coordinated: a connector runs straight up/down ±180 through touch vertices this rule has just pinned onto the same side, producing a non-simple ring. This, not the pole corner, is the dominant invalidity at orders 2–4.**

This branch pins a vertex that merely *touches* the antimeridian (off-seam neighbours on both sides) onto `prev`'s side of ±180. `stitch_segments` / `next_segment` then draws its connectors purely from segment **ends and starts** — it has no idea those touch vertices are sitting on the same meridian in between. Concretely, from the order-2 half-space family (`np.random.default_rng(20260809+2)`, first invalid draw, 120 cells, `Ring Self-intersection[180 19.4712206344907]`), the emitted shell is:

```
 27  (168.75,  30.0000)
 28  ( 180.00,  19.4712)   <- isolated seam touch, pinned to +180 by this branch
 29  (168.75,   9.5941)
 30  ( 180.00,   0.0000)   <- isolated seam touch, pinned to +180
 31  (168.75,  -9.5941)
 32  ( 180.00, -19.4712)   <- genuine cut end on +180
 33  ( 180.00,  90.0000)   <- next_segment's north-pole connector: straight up x=180
 34  (-180.00,  90.0000)
 35  (-180.00, -19.4712)
```

Edge 32→33 sweeps lat −19.47 → +90 along `x = 180` and passes exactly through vertices 30 and 28. `next_segment` only asked "is there a same-side segment *start* above −19.47?" — there is not, so it wrapped the pole — but the ring's own body is already on that meridian.

I classified all 15 invalid emits in the order-3 half-space family (same generator, seed `20260809+3`) by mechanism:

```
seam-collinear connector edge strictly containing another seam vertex of the same ring : 15/15
two collinear ±180 edges overlapping in latitude                                       :  0/15
```

so it is a single mechanism, and it is not the pole corner. Location tally over the full order-1 mask corpus, by the latitude GEOS reports:

```
step 1: 326 at a pole, 12 not at a pole
step 3: 642 at a pole,  73 not at a pole
```

— the pole cases dominate at order 1 (coarse cells, few seam touches), but at orders 2–4 (`Ring Self-intersection[180 19.47…]`, `[-180 0]`, `[-180 30]`) they are **all** non-pole. So question 3's "colinear lat-±90 overlaps where the covered region connects through a pole corner" describes the order-1 subset only.

Two supporting observations, so this is judged fairly:

* This is *not* a regression in aggregate. I measured the released 0.9.6 wheel (= `origin/main`) over the same 4095 masks: **2810–2829 raise, ~671–678 of the ~1266 that do emit are shapely-invalid**. This PR takes that to 0 raises / 338 invalid. The direction is a large win; the objection is that the residual is 338 and 715 (step 3), not 9, and that the mechanism named in question 3 is not the one that dominates.
* Fixing it looks local rather than "more surgery": before stitching, split each segment at its own ±180 vertices (or have `next_segment` treat a same-side seam *vertex* between the end and the chosen start as a stop), so the connector never overruns a point the ring already occupies.

(Aside, since I measured it while establishing the baseline: `origin/main` is *nondeterministic* on this cover — eight successive `to_geometry` calls in one process on mask 613 gave `['valid','valid','valid','ValueError','valid','valid','valid','valid']`, and the mix varies per process. This PR's build returns the same verdict eight times out of eight. Worth a line in the phase-5 divergence report, since "old behavior" for a given cover is not always a single value.)

src_rust/src/dissolve.rs
```diff
@@ -293,11 +349,93 @@ fn chain_rings(survivors: &[(u32, u32)], id_xyz: &[Vec3]) -> Vec<Vec<Vec3>> {
...
+/// even-odd boundary, but not an OGC-simple ring.  Splitting at each
+/// revisit yields the equivalent decomposition into simple rings touching
+/// at a point (the dual of the covered-side corner-touch pinch the pairing
+/// already resolves).  A split at a *pole* vertex re-pairs that pole's
+/// passages onto their adjacent wedges; [`ring_to_planar`]'s empty-bracket
+/// span rule renders each re-paired passage over exactly its own wedge, so
+/// the planar shoelace budget stays consistent.  Each cycle is rotated to
+/// start at its minimal vertex id, keeping the canonical-start determinism
+/// of issue #155.
+fn split_at_revisits(chain: Vec<u32>) -> Vec<Vec<u32>> {
```
espg, line 374: 🤖 *from Claude (review)*

**non-blocking (pre-existing, but squarely in this function's contract) — the corner-touch pinch resolves the *wrong* way for every `step > 1`, and no test covers a pinch at `step > 1`.**

The docstring's promise — "the equivalent decomposition into simple rings touching at a point" — holds at `step == 1` and fails at every larger step, for the base-cell corner where a north-cap cell touches a south-cap cell at the equator. Cover = base cells {0, 1, 8, 9} (they touch only at the equatorial corners (45, 0) and (135, 0), enclosing a lens-shaped hole):

```
order 1 step 1: shells 2 holes 0  valid True   Valid Geometry
order 1 step 2: shells 1 holes 1  valid False  Interior is disconnected[135 0]
order 1 step 3: shells 1 holes 1  valid False  Interior is disconnected[135 0]
order 1 step 4: shells 1 holes 1  valid False  Interior is disconnected[135 0]
order 1 step 8: shells 1 holes 1  valid False  Interior is disconnected[135 0]
order 2 step 1: shells 2 holes 0  valid True
order 2 step 2: shells 1 holes 1  valid False
order 3 step 1: shells 2 holes 0  valid True
order 3 step 2: shells 1 holes 1  valid False
```

Same on both engines (`_dissolved_rings_py` agrees ring-for-ring), and the parity oracle sees nothing (0 hard mismatches at either step) — the *region* is right, the decomposition is not: a shell plus a hole touching it at two points is an OGC "interior is disconnected" polygon, whereas step 1 correctly gives two touching shells. The flip is caused by the incident edge azimuths at the pinch vertex changing between chord-per-cell-edge and sub-edge chords, so `chain_rings`' LIFO bracket matching pairs the other way and `split_at_revisits` then has nothing to split.

I checked whether this is yours: it is **not** — the released 0.9.6 wheel (= `origin/main`) prints the identical table, so it predates the PR. Flagging rather than asking for a fix (§4), with two reasons it is worth a line here anyway:

1. The generic diagonal corner-touch *does* survive `step > 1` (the `test_dissolve_corner_touch_yields_simple_rings` fixture — two order-4 cells in face 4 — gives 2 valid shells at steps 1, 2, 3 and 4), so it is specifically the base-cell equatorial corner that flips. That makes it a cheap regression test to add either way.
2. `test_dissolve_corner_touch_yields_simple_rings` and the new corpus both exercise pinches at `step == 1` only, and this PR's step-3 slice is masks 1..60 (base cells 0–5, so never a north/south base-cell corner pair). Adding `[0, 1, 8, 9]` at step 3 to the corpus would pin whichever behaviour is decided.

mortie/tests/test_dissolve_hemisphere.py
```diff
@@ -0,0 +1,397 @@
...
+        last = x
+    while len(out) > 1 and max(
+            abs(a - b) for a, b in zip(sph(out[0]), sph(out[-1]))) < 1e-12:
+        out.pop()
+    return out
+
+
+@pytest.mark.parametrize("name,nested,order", _AUDIT_COVERS,
+                         ids=[c[0] for c in _AUDIT_COVERS])
+def test_dissolve_outline_areas_match_spherely(name, nested, order):
```
espg, line 315: 🤖 *from Claude (review)*

**non-blocking — the spherely goldens are a real external oracle for the seven fixtures they cover, but the route does not generalise to the hemisphere+ regime, so it cannot be cited as "an external oracle on the *classification*" in general.**

I ran exactly this test's construction (`_spherely_ring` dedup, `create_polygon(shell=…, oriented=True)`, frame counting 4π, holes reversed and subtracted) over all 4095 order-1 masks at step 1:

```
s2 refuses to construct the emit at all : 70 / 4095   e.g. mask 1606 -> ValueError('polygon is not valid: Edge 3 crosses edge 8')
|s2 area / exact area - 1| > 0.05       : 50 / 4095   worst 0.568 (mask 3694), several at exactly 0.500
the 7 pinned fixtures                   : all in the well-behaved subset
```

Mask 1606 is only 5 base cells and its *planar* emit is shapely-valid, so this is not the same set as the invalidity finding — it is the planar↔spherical mismatch: a ring that runs up `+180` and later down `-180`, or along `lat = ±90`, is two distinct planar edges but one degenerate spherical edge, and `_spherely_ring`'s consecutive-vertex dedup cannot collapse non-adjacent ones.

Importantly, **this is not a classification error**: I re-checked those masks with the parity oracle (0 hard mismatches on all of them) and against the planar shoelace budget — e.g. mask 3694 emits 0.6939 of the map against an exact 0.6667, mask 2007 emits 0.7500 against 0.7500 — so the shells/holes are right and only the s2 re-reading is wrong.

Two suggestions, both cheap:

* say in the test docstring (and in the PR body's "spherely goldens" section) that the route is valid for emits whose rings do not double the seam or the pole line, rather than presenting it as a general classification oracle; and
* if a general external oracle is wanted, `shapely.area` against `Σ planar shoelace` is one that *does* hold everywhere — I checked all 4095 masks and `abs(mp.area - Σ shoelace) < 1e-6` for every one, which is a decent (if weaker) invariant to pin at corpus scale.

mortie/tests/test_dissolve_hemisphere.py
```diff
@@ -0,0 +1,397 @@
...
+    strictly interior to their covering cell, hence bounded away from the
+    emitted outline (centres exactly on the ±180 seam are nudged west, with
+    membership re-evaluated at the nudged point).
+    """
+    import shapely
+
+    nested, depths = _rust_mort2nested(
+        np.ascontiguousarray(np.asarray(cover, dtype=np.uint64)))
+    order = int(depths.max())
+    assert int(depths.min()) == order  # single-order fixtures
```
espg, line 130: 🤖 *from Claude (review)*

**non-blocking — mixed-order MOCs are absent from every new test family, and this assertion is what locks them out; the calibration path they exercise (calibration order ≠ emission order) is unpinned.**

`assert int(depths.min()) == order  # single-order fixtures` means no MOC can go through the Python point sampler, and `_AUDIT_COVERS`, `_both_caps_cover`, the sweep-slice test and both Rust corpus tests are single-order too. The only MOC dissolve test in the suite is the pre-existing `test_dissolve_mixed_order_moc` (`mortie/tests/test_geometry.py`), a small sub-hemisphere cover that only compares areas.

That matters here specifically because of the new calibration:

```rust
let (nest0, order0) = mort2nested(morton[0]);
let ccw = spherical_signed_area(&cell_boundary_loop(order0, nest0, step)) > 0.0;
```

`survivor_edges` densifies a mixed-order MOC to `max_depth`, so on a MOC the calibration cell is measured at `morton[0]`'s **own** order while every chained ring is built at `max_depth`. The phase-1 contract does cover this (`cell_boundary_emission_orientation_is_uniform` pins handedness over orders 0–5), but the *composition* — a hemisphere+ MOC whose first word is an order-0 or order-1 cell and whose emission order is 4 — is only correct by that inference, never by a test. `debug_assert_cover_on_left` handles the mixed-order case explicitly (`moc::to_order`), which shows the author had it in mind; nothing calls it with a MOC.

For what it is worth I checked it empirically and it is fine: 12 hemisphere+ MOCs built as `compress_moc(lat > lat0)` at order 4 (`lat0 ∈ [-58, +8]`, 64–180 words, orders {0,1,2,3,4} mixed in every one) all dissolve, all point-sample clean at order 6, and the Rust and Python engines agree ring-for-ring. So this is a gap, not a bug — but a hemisphere+ MOC is the single most likely real-world input for this feature (`morton_coverage_moc` output fed straight to `to_geometry`), and it would be worth one fixture in `_AUDIT_COVERS`-style form plus the corresponding relaxation here (the Rust `assert_point_sampled` already handles mixed depth via `moc::to_order`; the Python helper just needs the same treatment instead of the assert).

mortie/tests/test_dissolve_hemisphere.py
```diff
@@ -0,0 +1,397 @@
...
+    for h in holes:
+        ring = _spherely_ring([tuple(map(float, v)) for v in h])
+        p = spherely.create_polygon(shell=ring[::-1], oriented=True)
+        total -= spherely.area(p) / spherely.EARTH_RADIUS_METERS ** 2
+    exact = len(nested) * np.pi / (3.0 * 4.0 ** order)
+    assert abs(total / exact - 1.0) < 0.05, (name, total, exact)
+    assert abs(total - _SPHERELY_GOLDENS[name]) < 1e-6, (name, total)
+
+
+def test_roundtrip_coverage_contains_cover_normalize_false():
```
espg, line 343: 🤖 *from Claude (review)*

**non-blocking — "the containment half holds for every hole-free emit, hemisphere+ included" has exactly one counterexample in the order-1 corpus, and it is the whole-sphere frame emit.**

I ran this test's body over all 4095 order-1 masks rather than the six audit covers (`to_geometry` → `from_geometry(..., order=1, normalize=False)` → is every source cell back?):

```
hole-free emits                 : 3572
of those, missing cells > 0     :    1   -> mask 4095 (all 12 base cells), 36 of 48 cells lost
holed emits with missing cells  :  485   (expected: the documented normalize=False hole-winding issue)
from_geometry raises            :    0
```

So the claim is very nearly true, which is worth saying — but the frame shell `[-180,180]×[-90,90]` that `classify_and_split` prepends for a whole-sphere cover does not round-trip through the parity fill: it comes back as 12 of 48 order-1 cells. `whole_sphere_cover_emits_frame` pins the emit, and this test pins the round trip, but no test connects the two, so the gap is invisible.

Suggest either extending the loop to include the whole-sphere cover with an explicit `pytest.xfail`/comment, or narrowing the PR-body sentence to "every hole-free emit that does not include the map frame". The 3571/3572 result is a stronger claim than the six fixtures currently support, so it is worth pinning at corpus scale in either case.

src_rust/src/dissolve/tests.rs
espg: 🤖 *from Claude*

Folded in 08803c1: the corpus cap is gone — `sweep_corpus_order1_combos_dissolve_point_sampled` now runs all 4095 non-empty masks (point-sampled; 4.4 s debug), so the hemisphere+ regime is inside the acceptance corpus, and the PR body's question (3) is rewritten around your measured numbers (338 invalid at step 1 / 715 at step 3, full-mask family) instead of the "9 of 3158" artifact. The parity result you measured (0 failures on the invalid masks, Σ shoelace == shapely.area on all 4095) is quoted there too.

What is *not* folded: making shapely validity hold corpus-wide — that is the standing decision for espg in question (3) (accept + follow-up carrying your corpus as the acceptance set, vs. requiring the connector fix in this PR), since it is another round of stitcher surgery.

src_rust/src/dissolve.rs
espg: 🤖 *from Claude*

Acknowledged, and thank you for running the mechanism to ground — the connector-through-touch-vertex diagnosis (with the pole/non-pole tally and the 15/15 order-3 confirmation) is now what the PR body's question (3) states, replacing my pole-corner misattribution. Your incidental finding (origin/main still per-call nondeterministic on mask 613) is recorded as a caveat on the phase-5 old-side counts.

The fix you sketch — splitting stitch connectors at seam-lying ring vertices so pieces touch at a point — is the option (a) follow-up posed for espg in question (3); not folded here per the same scope call as your parent finding.

src_rust/src/dissolve.rs
espg: 🤖 *from Claude*

Standing for espg: pre-existing on released 0.9.6 exactly as you measured, both engines, and out of this PR's scope (the step>1 pinch orderings are chaining territory, not the classifier). Recorded in the PR body as question (6) with your comment linked, as a candidate follow-up issue — not opened unilaterally per repo convention. The missing `step > 1` pinch test would belong to that follow-up.

mortie/tests/test_dissolve_hemisphere.py
espg: 🤖 *from Claude*

Folded as a scope note in the PR body's spherely section: the goldens stay a named-fixture oracle, with your measurement (s2 refusing ~70 and mis-measuring ~50 of 4095 full-mask emits on seam/pole doubling) quoted as the reason the route does not generalise corpus-wide. No test change — the golden test never claimed corpus scope, and the corpus-wide classification oracle remains the point sampler.

mortie/tests/test_dissolve_hemisphere.py
espg: 🤖 *from Claude*

Folded in 08803c1: `mixed_order_hemisphere_moc_dissolves` — base cells 0-5 at order 1 plus base cell 6 as its sixteen order-2 children (7.33 sr, mixed-order, hemisphere+) — dissolves and point-samples correctly, exercising exactly the path you named (calibration reads morton[0]'s own order while the boundary is densified to the finest order; emission handedness is order-uniform, so one witness cell suffices). Thanks for verifying the 12 MOC covers empirically before flagging it as gap-not-bug.

mortie/tests/test_dissolve_hemisphere.py
espg: 🤖 *from Claude*

Folded into the PR body's round-trip section: the whole-sphere frame-only emit is excluded from the stated invariant (the frame ring is a planar convention with no faithful spherical reading, so `from_geometry` cannot recover the cover from it), with your 3571/3572 measurement cited. The pinned test never included the whole-sphere cover, so no test change — the body text was the inaccuracy.

